PR #1, opened 2015-01-24
Integrated PyCUDD as BDD Solver.

marcogario:
- Added files in patches/ to explain how to install PyCUDD 
 - This also introduces a few fixes to pycudd and the method pycudd.GetDefaultManager()
- Implemented BddSolver as an incremental solver
- Added 'bdd' solver in Factory
- Most tests assumed that all solvers could work on QF_UFLIRA. Explicited added the requirement.
- Added ad-hoc tests

### DIFF
--- a/pysmt/solvers/bdd.py
+++ b/pysmt/solvers/bdd.py
@@ -192,10 +192,6 @@ class BddConverter(DagWalker):
         cube = self.ddmanager.IndicesToCube(indices, len(var_list))
         return cube
 
-    @deprecated("convert")
-    def expr_to_bdd(self, formula):
-        return self.walk(formula)
-
     def declare_variable(self, var):
         if not var.is_symbol(type_=types.BOOL): raise TypeError
         assert_ddmanager(self.ddmanager)
--- a/pysmt/solvers/bdd.py
+++ b/pysmt/solvers/bdd.py
@@ -143,9 +143,6 @@ class BddSolver(Solver):
             l = self.backtrack.pop()
             self.assertions_stack = self.assertions_stack[:l]
 
-    def exit(self):
-        raise NotImplementedError
-
     def __del__(self):
         del self.ddmanager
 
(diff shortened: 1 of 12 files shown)


pysmt/solvers/bdd.py
```diff
@@ -0,0 +1,324 @@
...
+    @clear_pending_pop
+    def declare_variable(self, var):
+        self.converter.declare_variable(var)
+
+    @clear_pending_pop
+    def add_assertion(self, formula, named=None):
+        assert_ddmanager(self.ddmanager)
+        bdd_expr = self.converter.convert(formula)
+        _, last_state = self.assertions_stack[-1]
+        new_state = last_state.And(bdd_expr)
```
mikand: At stack level n you have the conjunction of all the levels i <= n...
* A comment is needed
* Why doing this computation here and not in the ```solve()``` method?

pysmt/solvers/bdd.py
```diff
@@ -0,0 +1,324 @@
...
+            self.backtrack.append(len(self.assertions_stack))
+
+    @clear_pending_pop
+    def pop(self, levels=1):
+        for _ in xrange(levels):
+            l = self.backtrack.pop()
+            self.assertions_stack = self.assertions_stack[:l]
+
+    def exit(self):
+        raise NotImplementedError
```
mikand: Why raising error? Isn't ```pass``` a better option?

pysmt/solvers/bdd.py
```diff
@@ -0,0 +1,324 @@
...
+
+    def cube_from_var_list(self, var_list):
+        assert_ddmanager(self.ddmanager)
+        indices = pycudd.IntArray(len(var_list))
+        for i, v in enumerate(var_list):
+            indices[i] = self.var2node[v].NodeReadIndex()
+        cube = self.ddmanager.IndicesToCube(indices, len(var_list))
+        return cube
+
+    @deprecated("convert")
```
mikand: This is new, so we can avoid to create deprecated code, or is there a specific reason?

pysmt/solvers/bdd.py
marcogario: Fixed both points in d7d6616

pysmt/solvers/bdd.py
marcogario: Fixed in feafd06

PR #2, opened 2015-01-25
Dynamic Walker Function

marcogario:
Whenever a walker finds an unexpected node-type, it will ask the environment which function to use. This can be used to define new node-types. See test_dwf.py for an example usage.

### DIFF
--- a/pysmt/environment.py
+++ b/pysmt/environment.py
@@ -48,7 +48,7 @@ class Environment(object):
         self._factory = None
         # Configurations
         self.enable_infix_notation = False
-
+        self.dwf = {} # {nodetype, {walker: function}}
 
     @property
     def formula_manager(self):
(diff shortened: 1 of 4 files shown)


pysmt/environment.py
```diff
@@ -48,7 +48,7 @@ def __init__(self):
         self._factory = None
         # Configurations
         self.enable_infix_notation = False
-
+        self.dwf = {} # {nodetype, {walker: function}}
```
mikand: Maybe a bit of a comment to explain what the hack is DWF... ;-)

pysmt/test/test_dwf.py
```diff
@@ -0,0 +1,57 @@
...
+            self.walk(formula.arg(1))
+            self.stream.write(self.tb(")"))
+            return
+
+        # Shortcuts for function in env
+        add_dwf = get_env().add_dynamic_walker_function
+        create_node = get_env().formula_manager.create_node
+
+        # Define the new node type and register the walkers in the env
+        XOR = ALL_TYPES[-1] + 1
```
mikand: We shall think of a nicer way of obtaining fresh ids, but for now this is OK

PR #8, opened 2015-02-17
Installers

marcogario:
Clean-up of installers branch, separating stuff from installers from CI (Shippable)


install.py
```diff
@@ -0,0 +1,419 @@
...
+import os
+import tarfile
+import sys
+import platform
+import zipfile
+import argparse
+import sys
+import time
+
+#MG: TODO: Add credits for this class.
```
marcogario: Shall we add credits for this ProgressBar? Check license and maybe move to independent file.

install.py
```diff
@@ -0,0 +1,419 @@
...
+    while True:
+        buff = u.read(block_sz)
+        if not buff:
+            break
+
+        f.write(buff)
+        if len(meta.getheaders("Content-Length")) > 0:
+            bar + len(buff)
+            bar.show_progress()
+    print ""
```
marcogario: print()

marcogario:
We should add an option to specify a mirror from which to download the solvers. I think this would be a polite thing to do, especially if we start heavily running this script during CI.

install.py
```diff
@@ -0,0 +1,349 @@
...
+
+    # Extract the pycudd distribution
+    untar(archive, BASE_DIR)
+
+    # patch the distribution
+    os.system("cd %s; patch -p1 -i %s/patches/pycudd.patch" % (dir_path, CWD))
+
+    # Build the pycudd
+    # NOTE: -j is not supported by this building system
+    # MG: shouldn't this check if we are 64 or 32 bit?
```
marcogario: Should we check 64 vs 32?

PR #11, opened 2015-02-20
Thread-safe typing

marcogario:


pysmt/typing.py
```diff
@@ -130,6 +151,17 @@ def __str__(self):
...
         return True
 
+    def __eq__(self, other):
+        if other is None:
+            return False
+        if self.type_id != other.type_id:
+            return False
+        if id(self) == id(other):
+            return True
+        return str(self) == str(other)
```
mikand: OK for now, but why can't we write ```self.return_type == other.return_type and self.param_types == other.param_types```?

Issue #13, opened 2015-02-21
Create get_model shortcut


marcogario:
We should add a get_model shortcut that takes care of creating the solver, solving and returning a model (if one is found).
The signature would be similar to is_sat. It should return a Model (Eager or Lazy) or None if the formula is unsat.

marcogario:
PR #17

PR #17, opened 2015-02-23
Implemented get_model() shortcut as per isuue #13

mikand:


pysmt/shortcuts.py
```diff
@@ -237,6 +237,19 @@ def is_sat(formula, quantified=False, solver_name=None, logic=None):
                               solver_name=solver_name,
                               logic=logic)
 
+def get_model(formula, quantified=False, solver_name=None, logic=None):
+    """ Similar to :py:func:`is_sat` but returns a model if the formula is
+    satisfiable, otherwise None."""
+    env = get_env()
+    if formula not in env.formula_manager:
+        warnings.warn("Warning: Contextualizing formula during is_sat")
```
marcogario: Unless there are multiple managers (e.g., during multi-processing), we do not need this check. 
Also, we do not do it for the other shortcuts. Finally, it might make more sense to do this within the formula manager, instead than in the shortcut.
(note: typo is_sat/get_model)

pysmt/test/test_solving.py
```diff
@@ -63,6 +63,39 @@ def test_is_sat(self):
...
+        f = And(varA, Equals(varX, Real(8)))
+
+        res = get_model(f)
+        self.assertIsNotNone(res, "Formula was expected to be SAT")
+        self.assertTrue(res.get_value(varA) == TRUE())
+        self.assertTrue(res.get_value(varX) == Real(8))
+
+        for solver in get_env().factory.all_solvers():
+            res = get_model(f, solver_name=solver)
+            self.assertIsNotNone(res, "Formula was expected to be UNSAT")
```
marcogario: typo: UNSAT/SAT

pysmt/factory.py
```diff
@@ -261,6 +261,16 @@ def is_sat(self, formula, quantified=None, solver_name=None, logic=None):
...
         return solver.is_sat(formula)
 
+    def get_model(self, formula, quantified=None, solver_name=None, logic=None):
+        solver = self.Solver(quantified=quantified, name=solver_name, logic=logic)
+        solver.add_assertion(formula)
+        check = solver.solve()
+        retval = None
+        if check:
+            retval = solver.get_model()
+        solver.exit()
```
marcogario: Good catch! In all other shortcuts we are not destroying the solver!!! (e.g., see is_valid below). Maybe it would be better to use a context in all of these.

PR #26, opened 2015-03-01
Solvers mirror

marcogario:
*Merging to yices_and_cudd_ci_tests*

- Added Mirror for solvers download
- Explicit container to push completed builds
- Extended testing matrix to include multiple versions of python for the library only configuration
- BUGFIX: ./install --check was not raising an exception if the solver (PYSMT_SOLVER) was not installed


marcogario:
The fix on _check_ highlighted that CVC4 was not executing the tests (error in compilation).

PR #27, opened 2015-03-02
Yices and cudd ci tests

mikand:
Fixed support of Yices and CUDD and added new configurations for CI


pysmt/test/test_bdd.py
```diff
@@ -47,11 +46,14 @@ def test_basic_bdd_variables(self):
 
     @skipIfSolverNotAvailable("bdd")
     def test_basic_expr(self):
+        from pysmt.solvers.bdd import LockDdManager
```
marcogario: Move outside of the test

PR #31, opened 2015-03-06
Picosat Integration

mikand:
Added support for PicoSAT solver with dedicated wrapper


pysmt/cnf.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def convert_as_formula(self, formula):
+        lsts = self.convert(formula)
+
+        conj = []
+        for clause in lsts:
+            if len(clause) == 0:
+                return self.mgr.FALSE()
+            elif len(clause) == 1:
+                conj.append(next(iter(clause)))
```
marcogario: Is this the best way of getting the singleton element?

pysmt/solvers/pico.py
```diff
@@ -0,0 +1,133 @@
...
+
+        self.mgr = environment.formula_manager
+        self.pico = picosat.picosat_init()
+        self.converter = None
+        self.cnfizer = CNFizer(environment=environment)
+        self.latest_model = None
+        self._var_ids = {}
+
+
+    def _get_var_id(self, symbol):
```
marcogario: Couldn't we wrap the cnfizer and this function into a light-weight converter?

pysmt/test/test_cnf.py
```diff
@@ -0,0 +1,57 @@
...
+from pysmt.cnf import CNFizer
+from pysmt.logics import QF_BOOL, QF_LRA, QF_LIA
+from pysmt.test import TestCase, skipIfNoSolverForLogic
+from pysmt.test.examples import EXAMPLE_FORMULAS
+
+class TestCnf(TestCase):
+
+    def do_examples(self, logic):
+        conv = CNFizer()
+        for example in EXAMPLE_FORMULAS:
```
marcogario: These examples are relatively small. Can we take a (QF) SMT-LIB problem, build the CNF and then check that it yields the same result as the original problem?

pysmt/cnf.py
```diff
@@ -0,0 +1,217 @@
...
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+import pysmt.shortcuts as shortcuts
+from pysmt.walkers import DagWalker
+from pysmt.typing import BOOL
+
+class CNFizer(DagWalker):
+
+    THEORY_PLACEHOLDER = "__Placeholder__"
```
marcogario: This is used to avoid modifying Theory subformulas, right?

pysmt/cnf.py
```diff
@@ -0,0 +1,217 @@
...
+    def walk_function(self, formula,  args):
+        return CNFizer.THEORY_PLACEHOLDER
+
+    def walk_real_constant(self, formula,  args):
+        return CNFizer.THEORY_PLACEHOLDER
+
+    def walk_bool_constant(self, formula,  args):
+        if formula.is_true():
+            return formula, CNFizer.TRUE_CNF
+        else:
```
marcogario: The else seems incorrect.

pysmt/cnf.py
marcogario: Maybe we can merge all Theory Walkers to make it more explict:
```
 def walk_theory(...):
     return CNFIzer.THEORY_PLACEHOLDER
```

pysmt/cnf.py
```diff
@@ -0,0 +1,217 @@
...
+        return CNFizer.THEORY_PLACEHOLDER
+
+    def walk_minus(self, formula,  args):
+        return CNFizer.THEORY_PLACEHOLDER
+
+    def walk_times(self, formula,  args):
+        return CNFizer.THEORY_PLACEHOLDER
+
+    def walk_equals(self, formula, args):
+        assert all(a == CNFizer.THEORY_PLACEHOLDER for a in args)
```
marcogario: Is there a reason to do this here and not in, e.g., _walk_plus()_ ?

pysmt/cnf.py
```diff
@@ -0,0 +1,217 @@
...
+            res = self._introduced_variables[formula]
+        else:
+            res = self.mgr.FreshSymbol()
+            self._introduced_variables[formula] = res
+        return res
+
+    def convert(self, formula):
+        tl, cnf = self.walk(formula)
+        res = [frozenset([tl])]
+        for clause in cnf:
```
marcogario: This second part is not needed for correctness of convert(). Returning _res_ is enough.
Lets factor out this "simplification" part.

pysmt/cnf.py
```diff
@@ -0,0 +1,217 @@
...
+        conj = []
+        for clause in lsts:
+            if len(clause) == 0:
+                return self.mgr.FALSE()
+            elif len(clause) == 1:
+                conj.append(next(iter(clause)))
+            else:
+                conj.append(self.mgr.Or(clause))
+
+        if len(conj) == 0:
```
marcogario: This should be done automatically by And. and of an empty list is TRUE, of a singleton is the elemtn.

pysmt/cnf.py
```diff
@@ -0,0 +1,217 @@
...
+            cnf.append(frozenset([k, self.mgr.Not(a)]))
+            for clause in c:
+                cnf.append(clause)
+        return k, frozenset(cnf)
+
+    def walk_not(self, formula, args):
+        a, cnf = args[0]
+        if a.is_true():
+            return self.mgr.FALSE(), CNFizer.TRUE_CNF
+        elif a.is_false():
```
marcogario: Shouldn't this be CNFizer.FALSE_CNF ?

pysmt/test/test_cnf.py
marcogario: Done.  cdf9807

marcogario:
As far as I understand, the walker uses two returns values, the variable representing the current subformula, and the current CNF. 
However, during the visit, elements are only added to the current CNF. Thus, couldn't we make it an element of the walker, and each function will modify it? E.g., *self.g_cnf* ? This should also help with the memoization part, since we just need to know the variable representing the subformula, but its definition is provided elsewhere and we do not need to repeat it over and over.

pysmt/cnf.py
mikand: No, the contract of walk is the following: 
it returns a pair (v, cnf) such that the input formula is equivalent to v & cnf. v can be either a variable or a boolean constant and cnf is a set of clauses that are needed to explain v.

In this case I'm returning False, True that is logically false. The reason why I use TRUE_CNF instead of FALSE_CNF is that I need nothing to "explain" false. It would be incorrect to use FALSE_CNF. 

For this reason,

pysmt/cnf.py
mikand: Same as above

pysmt/cnf.py
mikand: Sure. Equals, LT, LE are boolean atoms, therefore can appear as leaves in the CNF, while Plus & Co are terms that cannot appear in the CNF.

pysmt/solvers/pico.py
mikand: Not sure. The whole converter design is not applicable to picosat as there is no concept of foumula. All you can do is assert clauses (and the index of the variables must be specified taking care of ```picosat.picosat_inc_max_var(self.pico)```) hence I think it is much more natural to have it here...

marcogario: Ok. Do you know how back conversion for unsat-cores might look like here? Would a converter make sense for that? If not, then lets leave it like this.

Issue #10, opened 2015-02-20
Integrate install.py in setup.py and documentation


marcogario:
Now that we have a way to install the solvers we should:

1) Document it in the README
2) Integrate it with the setup.py, so that it can either be run during _pip install_ or we add an alias _pysmt_install_ that can be called by the user.

marcogario:
Calling it automatically from pip install seems an overkill. Pull-request #35 addresses this by adding a command pysmt-install.

PR #35, opened 2015-03-07
Creates "binaries" during installation

marcogario:
We create 3 "binaries" after installation:
* pysmt : full shell that defaults to smtlib behavior
* pysmt-shell: shell that defaults to interactive
* pystm-install: alias for install.py

This addresses #10


marcogario:
I already uploaded a demo pkg for this (thus the change of version), this can be tested calling
``` pip install --user --pre pysmt ```
where the '--pre' is needed since this is tagged as a .dev version.

PR #38, opened 2015-03-09
Getting rid of generic number

mikand:
* Removed the use of GenericNumber and disambiguation by imposing that in a LIRA context real constants having integer value must have a trailing ".0".
E.g. "1 + 2.2" is forbidden if the logic is LIRA, "1.0 + 2.2" is the correct form

* Printers have been extended to correctly implement this behavior

* The MathSAT back converter that was using the disambiguation feature now cleanly recognizes the signature of the term to be built and introduces ToReal nodes whenever needed.


pysmt/smtlib/parser.py
```diff
@@ -261,20 +247,16 @@ def _minus_or_uminus(self, *args):
     def _equals_or_iff(self, left, right):
```
marcogario: Lets move this in Formula as SmartEquals or EqualsIff.

PR #42, opened 2015-03-12
Annotations and unsat cores

mikand:
Added support for SMTLIB annotations in parser


pysmt/smtlib/annotations.py
```diff
@@ -0,0 +1,94 @@
...
+    def add(self, formula, annotation, value=None):
+        """Adds an annotation for the given formula, possibly with the
+        specified value"""
+        term_annotations = self._annotations.setdefault(formula, {})
+        values = term_annotations.setdefault(annotation, set())
+        if value is not None:
+            values.add(value)
+
+
+    def remove(self, formula):
```
marcogario: All remove methods assume that formula is defined in annotations. Is this intended?

pysmt/smtlib/annotations.py
```diff
@@ -0,0 +1,94 @@
...
+        res = []
+        for f,amap in self._annotations.iteritems():
+            if annotation in amap:
+                if value is None:
+                    res.append(f)
+                else:
+                    if value in amap[annotation]:
+                        res.append(f)
+        return set(res)
+
```
marcogario: Another useful method could be:

- has_annotation(formula, annotation, value=None) (that is similar to all_annotated_formulae, but returns True/False).

PR #44, opened 2015-03-13
Get free Variables

marcogario:
Refactoring of FNode.get_dependencies() to use iterative walkers.

- Introduces the FreeVarsOracle, and the get_free_variables function/shortcut.
- Added Deprecation warning to get_dependencies
- Updated code-base to use get_free_variables

### DIFF
--- a/pysmt/oracles.py
+++ b/pysmt/oracles.py
@@ -20,6 +20,7 @@ properties of formulae.
 
  * QuantifierOracle says whether a formula is quantifier free
  * TheoryOracle says which logic is used in the formula.
+ * FreeVarsOracle says which variables are free in the formula
 """
 
 import pysmt.walkers as walkers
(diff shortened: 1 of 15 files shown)


pysmt/oracles.py
```diff
@@ -168,10 +169,73 @@ def get_theory(self, formula):
...
+
+    def walk_symbol(self, formula, args):
+        return frozenset([formula])
+
+    def walk_constant(self, formula, args):
+        return frozenset()
+
+    def walk_function(self, formula, args):
+        # MG: Do we need to create a list?
+        res = set([formula._content.payload])
```
mikand: I would avoid using _content.
This is equivalent to ```formula.function_name()```

pysmt/oracles.py
```diff
@@ -168,10 +169,73 @@ def get_theory(self, formula):
...
+        return self.walk(formula)
+
+    def walk_simple_args(self, formula, args):
+        res = set()
+        for arg in args:
+            res.update(arg)
+        return frozenset(res)
+
+    def walk_quantifier(self, formula, args):
+        return args[0].difference(formula._content.payload)
```
mikand: I would avoid using _content.
This is equivalent to formula.quantifier_vars()

pysmt/oracles.py
```diff
@@ -168,10 +169,73 @@ def get_theory(self, formula):
...
 
-
 # EOC TheoryOracle
 
 
+# Operators for which Args is an FNode
+DEPENDENCIES_SIMPLE_ARGS = (set(op.ALL_TYPES) - \
+                           (set([op.SYMBOL, op.FUNCTION]) | op.QUANTIFIERS | op.CONSTANTS))
+
+class FreeVarsOracle(pysmt.walkers.DagWalker):
```
mikand: A possible improvement (useful for function models) could be to have a parameter to control if/how function sysmbols are added in the resulting cone.

I see three modes:
* None: meaning that only variables are considered in the cone
* Function Symbols [default and current behavior]: function names are symbols appearing in the cpone, but function applications are discarded
* Function applications: function applications are considered atomic and are retured together with the cones of theit parameters.

marcogario: Lets keep this for a future. I think this goes in the direction of having an implementation in which we do Free Variables vs All Variables. Lets open a ticket with all "versions" that we want to have.

PR #49, opened 2015-03-15
Towards python 3

mikand:
This PR introduces support for Python3:
* Code has been updated to run on both Python2 and Python3 using six (six is now a dependency of pysmt)
* Currently, only the wrappers for msat and for picosat work with Python3
* CI configuration has been updated to run MathSAT and picosat on python 2.7, 3.3 and 3.4
* From now on, all the code must be Python2.7 and Python3 compatible


pysmt/formula.py
```diff
@@ -480,7 +481,7 @@ def Min(self, *args):
             a, b = exprs
             return self.Ite(self.LE(a, b), a, b)
         else:
-            h = len(exprs) / 2
+            h = int(len(exprs) / 2)
```
marcogario: Can we use // instead of casting?

pysmt/smtlib/parser.py
```diff
@@ -66,7 +69,23 @@ def get_formula_fname(script_fname, environment=None, strict=True):
     """
     Returns the formula asserted at the end of the given script
     """
-    with io.BufferedReader(io.FileIO(script_fname, 'r')) as script:
+    if PY2:
```
marcogario: Lets try to keep the py2/py3 things outside the flow. We can introduce a thin wrapper to use in the with statement:

```
with Reader(script_fname) as script:
```

and define Reader depending from PY2/PY3

pysmt/smtlib/parser.py
```diff
@@ -66,7 +69,23 @@ def get_formula_fname(script_fname, environment=None, strict=True):
...
+        with io.TextIOWrapper(open(script_fname, 'rb')) as script:
+            if strict:
+                return get_formula_strict(script, environment)
+            else:
+                return get_formula(script, environment)
+
+
+
+
+    with io.BufferedReader(open(script_fname, 'rt')) as script:
```
marcogario: Is this reachable?

pysmt/smtlib/parser.py
```diff
@@ -141,6 +160,7 @@ def __init__(self, handle):
 
     def tokens(self):
         c = self.read(1)
+        print(c)
```
marcogario: debug print

pysmt/smtlib/parser.py
```diff
@@ -514,8 +534,12 @@ def get_command_generator(self, script):
 
     def get_script_fname(self, script_fname):
         """Given a filename and a Solver, executes the solver on the file."""
-        with io.BufferedReader(io.FileIO(script_fname, 'r')) as script:
-            return self.get_script(script)
+        if PY2:
+            with io.BufferedReader(io.FileIO(script_fname, 'r')) as script:
```
marcogario: Same as above. It might be the case that this Buffered reading patter could go within pysmt.utils.

pysmt/solvers/z3.py
```diff
@@ -146,7 +147,7 @@ def pop(self, levels=1):
     def print_model(self, name_filter=None):
```
marcogario: Unrelated: It seems likne we have multiple implementations of print_model in each solver. We should factorize these.

marcogario:
Looks good! I think the only thing missing are:

- Update to changes
- Update README to clarify which version of python we support (and which solvers for which version)

Issue #16, opened 2015-02-22
Extend quantifier eliminator interface to support logic specification


marcogario:
Currently, we assume that a quantifier eliminator is capable of quantifying UFLIRA. 
A better approach would be to specify the logic as we do for solvers. This would also make it possible to integrate CUDD as Quantifier eliminator for BOOL.

mikand:
We decided to implement this feature: at the moment we have LIA, LRA or BOOL.

Same criterion of the ```Solver``` shortcut will be used

PR #58, opened 2015-03-24
Msat exist elim integration

mikand:
* Added two Quantifier Eliminators using the newly introduced MathSAT support. 

* Added relevant exceptions when a formula is queried to a quantifier eliminator to ensure that the quantification is supported. This will be changed when implementing #16 .


pysmt/solvers/msat.py
```diff
@@ -574,3 +579,81 @@ def declare_variable(self, var):
                                                  tp)
             self.symbol_to_decl[var] = decl
             self.decl_to_symbol[mathsat.msat_decl_id(decl)] = var
+
+
+# Check if we are working on a patched version of the mathsat API
```
marcogario: Since this is part of the official release of mathsat, I think the check is not necessary anymore. We can keep it in order to support older versions of mathsat. In that case, clarify the comment.

pysmt/solvers/msat.py
```diff
@@ -574,3 +579,81 @@ def declare_variable(self, var):
...
+        def eliminate_quantifiers(self, formula):
+            """
+            Returns a quantifier-free equivalent formula of the given
+            formula
+            """
+            return self.walk(formula)
+
+
+        def exist_elim(self, variables, formula, alg='fm'):
+            theory = self.env.theoryo.get_theory(formula)
```
marcogario: In pysmt.oracles there is a method get_logic that does exactly these 3 steps. I think we should use that.

pysmt/solvers/msat.py
```diff
@@ -574,3 +579,81 @@ def declare_variable(self, var):
...
+            logic = get_closer_pysmt_logic(logic)
+            if not logic <= LRA:
+                raise NotImplementedError("MathSAT qunatifier elimination only"\
+                                          " supports LRA (detected logic " \
+                                          "is: %s)" % str(logic))
+
+            fterm = self.converter.convert(formula)
+            tvars = [self.converter.convert(x) for x in variables]
+
+            assert alg in ['fm', 'lw']
```
marcogario: Is the function exist_elim designed to be called directly?
If no, then the "alg" argument is not needed, since we have self.alg .

Moreover, I would convert these 3 lines into a map.

pysmt/solvers/z3.py
```diff
@@ -380,6 +381,15 @@ def __init__(self, environment):
 
 
     def eliminate_quantifiers(self, formula):
+        theory = self.environment.theoryo.get_theory(formula)
```
marcogario: Same as above, use oracles.get_logic

pysmt/solvers/z3.py
```diff
@@ -380,6 +381,15 @@ def __init__(self, environment):
...
+        logic = Logic(name="Detected Logic", description="",
+                      quantifier_free=False, theory=theory)
+        logic = get_closer_pysmt_logic(logic)
+        if not logic <= LRA and not logic <= LIA:
+            raise NotImplementedError("Z3 qunatifier elimination only "\
```
marcogario: typo 'quantifier'

pysmt/test/test_qe.py
```diff
@@ -68,6 +70,51 @@ def test_qe_z3(self):
...
+                     "MathSAT quantifier elimination is not available.")
+    def test_qe_msat_fm(self):
+        qe = QuantifierEliminator(name='msat_fm')
+        self._bool_example(qe)
+        self._real_example(qe)
+        self._alternation_bool_example(qe)
+        self._alternation_real_example(qe)
+
+        with self.assertRaises(NotImplementedError):
+            self._int_example(qe)
```
marcogario: Repeat with

pysmt/test/test_qe.py
```diff
@@ -68,6 +70,51 @@ def test_qe_z3(self):
...
+    @unittest.skipIf('msat_lw' not in get_env().factory.all_qelims(),
+                     "MathSAT quantifier elimination is not available.")
+    def test_qe_msat_lw(self):
+        qe = QuantifierEliminator(name='msat_lw')
+        self._bool_example(qe)
+        self._real_example(qe)
+        self._alternation_bool_example(qe)
+        self._alternation_real_example(qe)
+
+        with self.assertRaises(NotImplementedError):
```
marcogario: Same as above

pysmt/test/test_qe.py
```diff
@@ -96,7 +143,7 @@ def _int_example(self, qe):
         f = ForAll([p], Implies(LT(Int(0), p), LT(q, p)))
         qf = qe.eliminate_quantifiers(f).simplify()
 
-        self.assertTrue(is_valid(Iff(qf, LE(q, Int(0)))))
+        self.assertTrue(is_valid(Iff(qf, LE(q, Int(0))), logic=QF_LIA))
```
marcogario: Why did you need to add this? If no logic is specified, it should be autodetected.
Maybe this is a symptom of a bug in the oracle?

marcogario:
TODOs: Modify CHANGES.rst to include the new feature.

PR #60, opened 2015-03-24
Enabling Smt-Lib wrappers testing in CI

marcogario:
This also fixes a bug in SmtLibParser Initialization that appeared only when testing with wrappers.


marcogario:
There seems to be some problem. Removing from now.

PR #68, opened 2015-04-08
Unsat cores support

mikand:
This PR introduces support for UNSAT cores in msat and z3

### DIFF
--- a/pysmt/factory.py
+++ b/pysmt/factory.py
@@ -385,6 +385,21 @@ class Factory(object):
                 retval = solver.get_model()
             return retval
 
+    def get_unsat_core(self, clauses, solver_name=None, logic=None):
+        if logic == AUTO_LOGIC:
+            logic = get_logic(self.environment.formula_manager.And(clauses),
+                              self.environment)
+
+        with self.UnsatCoreSolver(name=solver_name, logic=logic) \
+             as solver:
+            for c in clauses:
+                solver.add_assertion(c)
+            check = solver.solve()
+            if check:
+                return None
+
+            return solver.get_unsat_core()
+
     def is_valid(self, formula, solver_name=None, logic=None):
         if logic == AUTO_LOGIC:
             logic = get_logic(formula, self.environment)
--- a/pysmt/factory.py
+++ b/pysmt/factory.py
@@ -32,6 +32,9 @@ from pysmt.logics import AUTO as AUTO_LOGIC
 from pysmt.logics import most_generic_logic, get_closer_logic
 from pysmt.oracles import get_logic
 
+import pysmt.solvers.solver
+from pysmt.solvers.solver import SolverOptions
+
 DEFAULT_SOLVER_PREFERENCE_LIST = ['msat', 'z3', 'cvc4', 'yices', 'picosat', 'bdd']
 DEFAULT_QELIM_PREFERENCE_LIST = ['z3']
 DEFAULT_LOGIC = QF_UFLIRA
@@ -43,6 +46,7 @@ class Factory(object):
                  qelim_preference_list=None):
         self.environment = environment
         self._all_solvers = None
+        self._unsat_core_solvers = None
         self._all_qelims = None
         self._generic_solvers = {}
 
@@ -82,7 +86,7 @@ class Factory(object):
                 closer_logic = get_closer_logic(SolverClass.LOGICS, logic)
                 return SolverClass(environment=self.environment,
                                    logic=closer_logic,
-                                   options=None)
+                                   options=SolverOptions(generate_models=True))
             else:
                 raise NoSolverAvailableError("Solver %s is not available" % name)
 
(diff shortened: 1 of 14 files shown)


pysmt/factory.py
```diff
@@ -97,12 +100,60 @@ def get_solver(self, quantified=False, name=None, logic=None):
...
             return SolverClass(environment=self.environment,
                                logic=closer_logic,
-                               options=None)
+                               user_options={"generate_models":True})
         else:
             raise NoSolverAvailableError("No solver is available for logic %s" %\
                                          logic)
 
 
+    def get_unsat_core_solver(self, quantified=False, name=None,
```
marcogario: Maybe we can factor this function together with the one for picking a regular solver. Ideally, we are looking for a solver with some options, inside a set of solvers. In one case the set of solvers is all_solvers, in the other it is all_unsat_core_solvers.

pysmt/factory.py
```diff
@@ -259,6 +319,31 @@ def all_solvers(self, logic=None):
 
         return solvers
 
+    def all_unsat_core_solvers(self, logic=None):
```
marcogario: Same as above. We can factor this with all_solvers, by introducing an intermediate function that looks for a solver within a set.

pysmt/shortcuts.py
```diff
@@ -226,6 +226,14 @@ def Solver(quantified=False, name=None, logic=None):
                                     name=name,
                                     logic=logic)
```
marcogario: We have a shortcut get_model. Is there a good interface for a similar shortcut for unsat cores? (get_unsat_core)?

pysmt/smtlib/solver.py
```diff
@@ -37,11 +40,13 @@ def __init__(self, args, environment, logic, options=None):
         # Initialize solver
         self._send_command(SmtLibCommand(smtcmd.SET_OPTION, [":print-success", "false"]))
         self._send_command(SmtLibCommand(smtcmd.SET_OPTION, [":produce-models", "true"]))
-        if options is not None:
-            for o,v in iteritems(options):
-                self._send_command(SmtLibCommand(smtcmd.SET_OPTION, [o, v]))
+        # if self.options is not None:
```
marcogario: Why is this commented out?

pysmt/solvers/msat.py
```diff
@@ -67,31 +63,79 @@ def __init__(self, environment, logic, options=None, debugFile=None):
...
+        result = formula
+        if self.options.unsat_cores_mode == "named":
+            # If we want named unsat cores, we need to rewrite the
+            # formulae as implications
+            key = self.mgr.FreshSymbol(template="_assertion_%d")
+            result = (key, named, formula)
+            formula = self.mgr.Implies(key, formula)
+
+        term = self.converter.convert(formula)
         res = mathsat.msat_assert_formula(self.msat_env, term)
```
marcogario: Wouldn't we need to assert also 'key'?

pysmt/solvers/msat.py
```diff
@@ -67,31 +63,79 @@ def __init__(self, environment, logic, options=None, debugFile=None):
...
+        if self.options.unsat_cores_mode == "named":
+            return {t[0] : (t[1],t[2]) for t in self.assertions}
+        return None
 
     @clear_pending_pop
-    def solve(self, assumptions=None):
+    def _solve(self, assumptions=None):
         res = None
+
+        n_ass = self._named_assertions()
```
marcogario: Doing this at solving time seems unnecessarily complicated. Why can't we just assert the assumption when calling add_assertion (see comment above)?

pysmt/solvers/msat.py
```diff
@@ -116,8 +160,63 @@ def solve(self, assumptions=None):
...
+            raise SolverStatusError("The solver status has been modified by a" \
+                                    " '%s' command after the last call to" \
+                                    " solve()" % self.last_command)
+
+
+    def get_unsat_core(self):
+        """After a call to solve() yielding UNSAT, returns the unsat core as a
+        set of formulae"""
+        if self.options.unsat_cores_mode == "all":
+            self._check_unsat_core_config()
```
marcogario: The check should be outside the **if** statement.

pysmt/solvers/z3.py
```diff
@@ -128,21 +143,61 @@ def solve(self, assumptions=None):
...
+
+    def _named_assertions_map(self):
+        if self.options.unsat_cores_mode is not None:
+            return {t[0] : (t[1],t[2]) for t in self.assertions}
+        return None
+
+    def get_named_unsat_core(self):
+        """After a call to solve() yielding UNSAT, returns the unsat core as a
+        dict of names to formulae"""
+        if self.options.unsat_cores_mode is None:
```
marcogario: These checks where already implemented in MathSAT. Maybe we can factor them into the UnsatCoreSolver class?

pysmt/test/test_unsat_cores.py
```diff
@@ -0,0 +1,150 @@
...
+
+from pysmt.test import TestCase, skipIfSolverNotAvailable
+from pysmt.shortcuts import get_env, And, Not, Symbol, UnsatCoreSolver, is_unsat
+from pysmt.logics import QF_BOOL
+from pysmt.exceptions import SolverStatusError
+from pysmt.test.examples import get_example_formulae
+
+class TestUnsatCores(TestCase):
+
+    @unittest.skipIf(len(get_env().factory.all_unsat_core_solvers()) == 0,
```
marcogario: Add decorator to pysmt/test/__init__.py

pysmt/test/test_unsat_cores.py
```diff
@@ -0,0 +1,150 @@
...
+            r = s.solve()
+            self.assertFalse(r)
+            s.add_assertion(Symbol("y"))
+            with self.assertRaises(SolverStatusError):
+                s.get_unsat_core()
+
+
+    @skipIfSolverNotAvailable("msat")
+    def test_examples_msat(self):
+        for (f, _, satisfiability, logic) in get_example_formulae():
```
marcogario: Factor this out, in a method that takes the target solver. This will allow to unify test_examples_msat and test_examples_z3.

marcogario:
Looks good! Remember to update CHANGES.rst

pysmt/solvers/msat.py
mikand: This is done on purpose, to avoid double checking. In fact, the same function ```self._check_unsat_core_config()``` is called from within ```get_named_unsat_core()```

pysmt/solvers/msat.py
mikand: MathSAT natively supports "unnamed" unsat cores, for that case no assumprions are added (see ```add_assertion()```). We need to use solving under assumptions only if we are in the named case, in which case I add the assumption Boolean implicants to n_ass and I perform each call to solve under that assumptions (at least).

pysmt/factory.py
mikand: We should discuss this.
In practice we want to select a solver supporting unsat cosera AND configure it for unsat cores...

pysmt/solvers/z3.py
mikand: Not really, they assume that self is a IncrementalTrackingSolver wuile UnsatCoreSolver is an unrelated mixin

pysmt/solvers/msat.py
marcogario: ok.

pysmt/solvers/msat.py
marcogario: ok.

pysmt/solvers/z3.py
marcogario: ok, but I expect that for any UnsatCoreSolver you will be interested in running this check, i.e., verify that the solver is in a valid state. This might be a symptom that the current approach of extending solver interfaces is not good enough. Lets leave it like this, for now.

pysmt/shortcuts.py
marcogario: Done.

pysmt/factory.py
marcogario: TBD in #71

Issue #71, opened 2015-04-11
Factorize code for solver selection


marcogario:
The method Factory.get_unsat_core_solver duplicates the logic for solver selection from get_solver. Since that is a quite complex flow, we might think about factoring it and dividing it into two parts. 

The first part will be responsible for selecting the SolverClass according to the given options (logic, quantified, name etc.). The second part will be responsible for instantiating the solver with  adequate options.

PR #72, opened 2015-04-14
Quantifiler Elimination selector

mikand:
In this PR we introduce a logic-based selection of the quantifier eliminators, analogously to the Solver case. In addition the factory code has been re-factored.

This addresses issues #16 and #71.

### DIFF
--- a/pysmt/test/test_qe.py
+++ b/pysmt/test/test_qe.py
@@ -62,6 +62,9 @@ class TestQE(TestCase):
         with self.assertRaises(NoSolverAvailableError):
             QuantifierEliminator(name="msat", logic=LIA)
 
+    @skipIfNoQEForLogic(LRA)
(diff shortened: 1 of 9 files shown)


pysmt/test/test_qe.py
```diff
@@ -50,9 +51,19 @@ def test_qe_eq(self):
...
 
-    @unittest.skipIf('z3' not in get_env().factory.all_qelims(),
-                     "Z3 is not available.")
+        with self.assertRaises(NoSolverAvailableError):
+            QuantifierEliminator(name="nonexistent")
+
+        # MathSAT QE does not support LIA
+        with self.assertRaises(NoSolverAvailableError):
+            QuantifierEliminator(name="msat", logic=LIA)
+
```
marcogario: might it be that flow when the logic is not specified is missing?

Issue #47, opened 2015-03-13
Module operator is missing


marcogario:
Mod operator is missing

mikand:
The problem is that a quantifier elimination in LIA could produce a modular congruence.
We decided to go for an "incomplete" LIA quantifier elimination avoiding the MOD operator that is problematic for almost all the solvers (Only Z3 and partially MathSAT support it).

If a QE in LIA produces a MOD (using Z3) an exeption containing the Z3 expression will be returned

PR #73, opened 2015-04-14
Z3 lia qe error

mikand:
This PR implements error rising in Z3 quantifier eliminator when a formula containing an unsupported operator is found. This solves #47.

### DIFF
--- a/pysmt/exceptions.py
+++ b/pysmt/exceptions.py
@@ -67,3 +67,18 @@ class SolverStatusError(Exception):
     incorrectly called in the wrong status.
     """
     pass
+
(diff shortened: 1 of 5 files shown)


pysmt/exceptions.py
```diff
@@ -67,3 +67,18 @@ class SolverStatusError(Exception):
     incorrectly called in the wrong status.
     """
     pass
+
+class ConvertExperessionError(Exception):
```
marcogario: Typo: Experession/Expression

marcogario:
When you fix the typo, try to make a single commit out of these 3.

Issue #69, opened 2015-04-08
PicoSAT solving under assumptions


mikand:
Currently we use push/pop to support solving under assumptions in PicoSAT.
The PicoSAT API actually supports solving under assumptions, we should use that feature (similarly to MathSAT and Z3 solver wrappers)

PR #74, opened 2015-04-16
Added support for picosat assumptions

mikand:
This solves #69

### DIFF
--- a/pysmt/solvers/pico.py
+++ b/pysmt/solvers/pico.py
@@ -69,27 +69,48 @@ class PicosatSolver(Solver):
         # no need to declare variables
         pass
 
+    def _get_pico_lit(self, lit):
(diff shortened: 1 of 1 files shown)


pysmt/solvers/pico.py
```diff
@@ -69,27 +69,48 @@ def declare_variable(self, var):
...
-            self.add_assertion(self.mgr.And(assumptions))
-            self.pending_pop = True
+            cnf = []
+            for a in assumptions:
+                cnf += self.cnfizer.convert(a)
+
+            missing = []
+            for clause in cnf:
+                if len(clause) == 1:
+                    v = self._get_pico_lit(next(iter(clause)))
```
marcogario: Why is this not simply clause[0]?

Issue #34, opened 2015-03-07
Implement FNode.__len__()


marcogario:
Sometimes we want to get an idea of how big a formula is. Currently, the hackish way I do this is:
```
  len(str(f))
```

that tells me the length of the string representation of _f_ . This is not a good approach, since I need to build str(f) first and this could be expensive. We should implement the method FNode.__len__() to address this issue.

There are multiple metrics that one could consider, but I think the number of nodes of the underlying tree is a trivial one that makes sense.

marcogario:
Make this a FNode.size() and allow for different metrics as parameter. This should be implemented with a walker.

marcogario:
PR #75

PR #75, opened 2015-04-16
Fnode size

mikand:
Addressing #34 with dedicated oracle supporting four measures (for now) of the size of a formula.

### DIFF
--- a/pysmt/fnode.py
+++ b/pysmt/fnode.py
@@ -80,8 +80,8 @@ class FNode(object):
     def substitute(self, subs):
         return pysmt.shortcuts.substitute(self, subs=subs)
 
-    def size(self, counting_type=None):
-        return pysmt.shortcuts.get_formula_size(self, counting_type)
+    def size(self, measure=None):
+        return pysmt.shortcuts.get_formula_size(self, measure)
 
     def is_constant(self, _type=None, value=None):
         if self.node_type() not in CONSTANTS:
--- a/pysmt/fnode.py
+++ b/pysmt/fnode.py
@@ -80,6 +80,9 @@ class FNode(object):
     def substitute(self, subs):
         return pysmt.shortcuts.substitute(self, subs=subs)
 
+    def size(self, counting_type=None):
+        return pysmt.shortcuts.get_formula_size(self, counting_type)
+
     def is_constant(self, _type=None, value=None):
         if self.node_type() not in CONSTANTS:
             return False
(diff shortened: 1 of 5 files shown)


pysmt/fnode.py
```diff
@@ -80,6 +80,9 @@ def simplify(self):
     def substitute(self, subs):
         return pysmt.shortcuts.substitute(self, subs=subs)
 
+    def size(self, counting_type=None):
```
marcogario: s/counting_type/metric ?

pysmt/oracles.py
```diff
@@ -30,6 +30,73 @@
...
+     COUNTING_TYPE_LEAVES,
+     COUNTING_TYPE_DEPTH) = range(4)
+
+    def __init__(self, env=None):
+        walkers.DagWalker.__init__(self, env=env)
+
+        # Clear the mapping function
+        self.functions.clear()
+
+        # Propagate truth value, and force False when a Quantifier
```
marcogario: This comment seems to be outdated.

pysmt/oracles.py
```diff
@@ -30,6 +30,73 @@
...
+
+
+    def walk_count(self, formula, args, counting_type):
+        """ Returns the sizes of the formula """
+        if counting_type == SizeOracle.COUNTING_TYPE_TREE_NODES:
+            return 1 + sum(args)
+
+        if counting_type == SizeOracle.COUNTING_TYPE_DAG_NODES:
+            return frozenset([formula]) | frozenset([x for s in args for x in s])
+
```
marcogario: style: dueble newline

pysmt/oracles.py
```diff
@@ -30,6 +30,73 @@
...
+        # Check that no operator in undefined
+        assert self.is_complete(verbose=True)
+
+
+    def _get_key(self, formula, counting_type):
+        # Memoize using a tuple (counting_type, formula)
+        return (counting_type, formula)
+
+
+    def get_size(self, formula, counting_type=None):
```
marcogario: Each time this function is called, we only use one of the possible "counting_types"/metric. Therefore, it seems inconvenient to have to check which counting_type we are using within walk_count for every node.
I'd suggest that this method sets the value of walk_count.

```
def get_size(...):
  self.set_walk_method(counting_type)
  res = self.walk(formula)
  self.walk_count = self.walk_error
```

where self.set_walk_method assigns the appropriate function to all node types (as done during init())

PR #84, opened 2015-04-28
BV: Added Several examples

marcogario:
- Added several examples for BVs. 
- This extends also the substitutuer, the simplifier and the theory oracle with BVs
- Enforcing positive BV Number at construction time


pysmt/test/test_bv_simplification.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def check(self, f):
+        simp_f = f.simplify()
+        msat_f = self.msat_simplify(f)
+        self.assertEqual(msat_f, simp_f)
+
+
+    @skipIfSolverNotAvailable("msat")
+    def test_all_binary(self):
+        for l in self.all_bv_numbers():
```
marcogario:
```
itertools.product(self.all_bv_numbers(), self.all_bv_numbers(), self.bin_operators)
```

PR #90, opened 2015-05-05
BDD: iterative back-conversion 

mikand:
Removed recursive non-memoized back-conversion in BDD solver and implemented a purely-iterative, memoized version.

### DIFF
--- a/pysmt/solvers/bdd.py
+++ b/pysmt/solvers/bdd.py
@@ -280,6 +280,9 @@ class BddConverter(Converter, DagWalker):
         with LockDdManager(self.ddmanager):
             while len(stack) > 0:
                 current = stack.pop()
+                # If the node haven't been explored yet, it is not in back_memoization,
+                # If back_memoization contains None, we are exploring the children
+                # Otherwise, it contains the pySMT expression corresponding to the node
                 if current not in self.back_memoization:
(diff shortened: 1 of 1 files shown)


pysmt/solvers/bdd.py
```diff
@@ -273,79 +273,52 @@ def walk_bool_constant(self, formula, args):
...
-                res = self.fmgr.Or(self.fmgr.And(var, nt),
-                                   self.fmgr.And(self.fmgr.Not(var), ne))
-            else:
-                t = self.bdd_to_expr3(bdd.T(), invert)
-                e = self.bdd_to_expr3(bdd.E(), invert)
-                res = self.fmgr.Or(self.fmgr.And(var, t),
-                                   self.fmgr.And(self.fmgr.Not(var), e))
-        return res
+        with LockDdManager(self.ddmanager):
+            while len(stack) > 0:
```
marcogario: Add a comment on the semantics of back_memoization:
- If the node is not there, we haven't explored yet,
- If it is None, we are exploring the children
- Otherwise, it contains the pySMT expressoin

marcogario:
Great!

mikand:
I updated the PR in 7917c82bfc950589c40e6ccf0045fe2d8b633547 adding the missing comment

PR #97, opened 2015-05-07
BUGFIX: Fixed an assertion error arising when iterating over a Z3Model

mikand:
The problem was caused by a missing conversion from decl to term


mikand:
Wait a moment before accepting, i found a minor glitch

mikand:
Done.

Issue #89, opened 2015-05-04
BDD ordering support


mikand:
We shall consider adding support for static BDD ordering specification as well as dynamic reordering.
@cmattarei could be interested in this feature.

mikand:
From a preliminary analysis it seems that is sufficient to use ```NewVarAtLevel``` instead of ```NewVar``` in order to create variables in a pre-determined order. Otherwise variables are appended at the bottom. An alternative possibility is to pre-create all the variables in the desired order.

mikand:
56c8ec0d73483 gives a first shot at this issue.

marcogario:
PR #100

PR #100, opened 2015-05-07
BDD: Added static ordering support and dynamic reordering for BDDs

mikand:
Added support for dynamic reordering and specification of an initial ordering for BddSolver.

This addresses #89

### DIFF
--- a/pysmt/solvers/bdd.py
+++ b/pysmt/solvers/bdd.py
@@ -23,12 +23,13 @@ import pycudd
 import pysmt.logics
 
 from pysmt import typing as types
-from pysmt.solvers.solver import Solver, Converter
+from pysmt.solvers.solver import Solver, Converter, SolverOptions
 from pysmt.solvers.eager import EagerModel
 from pysmt.walkers import DagWalker
 from pysmt.decorators import clear_pending_pop
 from pysmt.oracles import get_logic
 from pysmt.solvers.qelim import QuantifierEliminator
+
 
 class LockDdManager(object):
     """Context class that must be used to guard any usage of pycudd. This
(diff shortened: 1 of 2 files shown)


pysmt/solvers/bdd.py
```diff
@@ -54,13 +55,56 @@ def __enter__(self):
...
+            LockDdManager._depth -= 1
+            if LockDdManager._depth == 0:
+                pycudd.ResetDefaultDdManager()
+
+
+
+class BddOptions(SolverOptions):
+
+    ## CUDD Reordering algorithms
+    CUDD_REORDER_SAME, \
```
marcogario: Indentation.

pysmt/solvers/bdd.py
```diff
@@ -54,13 +55,56 @@ def __enter__(self):
...
+        CUDD_REORDER_WINDOW3_CONV, \
+        CUDD_REORDER_WINDOW4_CONV, \
+        CUDD_REORDER_GROUP_SIFT, \
+        CUDD_REORDER_GROUP_SIFT_CONV, \
+        CUDD_REORDER_ANNEALING, \
+        CUDD_REORDER_GENETIC, \
+        CUDD_REORDER_LINEAR, \
+        CUDD_REORDER_LINEAR_CONVERGE, \
+        CUDD_REORDER_LAZY_SIFT, \
+        CUDD_REORDER_EXACT = range(22)
```
marcogario: This is not defined anywhere by pycudd?

pysmt/solvers/bdd.py
```diff
@@ -87,6 +148,22 @@ def __init__(self, environment, logic, user_options):
         self.backtrack = []
         self.latest_model = None
 
+    def get_default_options(self, logic=None, user_options=None):
+        res = BddOptions()
+        if user_options is not None:
+            if "unsat_cores_mode" in user_options:
```
marcogario: So, the proper way to create this is {"unsat_cores_mode": None}? Is this because of how the factory works? Wouldn't it be better just to assert that "unsat_cores_mode" not in user_options?

pysmt/solvers/bdd.py
```diff
@@ -87,6 +148,22 @@ def __init__(self, environment, logic, user_options):
...
+                if user_options["unsat_cores_mode"] != None:
+                    raise NotImplementedError("BddSolver does not "\
+                                              "support unsat cores")
+
+            for k in ["static_ordering",
```
marcogario: Maybe you can post-pone the initialization of BddOptions till here and do the unpacking of keywords arguments? If they are not defined, you will take the default
```
 BddOptions(**user_options)
```
I think it is nicer and less error-prone than *setattr*

pysmt/test/test_bdd.py
```diff
@@ -115,6 +122,41 @@ def test_quantifier_eliminator(self):
         g = qelim(f, solver_name="bdd")
         self.assertEqual(g, self.y)
 
+    @skipIfSolverNotAvailable("bdd")
+    def test_reordering(self):
+        from pysmt.solvers.bdd import BddSolver
```
marcogario: We should extend the factory to be able to pass user_options to the solver. Thus this would simply be a call to the factory *Solver* .

pysmt/test/test_bdd.py
```diff
@@ -115,6 +122,41 @@ def test_quantifier_eliminator(self):
...
+            self.assertTrue(s.solve())
+
+    @skipIfSolverNotAvailable("bdd")
+    def test_reordering_algorithms(self):
+        from pysmt.solvers.bdd import BddSolver, BddOptions
+        for algo in BddOptions.CUDD_ALL_REORDERING_ALGORITHMS:
+            with BddSolver(get_env(), pysmt.logics.BOOL,
+                           {"dynamic_reordering" : True,
+                            "reordering_algorithm" : algo}) as s:
+                s.add_assertion(self.big_tree)
```
marcogario: Maybe query the ddmanager and check that the current reordering is the one specified.

pysmt/test/test_bdd.py
```diff
@@ -115,6 +122,41 @@ def test_quantifier_eliminator(self):
...
+                self.assertTrue(s.solve())
+
+    @skipIfSolverNotAvailable("bdd")
+    def test_fixed_ordering(self):
+        from pysmt.solvers.bdd import BddSolver
+        with BddSolver(get_env(), pysmt.logics.BOOL,
+                       {"static_ordering" : [self.x, self.y]}) as s:
+            s.add_assertion(self.big_tree)
```
marcogario: Add case in which ordering is invalid (i.e., non-bool, or strings).

pysmt/test/test_bdd.py
```diff
@@ -115,6 +122,41 @@ def test_quantifier_eliminator(self):
...
+            self.assertTrue(s.solve())
+
+    @skipIfSolverNotAvailable("bdd")
+    def test_initial_ordering(self):
+        from pysmt.solvers.bdd import BddSolver
+        with BddSolver(get_env(), pysmt.logics.BOOL,
+                       {"static_ordering" : [self.x, self.y],
+                        "dynamic_reordering" : True}) as s:
+            s.add_assertion(self.big_tree)
```
marcogario: How can we test that this is really working? This test needs to better capture the semantic of this operation.

pysmt/solvers/bdd.py
marcogario: no, seems like no. This is a bit fragile, but until we have a better wrapper for cudd, this is ok.

pysmt/test/test_bdd.py
mikand: This is an orthogonal issue. See #105

Issue #92, opened 2015-05-06
issues with infix notation


agriggio:
It seems that the infix notation has still some rough edges.
Some examples:
From the shell.py, using `get_env().enable_infix_notation=True`

```
x = Symbol('x', INT)
f = (x > Int(10))
```

works, but:

```
f = (x > 10)
```

raises an exception `AttributeError: 'int' object has no attribute
'get_sons'`.

Also:

```
f = (x == Int(10))
```

does not do what one would expect (i.e. it checks whether `x` and
`Int(10)` are equal, instead of building the equality formula).

agriggio:
also, it would be nice to have infix notation for Boolean connectives &, |, ~

marcogario:
Thanks for reporting.  In case it helps, you can use f.And() etc for
boolean infix operators.

agriggio:
Yes, but it doesn't help :-) `f.And(f2)` is worse than `And(f, f2)`.
Infix notation is really handy in the shell, but I can live without it for more serious work.
However, automatic promotion of numbers would be nice to have (i.e. allow writing `Equals(x, 100)` instead of `Equals(x, Int(100)`)

mikand:
There is a bit of ambiguity for ```x == 5``` it could be either ```x == Real(5)``` or ```x == Int(5)``` because PySMT distinguishes reals and integers.

This could be solved by forcing to use ```5.0``` for real numbers, but I'm a bit scared by numeric stability...

mikand:
@marcogario: We decided to limit the infix notation for this and other issues (Boolean operators are not supported because of precedence issues and confusion between conjunction and bvand).
I'm for the "explicit is better than implicit", but since infix notation is optional we can go all-the-way with it... What do you think?

agriggio:
Why is `x == 5` ambiguous? Isn't the type of `x` enough to resolve the ambiguity?
Regarding explicit vs implicit, I agree, but the infix notation is handy for playing around, but much less fun if it is only halfway... (Btw, if you really want to push it, `a & b` could mean both `And(a,b)` and  `BVand(a, b)`, depending on the argument types :-)

mikand:
In cf8227f I gave a first shot at this issue:
I added &, |, ^, ~, -(unary) and the support for numeric constants. (Still missing BV overloads and casts)

The only operator I think is REALLY problematic is equality (==) as it is used (together with __hash__) when a formula is stored in a set or in a dict, so we cannot overload it....

[...]

PR #108, opened 2015-05-08
FNODE: Extended support to infix notation as per #92.

mikand:
Added support for the follwing infix operators:
* ```x & y```
* ```x | y```
* ```~x```
* ```-x```
* ```int(x)```
* ```float(x)```
* ```long(x)```

And added support for automatic propotion of python constants to pysmt constants:
E.g.
```python
1.5 + Symbol(x, Real) <= 8
```

### DIFF
--- a/pysmt/fnode.py
+++ b/pysmt/fnode.py
@@ -37,6 +37,7 @@ from pysmt.operators import (FORALL, EXISTS, AND, OR, NOT, IMPLIES, IFF,
                              BV_OPERATORS)
 from pysmt.typing import BOOL, REAL, INT, PYSMT_TYPES, BVType
 from pysmt.decorators import deprecated
+from pysmt.utils import is_python_integer, is_python_rational, is_python_boolean
 
 FNodeContent = collections.namedtuple("FNodeContent",
                                       ["node_type", "args", "payload"])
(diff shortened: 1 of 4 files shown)


pysmt/fnode.py
```diff
@@ -402,4 +425,45 @@ def __lt__(self, right):
...
+        return self._apply_infix(-1, pysmt.shortcuts.Times)
+
+    def __invert__(self):
+        env = pysmt.shortcuts.get_env()
+        if not env.enable_infix_notation:
+            raise Exception("Cannot use infix notation")
+        return pysmt.shortcuts.Not(self)
+
+    def __int__(self):
+        if self.is_int_constant():
```
marcogario: This is enable independently of infix_notation?

pysmt/test/test_formula.py
```diff
@@ -623,6 +624,46 @@ def test_infix(self):
...
+        self.assertEqual(And(x, y), x & y)
+        self.assertEqual(Or(x, y), x | y)
+
+        self.assertEqual(Or(x, TRUE()), x | True)
+        self.assertEqual(Or(x, TRUE()), True | x)
+
+        self.assertEqual(And(x, TRUE()), x & True)
+        self.assertEqual(And(x, TRUE()), True & x)
+
+        get_env().enable_infix_notation = False
```
marcogario: Ok. But this is not necessary. The environment is reset from one test to the next.

pysmt/fnode.py
mikand: Yes, my bad!

Issue #94, opened 2015-05-06
feature request: support for get_implicant()


agriggio:
It would be nice to have a `get_implicant(f)` function, that returns an
SMT-level implicant for a satisfiable formula `f` (aka a T-consistent truth
assignment for the atoms of `f`, aka a "propositional" model that is
T-consistent).
Alternatively, what's the easiest way of achieving this?

mikand:
This should be what you are looking for.

```python
def get_implicant(formula, solver_name=None,
                  logic=None, complete=False):
    model = get_model(formula, solver_name=solver_name, logic=logic)
    if model is None:
        return None
    else:
        def assignment_as_formula(var, val):
            if val.is_true():
                return var
            elif val.is_false():
                return Not(var)
            else:
                return Equals(var, val)

        if complete:
            return And(assignment_as_formula(x, model.get_value(x))
                       for x in formula.get_free_variables())
        else:
            # Partial model -> more general implicant
            return And(assignment_as_formula(x, v) for x,v in model)
```

mikand:
I opened a pull request #95 for a proper implementation of ```get_implicant``` in pysmt.

mikand:
@agriggio I just realized that this might not be what you were looking for... 
for a formula like ```( x | y ) & (a < 14 | a > 18)``` what is a possible "implicant"?

agriggio:
Yes, the above is not what I want :-)
For the example, a (prime) implicant would be `x & (a < 14)`, or a (non-prime) one `x & !y & (a > 18) & !(a < 14)`

marcogario:
PR #109

PR #109, opened 2015-05-08
Get implicant

mikand:
Implemented support of the ```get_implicant``` shortcut as described in #94


pysmt/factory.py
```diff
@@ -441,6 +441,31 @@ def get_model(self, formula, solver_name=None, logic=None):
...
+        if logic == AUTO_LOGIC:
+            logic = get_logic(formula, self.environment)
+
+        with self.Solver(name=solver_name, logic=logic) \
+             as solver:
+            solver.add_assertion(formula)
+            check = solver.solve()
+            if not check:
+                return None
+            else:
```
marcogario: I think it could be nice to move this function inside the Solver interface. We should avoid having too much logic in the Factory.

pysmt/factory.py
```diff
@@ -441,6 +441,31 @@ def get_model(self, formula, solver_name=None, logic=None):
...
+                model = solver.get_model()
+                atoms = formula.get_atoms()
+                res = []
+                for a in atoms:
+                    fv = a.get_free_variables()
+                    if any(v in model for v in fv):
+                        if solver.get_value(a).is_true():
```
marcogario: get_py_value ?

pysmt/oracles.py
```diff
@@ -327,6 +329,75 @@ def walk_function(self, formula, args):
...
+        return frozenset(x for a in args for x in a)
+
+    def walk_theory_relation(self, formula, args):
+        return frozenset([formula])
+
+    def walk_theory_op(self, formula, args):
+        return None
+
+    def walk_symbol(self, formula, args):
+        if formula.is_symbol(types.BOOL):
```
marcogario: This is returning also boolean symbols, but the description said that it returns theory atoms.

pysmt/test/test_solving.py
```diff
@@ -122,6 +122,38 @@ def test_get_model_sat(self):
...
+        varA = Symbol("A", BOOL)
+        varB = Symbol("B", BOOL)
+
+        f = And(varA, Not(varB))
+        g = f.substitute({varB:varA})
+
+        res = get_implicant(g, logic=QF_BOOL)
+        self.assertIsNone(res, "Formula was expected to be UNSAT")
+
+        for solver in get_env().factory.all_solvers(logic=QF_BOOL):
```
marcogario: Why this is done twice?

pysmt/test/test_solving.py
```diff
@@ -122,6 +122,38 @@ def test_get_model_sat(self):
...
+            self.assertIsNone(res, "Formula was expected to be UNSAT")
+
+    @skipIfNoSolverForLogic(QF_LRA)
+    def test_get_implicant_sat(self):
+        varA = Symbol("A", BOOL)
+        varX = Symbol("X", REAL)
+
+        f = And(varA, Equals(varX, Real(8)))
+
+        res = get_implicant(f, logic=QF_LRA)
```
marcogario: Same here. This is repeated below. Since the test is decorated with skipIfNoSolverForLogic, we are guaranteed that the for below will be entered.

marcogario:
Good!
Note: Tests for the atomOracle are missing

pysmt/factory.py
mikand: Which would be the semantics of a get_implicant at solver level? Note that the semantics "implicant of all the asserted formulae" would require us to keep the set of assertions in all the solvers...
Or was you suggesting a static method of Solver?

pysmt/factory.py
mikand: Do you like it better for some reason?

pysmt/factory.py
marcogario: I was thinking more of a static method (although it does not need to be static, we can just require that no assertion has been done yet). In my opinion, factory should only wrap methods of Solver, providing good defaults. 
However, this is not particularly blocking. If you see a nice way to do this, go ahead. Otherwise, we can do it later.

pysmt/factory.py
marcogario: I don't have strong preferences. However, it seems like we never use get_py_value. Maybe we should just deprecate it?

mikand:
For me this is OK, whenever you wnat you can comment or merge... :)

PR #117, opened 2015-05-20
BUGFIX: Fixed a bug when using ExactlyOne with a generator as argument

marcogario:


pysmt/formula.py
```diff
@@ -464,6 +464,7 @@ def ExactlyOne(self, *args):
            A -> !(B \/ C)
            B -> !(C)
         """
+        args = list(*args)
```
mikand: This works, but if args is a list you double the memory...
```python
if isinstance(args, types.GeneratorType) args = list(*args)
```
could be better.

But we never do this kind of trick, hence I guess it is OK

Issue #113, opened 2015-05-15
Provide a deterministic unique id to fnodes at creation time


marcogario:

mikand:
Is this for deterministic executions? Shall we broaden this issue by considering all the usages of sets and dictionaries?

marcogario:
PR #119

PR #119, opened 2015-05-21
Changed FNode's hash to be deterministic among succesive executions

mikand:
Changed FNode's hash to be deterministic among succesive executions and on different machines. This is done by adding an id to each FNode
that is unique within the Formula Manager it belongs to.

This addresses issue #113.


mikand:
OK, finally CI seems to work :)

PR #121, opened 2015-05-22
Ported pysmt from pycudd to repycudd

mikand:
Ported pysmt BDD solving features to use repycudd version of pycudd, thus having a re-entrant version of the library


marcogario:
Good!
Update README and CHANGES, to make it clear that we are wrapping repycudd and not pycudd.

mikand:
If you can, put online a mirror for CI, so that we can avoid the hack I added in install.py

marcogario:
@mikand, done. Please squash this PR.

mikand:
Done

PR #122, opened 2015-05-24
CI: Appveyor support

marcogario:
Added support for running tests on Appveyor. This provides some basic CI for windows.
NOTE: Only the library tests are running. No solver is currently being used.

This also creates a directory ci/ in the root of the project, where to store all auxiliary material for running CI, e.g., shippable.sh has been moved there.


mikand:
Nice! :)

Issue #102, opened 2015-05-08
make FNodes and walkers easier to use


agriggio:
It seems that FNodes have a lot of hardcoded information, that makes
using them somewhat inconvenient. For instance, I could not understand
how to retrieve the sort of a node in a generic way, without using big
"switch statements" over the node types (which are tedious and will
break as soon as some other node types are added). For symbols, there's
`FNode.symbol_type()`, but what about other nodes? For example, how do I
know that an `AND` is of Bool sort without knowing that it is an `AND`
operator?

Related to this, the use of walkers seems also quite verbose. From what
I understand, one needs to implement the various `walk_*` functions for
all the different nodes, because examining the `node_type()` is the only
way of retrieving all the information needed to properly operate on
formulas.

For a concrete example, I was trying to implement a walker to extract
all the atoms in the formula. I was trying to do something like this:

```
class GetAtomsWalker(pysmt.walkers.dag.DagWalker):
    def __init__(self):
        super(ImplicantWalker, self).__init__()
        for k in self.functions:
            self.functions[k] = self.walk_get_atom
        self.atoms = []

    def walk_get_atom(self, formula, **kwargs):
        if formula.is_quantifier():
            raise NotImplementedError("get_atoms for quantified formulae")
        
        if formula.symbol_type().is_bool_type():
            if not formula.is_boolean_operator():
                self.atoms.append(formula)
```

But this does not work, because `formula.symbol_type()` only works if
`formula.is_symbol()` is true, otherwise it fails (horribly :-)

marcogario:
For the first point, is using the operator groups in pysmt.operators enough for your need? Do you think it would be better to implement them in the FNode. E.g., FNode.is_bool_op() ? The implementation would simply embed the current approach:
```
def is_bool_op(self):
  return self.node_type() in operators.BOOL_OPERATORS
```

For what concerns the Walkers. The idea is exactly to set the functions map in order to group together nodes that need to be handled in the same way. Having solved the issue above, do you still see problems with this approach?

Thanks for the feedback!

agriggio:
Operator groups certainly help, but they still don't solve the original
issue, that is how do I retrieve the sort of a node without explicitly
checking what is its `node_type`. Maybe you could consider adding a
method `FNode.get_sort()`, that would do something like this:

```
def get_sort(self):
    if not self.is_term():
        raise Exception("not a term")
    if self.is_bool_op():
        return BOOL
    elif self.is_int_constant():
        return INT
    ...
[...]

PR #139, opened 2015-06-13
Walkers: added method to set the walking function 

mikand:
This addresses #102

### DIFF
--- a/pysmt/walkers/generic.py
+++ b/pysmt/walkers/generic.py
@@ -72,9 +72,19 @@ class Walker(object):
         self.functions[op.BV_ZEXT] = self.walk_bv_zext
         self.functions[op.BV_SEXT] = self.walk_bv_sext
 
-        assert self.is_complete()
+        undefined_types = set(op.ALL_TYPES) - set(self.functions.keys())
+        assert len(undefined_types) == 0, \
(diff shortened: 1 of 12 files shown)


marcogario:
What shall we do with calls to ```self.functions.clear()```? This refactoring shows that sub-classes do not need to access ```functions``` directly. Thus, the call to ```clear()``` needs to be wrapped.

The goal of using clear() and _is_complete, was to check that a new method had been defined for each node_type. However, this might be the wrong way of looking at this. We should rely on **testing** to guarantee that not only a method for each node has been defined, but also that each method makes sense.

Shall we remove the calls to ```clear()``` and deprecate ```is_complete()```? Can we do it in this branch?

pysmt/walkers/generic.py
```diff
@@ -76,6 +77,20 @@ def __init__(self, env=None):
 
         return
 
+
+    def set_function(self, function, node_types):
```
marcogario: What about using ```*node_types```?

mikand: OK

mikand:
@marcogario I agree that testing should look into this, but the is_complete method allows to catch subtle bugs early: it has the value of an assertion. Not sure if we really want to deprecate it...

PR #145, opened 2015-06-20
BV: Add support for CVC4

marcogario:


marcogario:
Note: This is branched out of z3, to use the new tests introduced there. Merging with master should not create problems _if_ done after bv_z3.

PR #148, opened 2015-06-20
Deprecate rand

marcogario:


marcogario:
*Note*: This builds on top of branch bv_z3 to use Z3 to test simplifications on all examples.

pysmt/simplifier.py
```diff
@@ -386,33 +368,33 @@ def walk_bv_neg(self, formula, args, **kwargs):
         return self.manager.BVNeg(args[0])
 
     def walk_bv_or(self, formula, args, **kwargs):
-        if args[0].is_bv_constant() and args[1].is_bv_constant:
+        if args[0].is_bv_constant() and args[1].is_bv_constant():
```
mikand: Ops... :)

PR #154, opened 2015-06-26
Smart Printing

marcogario:
Smart Printing is meant to provide a way of printing complex expressions using aliases.

Two typical use cases are non-primitive operators and DEFINEs (SMV-style).
For the first case, I provide an example in test_printing.py : `ExactlyOne(x1, ..., xn)` is translated into a disjunction and a set of mutual exclusion constraints. Using smart printing, we can print it as `ExactlyOne(x1, ..., xn)`. 

For now the integration with the rest of the system is weak. You need to manually create the 'smarties' (i.e., the map formula -> string) and call smart_serialize. I think we could:

- Integrate this in FNode.serialize() : If no smarties are provided, SmartPrinter falls-back to the HRPrinter
- Generate the smarties when building expressions, and have a global dictionary of smarties.

### DIFF
--- a/pysmt/printers.py
+++ b/pysmt/printers.py
@@ -22,251 +22,230 @@ from six.moves import cStringIO
 
 
 class HRPrinter(TreeWalker):
+    """Performs serialization of a formula in a human-readable way.
+
+    E.g., Implies(And(Symbol(x), Symbol(y)), Symbol(z))  ~>   '(x * y) -> z'
+    """
 
     def __init__(self, stream):
         TreeWalker.__init__(self)
         self.stream = stream
-        return
-
+        self.write = self.stream.write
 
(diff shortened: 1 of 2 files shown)


pysmt/printers.py
```diff
@@ -356,3 +326,71 @@ def serialize(self, formula, printer=None, threshold=None):
...
+
+
+class SmartPrinter(HRPrinter):
+    """Better serialization allowing special printing of subformula.
+
+    The formula is serialized according to the format defined in the
+    HRPrinter. However, everytime a formula that is present in
+    'smarties' is found, this is replaced.
+
+    E.g., smarties = {And(a,b): "ab"}
```
mikand: "smarties" is a TM, change the name...

PR #156, opened 2015-06-27
Signed bitvectors

mikand:
Added support for signed operators in the BV theory


pysmt/test/examples.py
```diff
@@ -306,6 +307,100 @@ def get_example_formulae(environment=None):
...
+                    logic=pysmt.logics.QF_BV),
+
+            # 0 s< bv16 & bv16 a>> 1 = bv16 >> 1
+            Example(expr=And(BVSLT(BVZero(16), bv16),
+                             Equals(BVAShr(bv16, BVOne(16)),
+                                    BVLShr(bv16, BVOne(16)))),
+                    is_valid=False,
+                    is_sat=True,
+                    logic=pysmt.logics.QF_BV),
+
```
marcogario: No creation of negative bitvector value? BV(-1) ?

pysmt/solvers/z3.py
```diff
@@ -40,6 +40,47 @@
 
 # patch z3api
 z3.is_ite = lambda x: z3.is_app_of(x, z3.Z3_OP_ITE)
```
marcogario: Ok for now, but we need to find a better solution to integrate these up-stream.

pysmt/test/test_back.py
```diff
@@ -50,34 +47,35 @@ def test_msat_back_simple(self):
...
 
+    def do_back(self, solver_name):
+        for formula, _, _, logic in get_example_formulae():
+            if logic.quantifier_free:
+                try:
+                    s = Solver(name=solver_name, logic=logic)
+                    term = s.converter.convert(formula)
+                    res = s.converter.back(term)
+                    self.assertValid(Iff(formula, res), logic=logic)
+                except NoSolverAvailableError:
```
marcogario: The exception is because the logic might not be supported by the solver?

marcogario:
There are only some minor points. In particular, I don't remember if we took any decision on whether *BV(-1, 32)* should be supported. I think that would be useful.

After that, squash and I'll accept it.

pysmt/test/test_back.py
mikand: Exactly

pysmt/test/examples.py
mikand: This is an interesting point. MathSAT takes the position of forbidding negative-valued BV constants.
I think it is better to avoid overloding the ```BV(val, width)``` shortcut allowing negative values.
Consider the following case:
* ```BV(3, 2)``` This is ok and results in "11" if the bitvecor is considered unsigned, but it is an overflow error if we consider it signed

We could either add a parameter to ```BV()``` or (better) create a new shortcut ```SignedBV()```. Note that we can already create negative numbers by simply negating a positive value: ```BVNeg(BV(1,5))```.

marcogario: +1 for SignedBv, maybe SBv?

PR #169, opened 2015-08-08
PARSER: Extensibility

mikand:
- Refactoring of the parser to allow extensibility
- a few minor hacks to gain performance (~2 seconds executing all the tests in tests/smtlib)

### DIFF
--- a/pysmt/smtlib/parser.py
+++ b/pysmt/smtlib/parser.py
@@ -20,6 +20,7 @@ from fractions import Fraction
 from warnings import warn
 from six import iteritems
 from six.moves import xrange
+import itertools
 
 import pysmt.smtlib.commands as smtcmd
 from pysmt.environment import get_env
@@ -28,7 +29,7 @@ from pysmt.logics import get_logic_by_name, UndefinedLogicError
 from pysmt.exceptions import UnknownSmtLibCommandError
 from pysmt.smtlib.script import SmtLibCommand, SmtLibScript
 from pysmt.smtlib.annotations import Annotations
-import pysmt.utils as utils
+from pysmt.utils import interactive_char_iterator
 
 def get_formula(script_stream, environment=None):
     """
@@ -64,7 +65,7 @@ def get_formula_strict(script_stream, environment=None):
 
 def get_formula_fname(script_fname, environment=None, strict=True):
     """Returns the formula asserted at the end of the given script."""
-    with utils.BufferedTextReader(script_fname) as script:
+    with open(script_fname) as script:
         if strict:
             return get_formula_strict(script, environment)
         else:
@@ -104,9 +105,10 @@ class SmtLibExecutionCache(object):
             if len(parameters) == 0:
                 return expression
             return self._define_adapter(parameters, expression)
-        if name in self.keys:
-            if len(self.keys[name]) > 0:
-                return self.keys[name][-1]
+        elif name in self.keys:
+            lst = self.keys[name]
+            if len(lst) > 0:
+                return lst[-1]
             else:
                 return None
         else:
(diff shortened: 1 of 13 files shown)


pysmt/smtlib/parser.py
```diff
@@ -28,7 +29,7 @@
 from pysmt.exceptions import UnknownSmtLibCommandError
 from pysmt.smtlib.script import SmtLibCommand, SmtLibScript
 from pysmt.smtlib.annotations import Annotations
-import pysmt.utils as utils
+from pysmt.utils import interactive_char_iterator
```
marcogario: ```interactive_char_iterator``` is only used in the parser. It makes sense to put it here, instead of in ```utils```

pysmt/smtlib/parser.py
```diff
@@ -123,153 +125,159 @@ def unbind_all(self, values):
...
+    else:
+        reader = interactive_char_iterator(handle)
+    c = next(reader)
+
+    eof = False
+    while not eof:
+        if c in specials:
+            # consume all the spaces
+            if c in spaces:
+                while c and c in spaces:
```
marcogario: This loop can be moved at the beginning of the outer ```while```, right?

```
while not eof:
    while c and c in spaces:
         c = next(reader)
```

If this is the case, we can probably replace this with an ```ifilter```

pysmt/smtlib/parser.py
```diff
@@ -392,7 +400,7 @@ def _get_basic_type(self, type_name, params=None):
     def _get_var(self, name, type_name, params=None):
         """Returns the PySMT variable corresponding to a declaration"""
         typename = self._get_basic_type(type_name, params)
-        return self._current_env.formula_manager.Symbol(name=name,
+        return self.env.formula_manager.Symbol(name=name,
```
marcogario: Maybe we should use ```FormulaManager.get_symbol()```. The difference being that ```get_symbol()``` will return ```None``` if the symbol is not defined, while ```Symbol``` will declare it.

pysmt/smtlib/parser.py
```diff
@@ -509,99 +490,120 @@ def _handle_annotation(self, pyterm, *attrs):
...
+            typename = self.parse_type(tokens, "expression")
+
+            var = self._get_var(vname, typename)
+            self.cache.bind(vname, var)
+            vrs.append(var)
+
+            self.consume_closing(tokens, "expression")
+            current = next(tokens)
+
+
```
marcogario: remove \n

pysmt/typing.py
```diff
@@ -65,7 +65,9 @@ def __eq__(self, other):
         return self.type_id == other.type_id
 
     def __ne__(self, other):
-        return not (self == other)
```
marcogario: Why this change? In general, this is not good, since we now require that any sub-class that modifies ```__eq__``` also  modifies ```__ne__```. See ```_BVType```.

marcogario:
Lets add a test in which we extend the parser with some new key-word and function.

pysmt/smtlib/parser.py
mikand: This is used when a declare-fun is found, hence we really want to obtain the symbol

pysmt/typing.py
mikand: The profiler shows this is way more efficient in this way...

pysmt/smtlib/parser.py
mikand: OK

pysmt/smtlib/parser.py
mikand: OK

pysmt/smtlib/parser.py
mikand: I think this is a general-purpose function, do we really want it here?

marcogario: I think so, since we only use it in the context of parsing.

pysmt/smtlib/parser.py
marcogario: ok.

pysmt/typing.py
marcogario: Ok, lets fix ```_BVType``` then.

PR #171, opened 2015-08-23
Boolector Solver

marcogario:
Boolector is now supporte by pySMT with the name 'btor'.


pysmt/solvers/btor.py
```diff
@@ -0,0 +1,303 @@
...
+    LOGICS = [QF_BV]
+
+    def __init__(self, environment, logic, user_options):
+        IncrementalTrackingSolver.__init__(self,
+                                           environment=environment,
+                                           logic=logic,
+                                           user_options=user_options)
+
+        self.btor = boolector.Boolector()
+        self.btor.Set_opt("model_gen", 1)
```
marcogario: To use assumptions we need to enable incremental interface.

pysmt/solvers/btor.py
```diff
@@ -0,0 +1,303 @@
...
+        else:
+            raise SolverReturnedUnknownResultError
+
+    def get_unsat_core(self):
+        raise NotImplementedError
+
+    @clear_pending_pop
+    def _push(self, levels=1):
+        # TODO: Incrementality can be achieved by keeping copies of the boolector instace.
+        # See self.btor.Clone()
```
marcogario: What is the impact of Clone on the converter?

pysmt/solvers/btor.py
```diff
@@ -0,0 +1,303 @@
...
+    def walk_bv_mul(self, formula, args, **kwargs):
+        return args[0]*args[1]
+
+    def walk_bv_udiv(self, formula, args, **kwargs):
+        return args[0] / args[1]
+
+    def walk_bv_urem(self, formula, args, **kwargs):
+        return self._btor.Urem(args[0], args[1])
+
+    def walk_bv_lshl(self, formula, args, **kwargs):
```
marcogario: Check if this is part of the SMT-LIB or specific to Boolector.
```BoolectorException: [pybtor] Bit width of operand 'b' must be equal to log2(bit width of a)```

**Comment by marcogario** [comment]
:fireworks:

File: pysmt/solvers/btor.py [hunk 4]
```diff
@@ -0,0 +1,347 @@
...
+        self._btor = btor
+        self.declared_vars = {}
+        return
+
+    @catch_conversion_error
+    def convert(self, formula):
+        return self.walk(formula)
+
+    def back(self, expr):
+        titem = self.convert(expr)
```
mikand: This is not part of ```back``` but of ```get_value```. Back shall take directly the btor representation and return the pysmt representation.

pysmt/solvers/btor.py
```diff
@@ -0,0 +1,347 @@
...
+        if expr.is_symbol() and  expr.symbol_type().is_bv_type():
```
mikand: you should write predicates on titem, if you cannot reconstruct the formula, an exception shall be thrown.

pysmt/solvers/btor.py
```diff
@@ -0,0 +1,347 @@
...
+    def walk_bool_constant(self, formula, **kwargs):
+        return self._btor.Const(formula.constant_value())
+
+    def walk_equals(self, formula, args, **kwargs):
+        return self._btor.Eq(*args)
+
+    def walk_function(self, formula, args, **kwargs):
+        #pylint: disable=star-args
+        raise NotImplementedError
+        f = formula.function_name()
```
mikand: this is dead code and should be removed

pysmt/solvers/solver.py
```diff
@@ -58,10 +58,18 @@ def is_sat(self, formula):
...
-        res = self.solve()
-        self.pending_pop = True
+        use_solving_under_assumption = False
+        try:
+            self.push()
+        except NotImplementedError:
+            use_solving_under_assumption = True
+
+        if use_solving_under_assumption:
+            res = self.solve(formula) # TODO: How do we handle model generation?
```
mikand: shouldn't this be ```self.solve(assumptions=[formula])```?

mikand:
Overall this is great! Once fixed the points we can go on and merge the PR.

pysmt/solvers/btor.py
marcogario: We should clarify when back conversion needs to be implemented. I am moving the code that was here to ```get_value```. I am not sure that a proper implementation of ```back``` is needed, at this point.

pysmt/solvers/btor.py
marcogario: This is indeed wrong. Boolector supports functions. I'll extend the logic to QF_UFBV and implement this missing piece.

marcogario:
I addressed most of the issues. I still need to:

- Make sure that all tests pass with the new QF_UF example
- Understand whether we really need the ```back```

marcogario:
Before closing this PR we need to update main README to include BTOR information

marcogario:
The problem with failing tests is MathSAT not supporting UF with boolean arguments. I am not clear on how to proceed. Is an UF with boolean arguments expanded into 2^n symbols?

What about mixed arguments? Is ```(Bool, Bool, Real, Real) -> Real ``` handled by using 4 functions of type ```(Real, Real) -> Real``` plus an ITE selection?

marcogario:
I created #184 to deal with the UF-Bool issue in a second moment.

marcogario:
@mikand , try to run all the tests on your machine. If they pass, please accept this PR so that we can release (independently of shippable/CI results).

mikand:
I successfully run the tests on pysmt-docker with python 2.7 (both all-solvers and btor-only)

PR #172, opened 2015-08-25
Example for All SMT

marcogario:
Added simple example of all-smt. This also shows how to call the logic oracle, to detect the logic of a formula.

### DIFF
--- a/examples/allsmt.py
+++ b/examples/allsmt.py
@@ -1,0 +1,41 @@
+# Perform ALL-SMT on Theory atoms
+#
+# This example shows:
+#  - How to get the logic of a formula
+#  - How to enumerate models
+#  - How to extract a partial model
+#  - How to use the special operator EqualsOrIff
+#
+from pysmt.shortcuts import Solver, Not, And, Symbol, Or
+from pysmt.shortcuts import LE, GE, Int, Plus, Equals, EqualsOrIff
+from pysmt.typing import INT
+from pysmt.oracles import get_logic
+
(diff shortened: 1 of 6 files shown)


examples/allsmt.py
```diff
@@ -0,0 +1,33 @@
...
+from pysmt.typing import INT
+from pysmt.oracles import get_logic
+
+def all_smt(formula, keys):
+    target_logic = get_logic(formula)
+    print("Target Logic: %s" % target_logic)
+    with Solver(logic=target_logic) as solver:
+        solver.add_assertion(formula)
+        while solver.solve():
+            partial_model = [Equals(k, solver.get_value(k)) for k in keys]
```
mikand: In order to be complete, check the type of k and if it is Bool use Iff, this better shows the difference between pysmt and z3py

pysmt/formula.py
```diff
@@ -519,6 +519,24 @@ def Max(self, *args):
...
 
+    def EqualsOrIff(self, left, right):
+        """Returns Equals() or Iff() depending on the type of the arguments.
+
+        This can be used to deal with ambiguous cases where we might be
+        dealing with both Theory and Boolean atoms.
+        """
+        if left.is_symbol():
+            type_ = left.symbol_type()
+        elif right.is_symbol():
```
mikand: You can do the same with constants, but all of this should be unnecessary, as all types are cached

marcogario: Shall I remove that piece of code?

PR #178, opened 2015-09-19
AUTO_LOGIC becomes the new default in methods of Factory.

marcogario:
Previously, all is_*, qelim, interpolation etc. methods where assuming a ```DEFAULT``` logic, if one was not provided.  All those methods can auto-detect the logic, since the problem is provided in input. Therefore, the new default behavior becomes using ```AUTO_LOGIC```.


marcogario:
This change was triggered by the fact that ```QF_UFLIRA``` is not the most generic logic anymore, since we have Boolector.

Issue #187, opened 2015-10-09
Implement Scalar type


marcogario:

marcogario:
Branch issue_187_enum_support, starts dealing with this issue. It currently only introduces the Enum Type at the Formula level. No translation to solver is performed (yet).

https://github.com/pysmt/pysmt/tree/issue_187_enum_support

yoni206:
The link https://github.com/pysmt/pysmt/tree/issue_187_enum_support is invalid. Does the branch still exist? and if so, could you please provide a current link to it? Thanks

marcogario:
That was a quite ols activity. I can try to find if I have a copy on my laptop.
In between we changed how we handle types, so that would need to be adapted. 

One reason why we did not move forward with this was that we were missing a good motivating use case.

If you can share some more info on what do you need to do, that would be good.

yoni206:
Well, ultimate goal is to support the theory of datatypes (smtlib 2.6). An intermediate goal would be to support datatypes with null-ary constructors (enums).

marcogario:
Ok, if this is the case, the I would suggest that we look into this in terms of supporting smtlib 2.6, meaning that we follow a design that is consistent with the standard, and we require to extend also the smtlib printers/parsers. 

2 questions:
- Are you using also the new "match" term (also introduced in 2.6)?
- do you know which solvers support the datatype as defined in the standard?

marcogario:
The old branch is available as https://github.com/pysmt/pysmt/tree/i187/enum_support . As mentioned, I would probably look into adding the support for datatypes directly.

yoni206:
Thank's for the link! I am not using `match` currently. AFAIK, cvc4 and z3 support it (at least, i can run some benchmarks on both).

odedp:
Any plan to support enum/datatypes soon? I am considering using pysmt for my project but I need to be able to define a finite type and quantify over it. Thanks!!

marcogario:
Not in the short term, but we would be happy to shepherd a patch in this direction. 

When you say enum/datatypes do you mean the SMTLIB dataypes or something else? 
If you want to open an issue describing more what you are trying to do, maybe we can support you with an alternative.

odedp:
Yes, I mean the SMTLIB2 datatype.

[...]

PR #189, opened 2015-10-10
Mathsat bool UF support

marcogario:


marcogario:
MathSAT SMT-Lib wrapper hangs because of this. Need to look into this.

marcogario:
The problem is due to the GenericWrapper making too many assumptions on the output of commands... In particular, MathSAT raises an error after the "define-fun" command. However, this is not checked, because we did not expect an output from that command. 

My proposal is to see if there is a way to tell MathSAT to exit on error.
The better solution would be to see if the ```(error ...)``` answer is standard in SMT-LIB and adapt the wrapper to look for it after issuing each command.

marcogario:
The (error ..) is standard from SMT-LIB.
My current propsal is to enable response from the solver ```(set-option :print-success true)``` and *check* that we get success or error, as defined in the standard.

marcogario:
Summary:
- 1c6ac77 : MathSAT converter performs some simple rewriting for UF with bool args. It is not super fancy, but it works for now. We can reconsider it when we will introduce the encoders.
-  a108759 : Makes the interaction with the wrappers more robust
- c6d60a7 : Activates Z3 as a wrapper for testing. This is probably a good idea, to avoid making changes to the smt-lib wrapper that work only on MathSAT. For now I activated it on all python versions. Maybe we can reduce this to 3.4 only.

pysmt/solvers/msat.py
```diff
@@ -1150,3 +1157,69 @@ def sequence_interpolant(self, formulas):
                 mathsat.msat_destroy_config(cfg)
             if env:
                 mathsat.msat_destroy_env(env)
+
+
+class MSatBoolUFRewriter(IdentityDagWalker):
```
mikand: Why is this class here? There is nothing that is mathsat-specific hence I think it could go in the rewritings.py module to be used in other situations as well (e.g. in the SMT-LIB wrapper)

mikand:
I think this is great. Maybe we can move the rewriter in anothe module and use it in the SMT-Lib wrapper (using a parameter or an option to enable or disable it)...

marcogario:
Ok to move it to rewriting. In general, the proper handling of encoders is left for a future version (I was targeting ```0.5``` -- See #187, #186). In fact, this is not back-converting the models, that is a necessary step for many encoders.

Currently, only MathSAT needs this. Cluttering the SMT-Lib wrapper with this option seems unnecessary, and I'd prefer to post-pone this to when we have a principled way of dealing with encoders, and register encoders into converters.

mikand:
OK to accept. You decide if you wanna move the rewriter or not.

marcogario:
Merging now. Proper handling deferred to issue #203

PR #190, opened 2015-10-12
Switched CI from shippable to travis

mikand:
Removed Shippable configuration files and switched CI to Travis.

Travis is used through its new container-based infrastructure and we exploit directory caching to avoid solver re-compilation


marcogario:
If I understand correctly, we are not using the pysmt-docker image, right?

.travis.yml
```diff
@@ -0,0 +1,92 @@
...
+  - ${TRAVIS_BUILD_DIR}/.smt_solvers/z3
+  - ${TRAVIS_BUILD_DIR}/.smt_solvers/CVC4_bin
+  - ${TRAVIS_BUILD_DIR}/.smt_solvers/picosat-960
+  - ${TRAVIS_BUILD_DIR}/.smt_solvers/repycudd-4861f4df8abc2ca205a6a09b30fdc8cfd29f6ebb
+  - ${TRAVIS_BUILD_DIR}/.smt_solvers/boolector-2.1.1-with-lingeling-b85
+  - ${TRAVIS_BUILD_DIR}/.smt_solvers/pyices-aa0b91c39aa00c19c2160e83aad822dc468ce328
+  - ${TRAVIS_BUILD_DIR}/.smt_solvers/yices-2.3.0
+  - ${TRAVIS_BUILD_DIR}/.smt_solvers/yices_bin
+  - ${TRAVIS_BUILD_DIR}/.smt_solvers/mathsat-5.3.8-linux-x86_64
+  - ${HOME}/.local/lib/python2.7/site-packages
```
marcogario: This is for six? Shouldn't this also be cached for python 3.4?

.travis.yml
```diff
@@ -0,0 +1,92 @@
...
+    - autoconf
+    - libtool
+    - wget
+    - curl
+    - libboost-dev
+    - python3-dev
+
+python:
+  - 2.7
+  - 3.4
```
marcogario: Is pypy not supported?

.travis.yml
```diff
@@ -0,0 +1,92 @@
...
+
+before_install:
+  - pip install six
+  - source ./travis.sh
+
+
+script:
+  - env
+  - python install.py --check
+  - nosetests pysmt -v -A "not slow"
```
marcogario: I think we should run all tests, and keep "not slow" only for local testing.

.travis.yml
mikand: This is for pyices that is installed with setup.py --user

.travis.yml
mikand: Good point, I can try

.travis.yml
mikand: OK

marcogario:
:+1: 

- Check pypy
- Remove "not slow"
- Change build badge in README.rst

PR #191, opened 2015-10-14
Examples: Multi-processing with pySMT

marcogario:
Simple example for using ```multiprocessing.Pool.map``` and ```multiprocessing.Pool.apply_async``` in pySMT.

### DIFF
--- a/examples/README.rst
+++ b/examples/README.rst
@@ -4,14 +4,14 @@ Examples
 This directory contains examples to get you started using
 pySMT. Suggested order:
 
-1. basic.py : First example of the README
-2. puzzle.py : Hello World word puzzle
-3. infix_notation.py : Hello World word puzzle using infix-notation
-4. combine_solvers.py : Combine multiple solvers to solve the same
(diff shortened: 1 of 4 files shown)


examples/README.rst
```diff
@@ -14,4 +14,5 @@ pySMT. Suggested order:
 8. efsmt.py: Shows how to combine two different solvers to solve an Exists Forall problem.
 9. allsmt.py: How to detect the logic of a formula and perform model enumeration.
 10. sudoku/sudoku.py: Solves sudoku problems on the command line using a simple encoding using either QF_LIA or QF_BV
-11. sudoku/gui.py: A small GUI that embeds sudoku.py and allows a user-friendly intercation with the sudoku solver.
+    sudoku/gui.py: A small GUI that embeds sudoku.py and allows a user-friendly intercation with the sudoku solver.
```
mikand: Why you removed the bullet for sudoku/gui?

marcogario: I collapsed that into a single point, and added a README for the sudoku directory. The rationale is that the two examples do not show two aspects of pySMT, and thus should be considered as a single (although fairly complex) example.

PR #194, opened 2015-10-16
Human Readable expression parsing

mikand:
This PR introduces a new parser for expressions as printed by the HRSerializer (FNode.serialize()) method.

The parser is a simple but efficient recursive top-down Pratt-Parser that should work well with simple not-too-deep formulae.


pysmt/parsing.py
```diff
@@ -0,0 +1,435 @@
...
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+
+import re
+from fractions import Fraction
+
+from pysmt.shortcuts import get_env
```
marcogario: This should not call shortcuts, but use the functions from environment instead. This was done to avoid circular dependencies, and make ```shortcuts``` the top-level module for everything "outside" pySMT.

pysmt/parsing.py
```diff
@@ -0,0 +1,435 @@
...
+    """This class produces a stream of token objects for the Pratt-parser"""
+
+    def __init__(self, env=None):
+        if env is None:
+            env = get_env()
+        self.env = env
+        self.mgr = env.formula_manager
+        self.get_type = env.stc.get_type
+
+        self.rules = []
```
marcogario: Where are "rules" defined?
You are appending (a lot) of tuples, but the semantics of the tuple is unclear.

marcogario: I have the feeling this should go somewhere else. Especially, if we intend to make this extensible.

pysmt/parsing.py
```diff
@@ -0,0 +1,435 @@
...
+            env = get_env()
+        self.env = env
+        self.mgr = env.formula_manager
+        self.get_type = env.stc.get_type
+        self.lexer = HRLexer(env)
+
+        self.token = None
+        self.tokenizer = None
+
+    def expression(self, rbp=0):
```
marcogario: ```_expression``` ?

pysmt/parsing.py
```diff
@@ -0,0 +1,435 @@
...
+        self.rules.append((r"(s%)", InfixOpAdapter(self.mgr.BVSRem, 80), False)) # div
+        self.rules.append((r"(u%)", InfixOpAdapter(self.mgr.BVURem, 80), False)) # div
+        self.rules.append((r"(\?)", ExprIf(), False)) # question
+        self.rules.append((r"(::)", InfixOpAdapter(self.mgr.BVConcat, 90), False)) # BVXor
+        self.rules.append((r"(:)", ExprElse(), False)) # colon
+        self.rules.append((r"(False)", Constant(self.mgr.FALSE()) , False)) # False
+        self.rules.append((r"(True)", Constant(self.mgr.TRUE()), False)) # True
+        self.rules.append((r"(,)", ExprComma(), False)) # comma
+        self.rules.append((r"(\.)", ExprDot(), False)) # dot
+        self.rules.append((r"(xor)", InfixOpAdapter(self.mgr.BVXor, 10), False)) # BVXor
```
marcogario: Comments from here on are wrong.

pysmt/parsing.py
```diff
@@ -0,0 +1,435 @@
...
+        GrammarSymbol.__init__(self)
+        self.operator = operator
+        self.lbp = lbp
+
+    def nud(self, parser):
+        right = parser.expression(self.lbp)
+        return self.operator(right)
+
+
+class Quantifier(GrammarSymbol):
```
marcogario: This is a specific Grammar Symbol, while the one before and after are "Factories".
Would it be possible to separate the "infrastructure" from the actual parsing elements?

pysmt/utils.py
```diff
@@ -45,3 +46,22 @@ def twos_complement(val, bits):
     if (val & (1 << (bits - 1))) != 0: # if sign bit is set
         val = val - (2 ** bits)        # compute negative value
     return val
+
+
+_simple_symbol_prog = re.compile(r"^[~!@\$%\^&\*_\-+=<>\.\?\/A-Za-z][~!@\$%\^&\*_\-+=<>\.\?\/A-Za-z0-9]*$")
+def quote(name, style='|'):
```
marcogario: Not a huge fan of the ```utils``` module. These are functions related to serialization/string conversion of symbols and expressions. I think this should go somewhere in the printing modules or directly in "FNode".

pysmt/test/test_hr_parsing.py
```diff
@@ -0,0 +1,90 @@
...
+
+class TestHRParser(TestCase):
+
+    def test_precedences(self):
+        p = HRParser()
+
+        a,b,c = (Symbol(v) for v in "abc")
+        x,y = (Symbol(v, REAL) for v in "xy")
+
+        tests = []
```
marcogario: This is not better than writing multiple assertions. When/If a test fails, you will not get the exact line that makes the test fail, but you will always get line 49. Printing additional info in the assertEqual mitigates the issue, but I do not see the need for this list in the first place.

pysmt/test/test_hr_parsing.py
```diff
@@ -0,0 +1,90 @@
...
+
+        finally:
+            if fname is not None:
+                os.unlink(fname)
+
+
+    def test_examples(self):
+        p = HRParser()
+
+        for (f, _, _, _) in get_example_formulae():
```
marcogario: In case you find it more readable, you can also do:
```
for example in get_example_formulae():
    f = example.expr
   ...
```

pysmt/test/test_hr_parsing.py
```diff
@@ -0,0 +1,90 @@
...
+            s = f.serialize()
+            res = p.parse(s)
+
+            check = (res == f)
+            if not check:
+                try:
+                    check = is_valid(Iff(res, f))
+                except NoSolverAvailableError:
```
marcogario: This means that we can only trust this test if there is a solver installed. Lets add a comment about it at the beginning of the test or (better yet) decorate the test by requiring a solver for BV and UFLIRA.
When testing the "all" configuration, we should be able to see that no test was skipped, and thus know that this test *really* passed.

marcogario:
The structure of the code is a bit confusing, especially when not knowing the details of the parser strategy.
In particular, it feels like the infrastructure machinery could be separated from the actual pySMT parser. If this is the case, maybe this will simplify the extensibility of the parser.

I can try to take a shot at it, if you want.

pysmt/parsing.py
marcogario: ok

pysmt/test/test_hr_parsing.py
marcogario: This does not apply anymore since we tests for each string.

marcogario:
I divided the pratt parser from the HR specific code. 
I am not sure about the named tuple ```Rule```. Are the names that I chose for the field descriptive? In particular, I called the middle value ```symbol```. I am not sure that is the right term.

I think this could be further improved, by generating the rules by "type". However, for now, I think this works, and extensible.

pysmt/parsing/pratt.py
```diff
@@ -0,0 +1,221 @@
...
+
+    def advance(self):
+        """Advance reading of one token"""
+        self.token = next(self.tokenizer)
+
+# EOC PrattParser
+
+
+# Rules for the Lexer
+Rule = namedtuple('Rule', ['regex', 'symbol', 'is_functional'])
```
mikand: I would have used "token" but "symbol" also works

pysmt/test/test_hr_parsing.py
mikand: Your second comment is not clear to me. Shall we change the test or not?

marcogario: The test is ok.

Issue #183, opened 2015-10-05
Shannon expander


mikand:
We may implement a quantifier eliminator for QF_BOOL (or better, for purely Boolean quantifiers) using the classical Shannon expansion.

mikand:
Closed in #195

PR #195, opened 2015-10-16
Implemented Shannon-Expansion

mikand:
Implemented the classical shannon expansion method as a quantifier eliminator for the BOOL logic. This addresses #183.

### DIFF
--- a/pysmt/solvers/qelim.py
+++ b/pysmt/solvers/qelim.py
@@ -18,8 +18,8 @@
 import pysmt.logics
 
 from pysmt.walkers.identitydag import IdentityDagWalker
-from pysmt.utils import powerset
-from pysmt.oracles import get_logic
+from pysmt.utils import all_assignments
+from pysmt.exceptions import InternalSolverError
 
 
 class QuantifierEliminator(object):
@@ -51,6 +51,7 @@ class QuantifierEliminator(object):
 
 
 class ShannonQuantifierEliminator(QuantifierEliminator, IdentityDagWalker):
+    """Quantifier Elimination using Shannon Expansion."""
 
     LOGICS = [pysmt.logics.BOOL]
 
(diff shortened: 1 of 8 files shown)


pysmt/solvers/qelim.py
```diff
@@ -42,3 +48,39 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+    LOGICS = [pysmt.logics.BOOL]
+
+    def __init__(self, environment, logic=None):
+        IdentityDagWalker.__init__(self, env=environment)
+        self.logic = logic
+
+    def eliminate_quantifiers(self, formula):
+        logic = get_logic(formula, self.env)
+        if not logic <= pysmt.logics.BOOL:
+            raise NotImplementedError("Shannon quantifier elimination only "\
```
marcogario: Isn't this too strong?
Shouldn't we check that the quantified variables are Boolean? The expression:
``` Exists x. x & (r < 5)``` becomes ```(False & (r < 5)) | (True & (r < 5))```, no?
Am I missing something?

File: pysmt/utils.py [hunk 2]
```diff
@@ -1,7 +1,12 @@
 import io
+import itertools
 from six import PY2
 from fractions import Fraction
 
+# powerset([1,2,3]) --> () (1,) (2,) (3,) (1,2) (1,3) (2,3) (1,2,3)
```
marcogario: Make this into a proper function.

pysmt/solvers/qelim.py
```diff
@@ -42,3 +48,39 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+                                      "(detected logic is: %s)" % str(logic))
+
+        return self.walk(formula)
+
+    def build_assignments(self, variables):
+        for s in powerset(variables):
+            yield dict((x, self.mgr.Bool(x in s)) for x in variables)
+
+    def walk_forall(self, formula, args, **kwargs):
+        conj = []
```
marcogario: Unify the two functions. Return the list of elements:
```
def walk_forall(...):
  return self.mgr.And(self._expand(formula, args))
```

pysmt/test/test_shannon_expansion.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def test_nested(self):
+        f = Exists([self.x], ForAll([self.y], Or(self.x, self.y)))
+        g = qelim(f, solver_name="shannon")
+        g = g.simplify()
+        self.assertEqual(g, TRUE())
+
+
+    def test_examples_solving(self):
+        for example in get_example_formulae():
```
marcogario: We really need to reconsider the interface of ```get_example_formulae()``` I think selection by logic is one of the most common things we do in the tests...

pysmt/test/test_shannon_expansion.py
```diff
@@ -0,0 +1,84 @@
...
+            else:
+                self.assertTrue(g.is_false())
+
+            f = ForAll(fv, example.expr)
+            g = qelim(f, solver_name="shannon").simplify()
+            if example.is_valid:
+                self.assertTrue(g.is_true())
+            else:
+                self.assertTrue(g.is_false())
+
```
marcogario: If we extend it for SMT formulae, then we should probably compare the result against a solver (i.e., Z3).

marcogario:
Good!

Open points:
- Can we extend to Theory?
- Squash everything
- Update docs/Changes

pysmt/test/test_shannon_expansion.py
mikand: We cannot solve the satisfiability problem if a single non-boolean variable is present in the formula...

pysmt/solvers/qelim.py
mikand: OK

pysmt/utils.py
mikand: OK

pysmt/solvers/qelim.py
mikand: In principle yes, the problem is that this QE is complete for BOOL, but incomplete for any other theory (we have no way of distinguishing a formula having only boolean variables quantified).

The pysmt framework for QE does not fit well for this case: shall we declare this QE as supporting any logic and then throw an execption once a non-boolean variable is quantified?

marcogario: I see. What do you think of this:

- This QE declares support only for BOOL
- The check within the QE is done only on quantified variables

This solves the issue of automatic selection (it never gets selected incorrectly), but one can force the use of this qe ```qelim(f, name="shannon")```, that will raise an error if quantified variables are not bool.

mikand: OK, seems reasonable

pysmt/test/test_shannon_expansion.py
marcogario: yes, I meant another QE. We should be able to check that the two quantifier-free expressions are equivalent:
```
f = Exists([x,y], lira_formula)
self.assertIsValid(Iff(qelim(f, name="shannon"), qelim(f)))
```

marcogario: A special test was added for the mixed case.

PR #196, opened 2015-10-18
Better handling of Solver API discovery.

marcogario:
Using the ```ImportError``` exception was masking other problems in the solver wrapper. These made debugging harder, since errors were masked by simply saying that the solver was not available.

To address this, the exception ```SolverAPINotFound``` is introduced.


mikand:
Nice!

PR #207, opened 2015-11-04
Issue 152: Most General substitution

marcogario:
Made Substituter a super-class for:
 - MGSubstituter (Most General)
 - MSSubstituter (Most Specific)

This makes it possible to restore the previous behavior, by changing which one is instantiated by the environment.

**Note**: we only allow one global substituter, in order to ensure a consistent behavior. One could argue that an option to the ```substitute()``` function would be cleaner/nicer.

**This is a transitional commit** The default is still the old one (MSS), but a warning has been added in the substitute method.

### DIFF
--- a/pysmt/substituter.py
+++ b/pysmt/substituter.py
@@ -15,25 +15,93 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 #
+import warnings
+
 from six import iteritems
 
 import pysmt.walkers
 import pysmt.operators as op
 
-class Substituter(pysmt.walkers.DagWalker):
 
+class Substituter(pysmt.walkers.DagWalker):
+    """Performs substitution of a set of terms within a formula.
+
+    Let f be a formula ans subs be a map from formula to formula.
+    Substitution returns a formula f' in which all occurrences of the
+    keys of the map have been replaced by their value.
+
+    There are a few considerations to take into account:
+      - In which order to apply the substitution
+      - How to deal with quantified subformulas
+
+    The order in which we apply the substitutions gives raise to two
+    different approaches: Most General Substitution and Most Specific
+    Substitution. Lets consider the example:
+
+      f = (a & b)
+      subs = {a -> c, (c & b) -> d, (a & b) -> c}
+
+    With the Most General Substitution (MGS) we obtain:
+      f' = c
+    with the Most Specific Substitution (MSS) we obtain:
+      f' = d
+
+    The default behavior before version 0.5 was MSS. However, this
+    leads to unexpected results when dealing with literals, i.e.,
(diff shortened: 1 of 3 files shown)


pysmt/substituter.py
```diff
@@ -15,55 +15,54 @@
...
+      f' = d
 
-        return self.walk(formula, substitutions=subs)
+    The default behavior before version 0.5 was MSS. However, this
+    leads to unexpected results when dealing with literals, i.e.,
+    substitutions in which both x and Not(x) appear, do not work as
+    expected.  In case of doubt, it is recommended to issue two
+    separate calls to the substitution procedure.
 
+    Quantifiers: ???
```
mikand: Why '???'. The behavior with quantifiers should be clear

pysmt/substituter.py
```diff
@@ -99,6 +98,80 @@ def _push_with_children_to_stack(self, formula, **kwargs):
...
+                    "Value %d does not belong to the Formula Manager." % i)
+
+        self.orig_subs = subs
+        return self.walk(formula, substitutions=subs)
+
+
+class MGSubstituter(Substituter):
+    """Performs Most Specific Substitution.
+
+    This is the default behavior sinve version 0.5
```
mikand: Typo

pysmt/substituter.py
```diff
@@ -99,6 +98,80 @@ def _push_with_children_to_stack(self, formula, **kwargs):
...
+
+
+    def walk_identity_or_replace(self, formula, args, **kwargs):
+        """If the formula appears in the substitution, return the substitution.
+           Otherwise, rebuild the formula, with its new children.
+        """
+        if formula in self.orig_subs:
+            return self.orig_subs[formula]
+        else:
+            return self._inner_idw.functions[formula.node_type()](formula, args=args)
```
mikand: It took me a while to realize how this works, maybe a line of comment on the use of args....

Moreover, you should also pass '''**kwargs''' to the inner function.

pysmt/substituter.py
```diff
@@ -99,6 +98,80 @@ def _push_with_children_to_stack(self, formula, **kwargs):
...
+
+# EOC MGSubstituter
+
+
+class MSSubstituter(Substituter):
```
mikand: It occurs to me that maybe we can shorten the implementation of this class analogously to what you did on the MGSubstiturer, but we apply the substitute method on the ```args```?

pysmt/substituter.py
marcogario: Yes, that was meant to remind me of the issue we discussed... I should have removed it :smile:

pysmt/substituter.py
marcogario: Done.

PR #208, opened 2015-11-06
MathSAT solver: new implementation of MathSAT5Solver.get_model()

agriggio:
exploiting the new API function msat_get_model()

The next version of MathSAT will include a new API function
`msat_get_model()`, returning a `msat_model` object that can be used
even after the assertion stack of the `msat_env` is modified.

This changeset modifies the MathSAT5 wrapper of PySMT to make use of the
new feature.

I'm opening the PR now for feedback, although the merge should probably
be postponed until the new version of MathSAT is out.

### DIFF
--- a/pysmt/solvers/msat.py
+++ b/pysmt/solvers/msat.py
@@ -34,9 +34,8 @@ from pysmt.oracles import get_logic
 import pysmt.operators as op
 from pysmt import typing as types
 from pysmt.solvers.solver import (IncrementalTrackingSolver, UnsatCoreSolver,
-                                  Converter)
+                                  Model, Converter)
 from pysmt.solvers.smtlib import SmtLibBasicSolver, SmtLibIgnoreMixin
-from pysmt.solvers.eager import EagerModel
 from pysmt.walkers import DagWalker
 from pysmt.exceptions import (SolverReturnedUnknownResultError,
                               SolverNotConfiguredForUnsatCoresError,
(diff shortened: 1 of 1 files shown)


pysmt/solvers/msat.py
```diff
@@ -48,6 +47,88 @@
...
+
+    def unref(self):
+        assert self.refs > 0
+        self.refs -= 1
+        if not self.refs:
+            mathsat.msat_destroy_env(self.msat_env)
+
+    def __call__(self):
+        return self.msat_env
+
```
marcogario: I think this can be simplified by adding a ```__del__``` method in _MSatEnvWrapper, and let python deal with the ref-counting. If every object that uses the msat_env keeps an internal copy (as it seems to be the case), everything should work.

pysmt/solvers/msat.py
```diff
@@ -48,6 +47,88 @@
 from pysmt.walkers.identitydag import IdentityDagWalker
 
 
+class _MSatEnvWrapper(object):
```
marcogario: Minor: I think this can just be called ```MSatEnvWrapper``` without the leading underscore.
Due to the ref-counting handling, this class becomes the preferred method to handle the mathsat environment at the python level.

pysmt/solvers/msat.py
```diff
@@ -272,49 +354,19 @@ def get_value(self, item):
...
-        return EagerModel(assignment=assignment, environment=self.environment)
-
+        return MathSAT5Model(self.environment, self.msat_env_wrapper.ref())
 
     def exit(self):
         if not self._destroyed:
             self._destroyed = True
-            mathsat.msat_destroy_env(self.msat_env)
+            self.msat_env_wrapper.unref()
             mathsat.msat_destroy_config(self.msat_config)
```
marcogario: Do we need to guard also msat_config? From ```mathsat.h``` (msat_destroy_config):
```
 * It is an error to destroy a configuration that is still being used by an
 * environment.
```
 If so, I think we can safely assume that 1 env = 1 conf, and move the freeing of the msat_config within the ```MSatEnvWrapper```

pysmt/solvers/msat.py
agriggio: Ok, but if you want to go this route, I think the patch needs to become
more pervasive, and change every use of a "bare" msat_env... 
I used the explicit mechanism exactly to highlight that this is a
special case.

pysmt/solvers/msat.py
agriggio: Ok

pysmt/solvers/msat.py
agriggio: No, the config can be safely destroyed, the `msat_env` keeps its internal copy (the documentation is wrong here)

marcogario:
Cool! Thanks for submitting this! Having the solver deal with the model, considerably simplifies the work of pySMT  :smiley: 

Tests are failing because we are using an older version of MathSAT (obviously).

marcogario:
See PR #214 .

Issue #198, opened 2015-10-20
install.py: Refactoring


marcogario:
install.py needs to be refactored, because it is not manageable, if we want to extend to more platforms.

In particular, we should have a main class: ```SolverInstaller``` that deals with the main flow of installation:

- Download
- Unpack
- (Optionally) Compile
- Create PYTHONPATH

Each Solver Installer would be a subclass of the above, in order to provide specific info like:

- Where to download from (taking into account platform etc)
- How to compile
- How to create the PYTHONPATH

Additionally, it would be nice if the download function was a bit smarter, and it attempted first to use the mirror and then connect to the original URL in case of failure.

marcogario:
Additionally, install.py should provide a way to obtain the necessary pythonpath information, so that we can do:
```
$ ./install.py --env --z3
export PYTHONPATH=$PYTHONPATH:/..../

$ `./install.py --env --z3`
```

marcogario:
The current version of install.py does not handle errors in the commands properly. E.g., it is quite common that the compilation failed but install.py thinks everything is ok. The new version should handle this properly.

mikand:
PR #209

PR #209, opened 2015-11-07
New solver installer architecture

mikand:
Refactored solver installing architecture following #198.

Each solver has now a dedicated class for installation and all the bindings get copied in a specific folder.

The command line interface of install.py has been cleaned and each solver is now versioned.


pysmt/cmd/install.py
```diff
@@ -490,35 +91,40 @@ def check_install():
...
+    for i in INSTALLERS:
+        name = i.InstallerClass.SOLVER
+        parser.add_argument('--%s' % name, dest=name, action='store_true',
+                            default=False, help='Install %s' % name)
+
+    parser.add_argument('--all', dest='all_solvers', action='store_true',
+                        default=False,
+                        help='Install all the solvers')
+
+    parser.add_argument('--force-redo', dest='force_redo', action='store_true',
```
marcogario: I think this can be simplified into ```--force```

pysmt/cmd/install.py
```diff
@@ -547,46 +153,54 @@ def print_welcome():
...
+            installer = i.InstallerClass(install_dir=install_dir,
+                                         bindings_dir=bindings_dir,
+                                         solver_version=i.version,
+                                         mirror_link=mirror_url,
+                                         **i.extra_params)
+
+            if not installer.is_installed():
+                if not options.skip_intro:
+                    print_welcome()
+                    options.skip_intro = True
```
marcogario: Can't this be moved outside of the loop, to do it only once?

pysmt/cmd/install.py
```diff
@@ -547,46 +153,54 @@ def print_welcome():
 def main():
     options = parse_options()
```
marcogario: If no arguments are given, I think we should print the help message. Currently, the script exits without say anything.

pysmt/cmd/installers.py
```diff
@@ -0,0 +1,630 @@
...
+        os.rename(source_file, dest_file)
+
+
+class MSatInstaller(SolverInstaller):
+
+    SOLVER = "msat"
+
+    def __init__(self, install_dir, bindings_dir, solver_version, os_name=None,
+                 architecture=None, python_version=None, mirror_link=None):
+        SolverInstaller.__init__(self, install_dir, bindings_dir, solver_version,
```
marcogario: Use keyword arguments. There are too many parameters, and we will eventually mix-up some of them.

pysmt/cmd/installers.py
```diff
@@ -0,0 +1,630 @@
...
+                                 os_name, architecture, python_version, mirror_link)
+
+    def download(self):
+        pass # Nothing to download
+
+    def unpack(self):
+        pass # Nothing to unpack
+
+    def compile(self):
+        pip.main(['install', '--pre', '--target', self.bindings_dir, 'pypicosat'])
```
marcogario: ```--pre``` is not needed anymore. We have a stable version out.

pysmt/cmd/installers.py
```diff
@@ -0,0 +1,630 @@
...
+        self.base_dir = os.path.join(self.install_dir, self.SOLVER)
+        if not os.path.exists(self.base_dir):
+            os.mkdir(self.base_dir)
+
+
+    def download(self):
+        """Downloads the archive from one of the mirrors"""
+        if not os.path.exists(self.archive_path):
+            for _ in xrange(self.trials_404):
+                for link in self.download_links:
```
marcogario: For debugging purposes, I think it would be useful to know from which mirror the package was downloaded. Maybe the index is enough...

pysmt/cmd/installers.py
```diff
@@ -0,0 +1,630 @@
...
+                 architecture=None, python_version=None, mirror_link=None):
+        SolverInstaller.__init__(self, install_dir, bindings_dir, solver_version,
+                                 os_name, architecture, python_version, mirror_link)
+        self.archive_name = "mathsat-%s-%s-%s.tar.gz" % (self.solver_version,
+                                                  self.os_name, self.architecture)
+        self.archive_path = os.path.join(self.base_dir, self.archive_name)
+
+        if self.mirror_link is not None:
+            self.download_links.append("%s/%s" % (self.mirror_link, self.archive_name))
+        self.download_links.append( "http://mathsat.fbk.eu/download.php?file=%s"\
```
marcogario: This is repeated in all solvers, and it seems to be something that should go into the parent class.
I see 2 ways of changing this:
- add a method ```_create_download_links``` that covers the lines 245-247. This is not particularly, robust, because it assumes that ```archive_name``` has already been set... but it gets the job done.
- convert ```download_links``` into a generator, that creates the links on the fly (this would be a nicer solution if integrated with the trials_404: ```download_links(trials_404=3)```

pysmt/cmd/installers.py
```diff
@@ -0,0 +1,630 @@
...
+
+    SOLVER = "msat"
+
+    def __init__(self, install_dir, bindings_dir, solver_version, os_name=None,
+                 architecture=None, python_version=None, mirror_link=None):
+        SolverInstaller.__init__(self, install_dir, bindings_dir, solver_version,
+                                 os_name, architecture, python_version, mirror_link)
+        self.archive_name = "mathsat-%s-%s-%s.tar.gz" % (self.solver_version,
+                                                  self.os_name, self.architecture)
+        self.archive_path = os.path.join(self.base_dir, self.archive_name)
```
marcogario: Convert this into a property of ```SolverInstaller```

pysmt/cmd/installers.py
```diff
@@ -0,0 +1,630 @@
...
+class YicesInstaller(SolverInstaller):
+
+    SOLVER = "yices"
+
+    def __init__(self, install_dir, bindings_dir, solver_version, os_name=None,
+                 architecture=None, python_version=None, mirror_link=None):
+        SolverInstaller.__init__(self, install_dir, bindings_dir, solver_version,
+                                 os_name, architecture, python_version, mirror_link)
+
+        assert self.os_name == "linux"
```
marcogario: This check is only done for yices. Is it because of ```pyices```?

In general, I don't know whether we should perform this type of checks.

marcogario:
:ok_hand: 

Maybe we want to break ```installers.py``` into a module that mimicks the structure of ```pysmt.solvers```? We could import all Installers into ```___init.py__```, in order to preserve the current import structure (i.e., ```pysmt.cmd.installers.MSatInstaller```).

pysmt/cmd/install.py
mikand: It is done only once, (```skip_intro``` is set to true the first loop), but it is there to be executed only if a solver gets instaled (if it is found no message is printed).

pysmt/cmd/install.py
mikand: no arguments means install all the solvers in ```PYSMT_SOLVER``` env variable...

We could print a messgae like "Nothing to do. Try install.py --help" if no argument and ```PYSMT_SOLVER``` is empty.

pysmt/cmd/installers.py
mikand: OK

pysmt/cmd/installers.py
mikand: OK

pysmt/cmd/installers.py
mikand: OK

mikand:
I'm done with the modifications. I greatly reduced the amount of code following your suggestions.

pysmt/cmd/installers/base.py
```diff
@@ -0,0 +1,205 @@
+import os
+import platform
+import sys
+import shutil
+
+from six.moves import xrange
+from six.moves.urllib import request as urllib2
+from six.moves.urllib.error import HTTPError
+
+from pysmt.cmd.installers.utils import unzip, untar
```
marcogario: I think it would be better to move those functions within this class. Base should provide all the functionalities that you need.

marcogario:
:+1: after removing ```.utils```

PR #212, opened 2015-11-08
Examples: SMT-LIB Parser

marcogario:
This example covers the use of the SMT-LIB parsing functions in pySMT.
In particular, we discuss:

 - Read a file in SMT-LIB format
 - Write a file in SMT-LIB format
 - Manipulation of the SMT-LIB script object
 - Annotations
 - Parser extension with custom commands

This PR fixes also a bug in the parsing of parameters within define-fun.

### DIFF
--- a/examples/smtlib.py
+++ b/examples/smtlib.py
@@ -1,0 +1,223 @@
+# The last part of the example requires a QF_LIA solver to be installed.
+#
+#
+# This example shows how to interact with files in the SMT-LIB
+# format. In particular:
+#
+# 1. How to read a file in SMT-LIB format
+# 2. How to write a file in SMT-LIB format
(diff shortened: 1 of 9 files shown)


examples/smtlib.py
```diff
@@ -0,0 +1,229 @@
...
+(check-sat)
+(assert .def_1)
+(check-sat)
+(pop)
+(check-sat)
+"""
+
+def mock_smtlib_file():
+    """Utility functions to pretend that the string is a buffer."""
+    # The demo file is at the end of the example
```
mikand: end of the example?

marcogario: Oops. I simplified this part of the example.

PR #214, opened 2015-11-10
MathSAT Solver: exploiting the new API function msat_get_model()

marcogario:
This extends the patch by @agriggio (PR #208) that introduces the new MathSAT5Model object.

We introduce the class ```MSatEnv``` that is used to manage the msat_environment object. In this way, it is possible to decouple the Solver from the Environment, and let the ```MathSAT5Model``` keep an internal reference to the ```MSatEnv``` that survives after the Solver has been closed.

This changes the way that mathsat api functions are called!

### DIFF
--- a/examples/allsat.py
+++ b/examples/allsat.py
@@ -31,7 +31,7 @@ msat.add_assertion(f) # This is still at pySMT level
 result = []
 # Directly invoke the mathsat API !!!
 # The second term is a list of "important variables"
-mathsat.msat_all_sat(msat.msat_env,
+mathsat.msat_all_sat(msat.msat_env(),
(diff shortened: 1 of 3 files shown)


marcogario:
This depends on a development version of MathSAT -- Cannot be merged for now.

mikand:
I updated installer to use the new version of mathsat. Shall we save backward compatibility wrt old MathSAT versions?

agriggio:
I see no reason to keep backward compatibility

mikand:
For me, this is OK, @marcogario merge if you see no issues.

marcogario:
:+1: but I need to update https://github.com/pysmt/pysmt/blob/master/examples/allsat.py before merging.

PR #215, opened 2015-11-12
SMT-LIB: Improved parsing compliance wrt SMT-LIB 2 and 2.5

marcogario:
- Default parser now deals with SMT-LIB 2.5
- SmtLib20Parser can be used for strict SMT-LIB 2.0 compliance
- Improved handling of not implemented commands in parser
- Test and documentation

Some commands are still not supported by the parser.

### DIFF
--- a/pysmt/smtlib/commands.py
+++ b/pysmt/smtlib/commands.py
@@ -17,33 +17,69 @@
 #
 """Defines constants for the commands of the SMT-LIB"""
 
-SET_INFO='set-info'
-RESET_ASSERTIONS='reset-assertions'
-GET_VALUE='get-value'
-SET_OPTION='set-option'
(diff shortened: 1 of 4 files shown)


marcogario:
These two commands should fit in our parser (I just had problems figuring out how):

* ```(echo <string>)```
* ```(check-sat-assuming (...))```

However, the following is probably not supported by pySMT:

* ```(define-fun-rec ....)```
* ```(define-funs-rec ....)```
* ```(define-sort ...)```
* ```(declare-sort ...)```

mikand:
@marcogario : I implemented the echo command (including its execution in the interpreter evaluation). Is there something else to be done in this PR?

mikand:
@marcogario This is ready for feedback

PR #218, opened 2015-11-16
YICES BUGFIX Fixed a failed assertion in Yices push method

mikand:
Yices does not support a push after a check_sat call that returned
UNSAT. I added a wrapping logic to handle this case that can arise in
practical applications.

### DIFF
--- a/pysmt/solvers/yices.py
+++ b/pysmt/solvers/yices.py
@@ -81,6 +81,7 @@ class YicesSolver(Solver, SmtLibBasicSolver, SmtLibIgnoreMixin):
         self.converter = YicesConverter(environment)
         self.mgr = environment.formula_manager
         self.model = None
+        self.failed_pushes = 0
         return
 
     @clear_pending_pop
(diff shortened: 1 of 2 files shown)


pysmt/solvers/yices.py
```diff
@@ -137,13 +138,23 @@ def all_sat(self, important, callback):
     def push(self, levels=1):
         for _ in xrange(levels):
             c = libyices.yices_push(self.yices)
-            assert c == 0
+            if c != 0:
+                # 4 is STATUS_UNSAT
+                assert libyices.yices_context_status(self.yices) == 4
```
marcogario: Maybe ```InternalSolverError``` ?

marcogario:
Ok, lets convert the assertions into proper exceptions. Then you can merge the code.

mikand:
Done.

mikand:
Fixed and

mikand:
Squashed and fixed

PR #222, opened 2015-11-19
UndefinedSymbolError exception to be used when trying to acces an undefined Symbol

marcogario:
This makes the API of FormulaManager.get_symbol() nicer: instead of returning None, we raise a proper exception. Since this is a dedicated exception, code that was depending on the return value being None, can be updated with a try.


pysmt/formula.py
```diff
@@ -105,20 +105,20 @@ def new_fresh_symbol(self, typename, base="FV%d"):
...
             return self.symbols[name]
-        return None
+        except KeyError:
+            raise UndefinedSymbolError(name)
 
     def get_all_symbols(self):
         return self.symbols.values()
 
     def get_or_create_symbol(self, name, typename):
+        if name not in self.symbols:
```
mikand: In this way we are querying the map twice if the symbol exists. This code is quite core so this may have a significant impact.

I propose something like:
```
s = self.symbols.get(name, None)
if s is None:
    return self.create_symbol(name, typename)
...
return s
```
Otherwise (following EAFP) you can use a try.

mikand:
Great! (y)

You can accept upon fixing the note.

pysmt/formula.py
marcogario: Yes, you are right.

Issue #220, opened 2015-11-18
Handle Logic as string in shortcuts/factory 


marcogario:
In some cases it would be easier to pass a string with the name of the logic to shortcuts method. 

If we so so, currently, we get that no solver for the logic exists. This is misleading. 

@agriggio suggests to handle strings automatically.

marcogario:
I think that this mostly relates to ```Solver```, since shortcuts have automated logic selection. Nevertheless, i think that whatever decision we take should be applied uniformly.

mikand:
Agreed, we may pass the logic specified by the user in a function that checks whether it is a logic object or a string for all the top-level functions in pysmt.

marcogario:
The method for this exists and is ```pysmt.logics.get_logic```. My suggestion, would be to have a decorator for this ```@logic_as_string```, that checks if a type is string, and if so calls ```get_logic```.

I'd avoid requiring that the object is either a string or a logic, it seems over-constraining.

marcogario:
```
def accept_logic_as_string(f):
    """Converts a logic provided as a string into a pysmt.logic object."""

    @wraps(f)
    def accept_logic_as_string_wrap(*args, **kwargs):
        logic = kwargs['logic']
        if logic is not None and type(logic) == str:
            logic = pysmt.logics.get_logic_by_name(logic)
            kwargs['logic'] = logic
        return f(*args, **kwargs)
    return accept_logic_as_string_wrap
```

This was the decorator that I had in mind. However, this requires that (if given) the logic is given as a keyword argument. I think this is a safe assumption, but it seems more complex than necessary.

marcogario:
See #223 .

agriggio:
Hmmm, this seems suboptimal... Now you would get a different
behaviour depending on whether you use keyword or positional
arguments. Also, what if the logic is a unicode object? How about this
version instead (untested, of course :^)?

```python
def accept_logic_as_string(f, pos=None):
    """Converts a logic provided as a string into a pysmt.logic object.
If pos is not None, it indicates the position of the logic argument in *args.
    """
    assert pos is None or pos >= 0

    @wraps(f)
    def accept_logic_as_string_wrap(*args, **kwargs):
        if pos is not None and pos < len(args):
            logic = args[pos]
        else:
            logic = kwargs.get('logic')
        if logic is not None:
            try:
                logic = pysmt.logics.get_logic_by_name(logic)
            except pysmt.logics.UndefinedLogicError:
                pass
[...]

PR #223, opened 2015-11-19
Issue #220: Logic as String

marcogario:
Top level functions from Factory (and thus most shortcuts) can now
take a logic as a string. Instead of writing

  is_sat(f, logic=pysmt.logics.QF_LRA)

it is now possible to write:

  is_sat(f, logic="QF_LRA")

This is handled in Factory._get_solver_class(). Other code within
pySMT should always work under the assumption that logics are given as
a pysmt.logics object.

### DIFF
--- a/pysmt/logics.py
+++ b/pysmt/logics.py
@@ -594,10 +594,18 @@ PYSMT_QF_LOGICS = frozenset(_l for _l in PYSMT_LOGICS if _l.quantifier_free)
 
 def get_logic_by_name(name):
     """Returns the Logic that matches the provided name."""
-
     for logic in LOGICS:
-        if logic.name == name: return logic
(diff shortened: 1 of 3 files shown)


pysmt/logics.py
```diff
@@ -594,11 +594,19 @@ def __hash__(self):
...
-        if logic.name == name: return logic
+        if logic.name.lower() == name.lower(): return logic
     raise UndefinedLogicError(name)
 
+def convert_logic_from_string(name):
+    """Helper function to parse function arguments.
+
+    This takes a logic or a string or None, and returns a logic or None.
+    """
+    if name is not None and type(name) == str:
```
marcogario: In PY2 we need to include unicode as a possible type. This is not a problem for PY3.

PR #225, opened 2015-11-20
Ported to Yices 2.4.1

mikand:
Ported wrapper to Yices 2.4.1


pysmt/solvers/yices.py
```diff
@@ -216,7 +216,7 @@ def get_value(self, item):
             elif ty.is_bv_type():
                 width = ty.width
                 res = (ctypes.c_int32 * width)()
-                libyices.yices_val_get_bv(self.model, ctypes.byref(yval), res)
+                libyices.yices_get_bv_value(self.model, titem, res)
```
marcogario: Why is there this change? Is it a change in the API? Shall we add a regression test for this?

mikand: The old API still exists, but for whatever reason, the old API always returns 0 on the new solver...

PR #226, opened 2015-11-20
BUGFIX: fixed logic dumping in SmtLib

mikand:
The logic to be specified in the script was hardcoded as UFLIRA. Now, it is selected by analyzing the formula using the logic oracle.


marcogario:
Extend the parser tests from examples: pysmt/test/smtlib/test_parser_examples.py

marcogario:
What is the expected SMT-LIB logic for QF_BOOL? We currently pick QF_LIA. I was wondering whether QF_UF would be a better choice.

Issue #205, opened 2015-10-30
Possible regression in CVC4


marcogario:
With the latest version of CVC4 (b455f5cde8b84b7951d309604b75a76afd8b8bfa) tests on SMT-LIB are going into timeout. This is an example:

```
$ time python shell.py --file pysmt/test/smtlib/small_set/QF_LIA/prp-23-47.smt2 
CVC4: Calling Check-SAT
unsat

real	0m31.337s
user	0m31.236s
sys	0m0.088s

$ time .smt_solvers/CVC4_bin/bin/cvc4 pysmt/test/smtlib/small_set/QF_LIA/prp-23-47.smt2 
unsat
real	0m0.516s
user	0m0.496s
sys	0m0.020s
```

The parsing and conversion towards CVC4 is fast. The time is spent in the solving. Unfortunately, I cannot reproduce the problem when going through an SMT-LIB file.

I think this might have to do with new defaults for CVC4.

marcogario:
As a side remark, we should compile CVC with ''--best --enable-gpl"

marcogario:
I tested with the current configuration, and the problem is present also there. Therefore.

- I will proceed to upgrade CVC4 to the latest version
- We need to better understand the options that are set by default from the command-line (i.e., the example above seems to use ITE-Simplification, that is disabled by default).

marcogario:
I identified the regression has been introduced in https://github.com/CVC4/CVC4/commit/d101e7fed051685673c13317cb45166ba5ef7798 .

During incremental solving (I guess API calls follow this rule) the ITE simplification is disabled. 
Maybe there is a way to re-enable this from configuration.

mikand:
This is related to #233 . We should update CVC4 version as sson as possible.

marcogario:
This seems to be fixed with #250 . In #251 we upgrade to the latest CVC4 version.

PR #227, opened 2015-11-21
Fixed CVC4 Installer to use binding dir for CVC4_bin

marcogario:
This solves the caching problem in Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -109,7 +109,6 @@ matrix:
     - python: pypy3
       env: PYSMT_SOLVER="all"
 
-
 install:
   # This is for btor that fails to find the python 3 library to link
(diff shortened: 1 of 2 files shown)


mikand:
Great! :+1: 

But it seems that Z3 and Yices are not cached as well, shall we handle this here or in another PR?

mikand:
ANother note: maybe we can add the compilation flag you reported in #205

marcogario:
The main goal was to reduce the time it takes to perform CI. Now it takes at most 5 minutes. I'd suggest that we look at the other cases in a different PR.
Also, note that caches are divided by python version, so I am wondering if the hack done for boolector is really needed.

Also for the flag, I am opening a new branch, with an updated version of CVC4.

marcogario:
I am closing this, so that the next PRs can benefit from faster building times.

PR #229, opened 2015-11-26
BV Infix notation

marcogario:
BV: Support for BV Infix notation

Tests explain quirks related to mixing python constants and pySMT
constants.
Moreover, infix operators are reserved to unsigned operations. Signed
operators (and operators that cannot easily be mapped into python
operators) need to rely on the l.Operator(r) syntax.

Apart from this, the rest of the infix notation is straight-forward

### DIFF
--- a/pysmt/fnode.py
+++ b/pysmt/fnode.py
@@ -541,9 +541,17 @@ class FNode(object):
         return self._content.payload
 
     # Infix Notation
-    def _apply_infix(self, right, function):
+    def _apply_infix(self, right, function, bv_function=None):
         if _env().enable_infix_notation:
             mgr = _mgr()
+            # BVs
+            # Default bv_function to function
+            if bv_function is None: bv_function = function
+            if _is_bv(self):
+                if is_python_integer(right):
+                    right = mgr.BV(right, width=self.bv_width())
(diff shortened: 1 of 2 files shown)


pysmt/fnode.py
```diff
@@ -660,3 +760,8 @@ def _env():
 def _mgr():
     """Aux function to obtain the formula manager."""
     return pysmt.environment.get_env().formula_manager
+
+def _is_bv(node):
+    """Aux function to check if a fnode is a BV."""
+    return (node.is_symbol() and node.symbol_type().is_bv_type()) or \
```
mikand: This should be correct, but why not relying on the type checker?

mikand:
Sounds good! I just added a comment, but we can merge this as soon as you want

pysmt/fnode.py
marcogario: To avoid operations that rely on the global environment. Here the node does not know to which env it belongs to. This is shorter (measured as indirection of the call) and independent from the environment.

mikand: OK for now, but we shall consider a "shallow" type-checker that only returns the type of the current node without recursion to check type soundness...

PR #240, opened 2016-04-04
Sorting networks

cmattarei:
Added support to sorting networks routines.


pysmt/test/test_formula.py
```diff
@@ -409,15 +410,16 @@ def test_function(self):
...
-    def test_0arity_function(self):
-        # Calling FunctionType on a 0-arity list of parameters returns
-        # the type itself.
-        t = FunctionType(REAL, [])
-        # After this call: t = REAL
-        # s1 is a symbol of type real
-        s1 = self.mgr.Symbol("s1", t)
-        s1b = self.mgr.Function(s1, [])
-        self.assertEqual(s1, s1b)
+    # @skip
```
mikand: Why commenting-out this test?

pysmt/test/test_formula.py
```diff
@@ -938,6 +940,49 @@ def test_typing(self):
...
+            # Create the Expressions
+            at_least_5 = self.mgr.AtLeastK(5, symbols)
+            at_least_10 = self.mgr.AtLeastK(10, symbols)
+            at_most_5 = self.mgr.AtMostK(5, symbols)
+            at_most_10 = self.mgr.AtMostK(10, symbols)
+            exactly_5 = self.mgr.ExactlyK(5, symbols)
+            exactly_10 = self.mgr.ExactlyK(10, symbols)
+
+            # Randomly pick some symbols
+            shuffle(symbols)
```
mikand: It would be better to have deterministic tests...

pysmt/sorting_networks.py
```diff
@@ -0,0 +1,333 @@
...
+        Given formulae = phi1, phi2, ..., phiN, and defining the operator
+        true_evals(formulae) that returns the number of functions that evaluates to true.
+        The result of sorting_network(formulae) = gamma1, gamma2, ..., gammaN where
+        gamma1 <-> (true_evals(formulae) >= 1),
+        gamma2 <-> (true_evals(formulae) >= 2),
+        ...
+        gammaN <-> (true_evals(formulae) >= N),
+        """
+        key = tuple(formulae)
+        return self._cache.setdefault(key, self._sorting_network(formulae))
```
mikand: You should check if ```key``` is in ```self._cache``` first, otherwise you uselessly compute the sorting network in any case!

(The actual parameters are computed before calling functions...)

marcogario:
@cmattarei can you push the latest version of this code + tests?

PR #247, opened 2016-04-20
Basic performance optimizations

marcogario:
1. Avoid use of len() to test emptiness of a dictionary
   (DagWalker._get_key) and a list (DagWalker._process_stack)
   This leads to 75% less calls to len() and a 10% runtime improvement.

2. Use __slots__ for FNode. This leads to 30% memory reduction.

### DIFF
--- a/pysmt/test/test_formula.py
+++ b/pysmt/test/test_formula.py
@@ -610,7 +610,11 @@ class TestFormulaManager(TestCase):
 
         self.assertEqual(str(f), "(A & (! B))", str(f))
 
-        serialized = pickle.dumps(f)
+        # NOTE: We cannot use the textual format for pickle, because
+        # we are using slots. We can, however, use more advanced
+        # versions of pickle. Therefore, we specify here to use the
+        # latest protocol.
+        serialized = pickle.dumps(f, pickle.HIGHEST_PROTOCOL)
 
         f_new = pickle.loads(serialized)
         f_new = dst_mgr.normalize(f)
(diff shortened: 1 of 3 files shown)


marcogario:
This has been tested on external tools. The following simple script can be used to check the optimizations.

```
from pysmt.shortcuts import FreshSymbol, Symbol, And, Or

def main(max_i):
    q = 4
    f = Symbol("x")
    for i in xrange(max_i):
        g = And([FreshSymbol() for _ in xrange(q)])
        f = Or(f, g, Symbol(str(i)))
    print(f.size())
    return f

if __name__ == "__main__":
    main(20**4)
```

**NOTE**: The use of ```__slots___``` makes it impossible to inject new methods within FNode. This might break code that was relying on that (e.g., pyTS). However, the proper way of doing so, is subclassing FNode and implementing the new methods.

pysmt/test/test_formula.py
```diff
@@ -610,7 +610,7 @@ def test_pickling(self):
 
         self.assertEqual(str(f), "(A & (! B))", str(f))
 
-        serialized = pickle.dumps(f)
+        serialized = pickle.dumps(f, -1)
```
mikand: ```-1``` here is confusing, you should use HIGHEST_PROTOCOL (https://docs.python.org/2/library/pickle.html#pickle.dumps)

But anyway, why is this needed?

marcogario: This is needed because the use of slots is incompatible with the use of the oldest version of pickle's protocol.

PR #250, opened 2016-05-07
Improved handling of options in Solvers.

marcogario:
!!! Breaking change !!!
Options are now free keyword arguments in the Solver() constructor.

- The parameter ```user_options``` has been removed
- The function ```get_default_options``` has been removed
- The class SolverOptions is used to validate these options. The rest of option handling has been simplified accordingly.

Calls to Factory methods now set meaningful defaults:

- Calls to Factory.is_* are setting the options for model generation and incrementality to False
- Calls to Factory.get_model are done enabling model generation, but disabling incrementality
- Factory.Solver, returns a solver with both model generation and incrementality enabled

### DIFF
--- a/pysmt/factory.py
+++ b/pysmt/factory.py
@@ -78,7 +78,7 @@ class Factory(object):
         self._get_available_interpolators()
 
 
-    def get_solver(self, quantified=False, name=None, logic=None):
+    def get_solver(self, quantified=False, name=None, logic=None, options=None):
         assert quantified is False or logic is None, \
             "Cannot specify both quantified and logic."
 
@@ -95,7 +95,7 @@ class Factory(object):
 
         return SolverClass(environment=self.environment,
                            logic=closer_logic,
-                           user_options={"generate_models" : True})
+                           user_options=options)
 
 
     def get_unsat_core_solver(self, quantified=False, name=None,
@@ -416,10 +416,14 @@ class Factory(object):
     ##
     ## Wrappers: These functions are exported in shortcuts
     ##
-    def Solver(self, quantified=False, name=None, logic=None):
+    def Solver(self, quantified=False, name=None, logic=None, options=None):
+        if options is None:
+            options={"generate_models" : True,
+                     "incremental" : True}
         return self.get_solver(quantified=quantified,
                                name=name,
-                               logic=logic)
+                               logic=logic,
+                               options=options)
 
     def UnsatCoreSolver(self, quantified=False, name=None, logic=None,
                         unsat_cores_mode="all"):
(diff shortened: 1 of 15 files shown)


pysmt/factory.py
```diff
@@ -416,10 +416,14 @@ class field or if a logic subsuming the given logic is
     ##
     ## Wrappers: These functions are exported in shortcuts
     ##
-    def Solver(self, quantified=False, name=None, logic=None):
+    def Solver(self, quantified=False, name=None, logic=None, options=None):
```
mikand: Maybe we can allow the use of `**kwargs` to specify something like:
```python
Solver(logic=logic, incremental=False)
```
instead of

```python
Solver(logic=logic, options={"incremental": False})
```

??

pysmt/solvers/solver.py
```diff
@@ -59,18 +61,8 @@ def is_sat(self, formula):
         assert formula in self.environment.formula_manager, \
                "Formula does not belong to the current Formula Manager"
 
-        use_solving_under_assumption = False
```
mikand: Why can't we rely on solving under assumptions instead of breaking the contract of `is_sat`? 

If a solver does not support incrementality nor solving under assumptions, then an error can be thrown...

mikand:
OK for the option handling, but I'm not a fan of the new `is_sat` function.

The whole idea behind `s.is_sat(formula)` was to easily check the consistency of a formula wrt the state of the solver. With the new version it is equivalent to:
```python
s.add_assertion(formula)
s.solve()
```

The question is, why having a function to serve the purpose of 2 lines of pysmt?

pysmt/factory.py
marcogario: Yes, that would probably be better.

marcogario:
- ```Solver.solve(x)``` has exactly the same behavior (or should) than ```is_sat```. I think the main question is whether ```is_unsat``` and ```is_valid``` make sense with the semantic that preserves the state of the solver.

- I don't really like these methods (yes, I know I was probably the one to introduce them :) ). I think I have never used them in real code: you need to be careful about what you assert and what not. Explicitly calling push/pop is usually cleaner and more readable. However, I see how this PR does not address this point.

I will revert the changes to ```Solver.is_sat```. The problem of whether to keep these methods remains, but we can discuss it elsewhere.

mikand:
I really like the new features of this branch. In particular I LOVE the solver-specific options:
```python
Solver(name="bdd", dynamic_reordering=True)
```

One further step would be to pass (if possible) general options to the underlying solver:
```python
s = Solver(name="msat", incremental=True, config_options={"theory.arr.enabled": "false"})
```
but we can leave this for a future PR.

pysmt/test/test_bdd.py
```diff
@@ -131,8 +130,8 @@ def test_reordering_algorithms(self):
         from pysmt.solvers.bdd import BddSolver, BddOptions
         for algo in BddOptions.CUDD_ALL_REORDERING_ALGORITHMS:
             with BddSolver(get_env(), pysmt.logics.BOOL,
```
mikand: This could now become
```python
with Solver(name="bdd", dynamic_reordering=True) as s:
```

mikand:
For me this can be merged :+1:

PR #251, opened 2016-05-07
Upgrade CVC4 to c15ff4

marcogario:


marcogario:
@mikand This depends on #250 . To avoid the regression in CVC4, we had to force the use of non-incremental solving.

marcogario:
After talking with @barretcw, the regression has been resolved on CVC's side, in cvc4/c15ff43597b41ea457befecb1b0e2402e28cb523 .

I will proceed to update the version of CVC4, so that this does not depend on #250 anymore.

PR #252, opened 2016-05-07
Enable Quantifiers in CVC4

marcogario:


pysmt/solvers/cvc4.py
```diff
@@ -392,6 +394,6 @@ def _rename_bound_variables(self, formula, variables):
         new_var_list = [mkBoundVar(x.symbol_name(),
                                    self._type_to_cvc4(x.symbol_type())) \
                         for x in variables]
-        old_var_list = [self.walk_symbol(x, []) for x in variables]
+        old_var_list = [self.walk_symbol(x) for x in variables]
```
mikand: Why is the signature of walk_symbol different from the other walk_* methods? This is uncommon in the walkers and shall be uniformed. (I guess it works just because we always pass args with the keyword arguments while tranversing, hence it is digested by the **kwargs patrameter...)

marcogario:
@mikand can I merge this?

PR #254, opened 2016-05-09
Added String Theorem

sureshgl:
Please kindly review my changes.

### DIFF
--- a/pysmt/logics.py
+++ b/pysmt/logics.py
@@ -35,7 +35,8 @@ class Theory(object):
                  integer_difference = False,
                  real_difference = False,
                  linear = True,
-                 uninterpreted = False):
+                 uninterpreted = False,
+                 strings=False):
         self.arrays = arrays
         self.bit_vectors = bit_vectors
         self.floating_point = floating_point
@@ -45,6 +46,7 @@ class Theory(object):
         self.real_difference = real_difference
         self.linear = linear
         self.uninterpreted = uninterpreted
+        self.strings = strings
 
         return
 
@@ -58,7 +60,11 @@ class Theory(object):
         res = self.copy()
         res.linear = value
         return res
-
+       
+    def set_string(self,value=True):
+        res = self.copy()
+        res.strings=value
+        
     def set_difference_logic(self, value=True):
         res = self.copy()
         if res.integer_arithmetic:
@@ -76,7 +82,8 @@ class Theory(object):
                             integer_difference = self.integer_difference,
                             real_difference = self.real_difference,
                             linear = self.linear,
-                            uninterpreted = self.uninterpreted)
+                            uninterpreted = self.uninterpreted,
+                            strings = self.strings)
 
         return new_theory
 
@@ -110,7 +117,8 @@ class Theory(object):
             integer_difference=integer_difference,
             real_difference=real_difference,
             linear=self.linear | other.linear,
-            uninterpreted=self.uninterpreted | other.uninterpreted)
+            uninterpreted=self.uninterpreted | other.uninterpreted,
+            strings=self.strings|other.strings)
 
     def __eq__(self, other):
         if other is None or (not isinstance(other, Theory)):
@@ -124,7 +132,8 @@ class Theory(object):
             self.integer_difference == other.integer_difference and \
             self.real_difference == other.real_difference and \
             self.linear == other.linear and \
-            self.uninterpreted == other.uninterpreted
+            self.uninterpreted == other.uninterpreted and \
+            self.strings == other.strings
 
     def __ne__(self, other):
         return not (self == other)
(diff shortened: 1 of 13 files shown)


mikand:
Hi @sureshgl !

First of all, thanks for contributing to pySMT!

Your PR sounds interesting, and relates to isue #254, but currently it does not work.

It seems that you forgot an import in the `typings.py` module:
$ python install.py --check
  File "install.py", line 18, in <module>
    pysmt.cmd.install.main()
  File "/home/travis/build/pysmt/pysmt/pysmt/cmd/install.py", line 179, in main
    check_installed([x.InstallerClass.SOLVER for x in solvers_to_install])
  File "/home/travis/build/pysmt/pysmt/pysmt/cmd/install.py", line 63, in check_installed
    from pysmt.shortcuts import Solver, QuantifierEliminator
  File "/home/travis/build/pysmt/pysmt/pysmt/shortcuts.py", line 33, in <module>
    import pysmt.typing as types
  File "/home/travis/build/pysmt/pysmt/pysmt/typing.py", line 302, in <module>
    PYSMT_TYPES = frozenset([BOOL, REAL, INT, STRING])
NameError: name 'STRING' is not defined

You should be able to see the results of our CI service that runs the test suite and reports errors (https://travis-ci.org/pysmt/pysmt).

One suggestion, you can run the tests locally on your machine to ensure everything works before committing as follows:
```
$ nosetests pysmt
```
from the top-level repo directory (i.e. where install.py is located)

Let me know if you need further support from our side,

thanks,

A.

pysmt/typing.py
```diff
@@ -279,12 +279,25 @@ def __ne__(self, other):
...
+            return "() String"
+        else:
+            return "String"
+    def __str__(self):
+        return "String"
 
 # Singletons for the basic types
 BOOL = BooleanType()
 REAL = RealType()
 INT = IntType()
```
mikand: Probably the error our CI signals is here:

you need to add a line
```python
STRING = StringType()
```
Otherwise the constant 'STRING' in line 302 is undefined.

pysmt/logics.py
```diff
@@ -594,7 +615,7 @@ def __hash__(self):
 PYSMT_LOGICS = frozenset([QF_BOOL, QF_IDL, QF_LIA, QF_LRA, QF_RDL, QF_UF, QF_UFIDL,
                           QF_UFLIA, QF_UFLRA, QF_UFLIRA,
                           BOOL, LRA, LIA, UFLIRA, UFLRA,
-                          QF_BV, QF_UFBV])
+                          QF_BV, QF_UFBV, QF_SLIA])
```
marcogario: This implies that all solvers that currently are marked as supporting PYSMT_LOGICS, need to be extended to support strings (MathSAT and Z3 for sure).
There are two options here:

* Do not put QF_SLIA in PYSMT_LOGICS, and only mark CVC4 as supporting it
* Explicitely remove QF_SLIA in all other solvers ```LOGICS``` field.

I'd prefer the first one, since so far only CVC4 supports Strings.

pysmt/logics.py
```diff
@@ -548,6 +560,14 @@ def __hash__(self):
               linear=False,
               uninterpreted=True)
 
+QF_SLIA = Logic(name="QF_SLIA",
+              description=\
+""" Don't know clearly what to write here //GL """,
```
marcogario: You can copy the description of the logic provided by SMT-LIB

pysmt/test/test_string.py
```diff
@@ -0,0 +1,38 @@
...
+from pysmt.printers import  HRPrinter
+from pysmt.typing import INT,STRING
+
+class TestSTRING(TestCase):
+
+    @skipIfNoSolverForLogic(QF_SLIA)
+    def test_string(self):
+        s1 = Symbol("s1",STRING)
+        s2 = Symbol("s2",STRING)
+        cvc4SMTSolver = Solver(name="cvc4", logic=QF_SLIA)
```
marcogario: There is no need to specify the name of the solver. Right now you need to, because of the fact that ```QF_SLIA``` has been added to ```PYSMT_LOGICS``` (see related comment), and therefore, this test runs also for MathSAT and Z3. Moreover, there are some methods in logics.py that still need to be extended to deal with strings (e.g., ```get_logic```). Once that is done, the fact that you marked the test with ```skipIfNoSolverForLogic``` is enough to be sure that only CVC4 (and in the future other solvers supporting Strings) will be tested.

For the same reason, rename the variable to simply ```solver```

pysmt/test/test_string.py
```diff
@@ -0,0 +1,38 @@
...
+        cvc4SMTSolver.add_assertion(Not(Equals(s1,s2)))
+        cvc4SMTSolver.add_assertion(Equals(Length(s2), Length(s1)))
+        res = cvc4SMTSolver.solve()
```
marcogario: Indeed, this raises an error in CVC4, however, if you only assert the previous statement, everything works fine. Most likely, there is a problem in the way the conversion of ```length``` is done in ```CVC4Converter```.

marcogario:
I am closing this PR. All future work on strings should be based off the new branch ```strings```. In that branch I took the changes from here, fixed the failing tests and addressed most of the initial comments.

@sureshgl , open a new PR using the ```strings``` branch as destination for implementing all other strings operators.

PR #256, opened 2016-05-10
Integrating theory of arrays.

satyauppalapati:
Integrated theory of arrays. Added ARRAY_INT, ARRAY_REAL, ARRAY_BOOL types and added 'select' and 'store' APIs.

### DIFF
--- a/pysmt/fnode.py
+++ b/pysmt/fnode.py
@@ -36,10 +36,11 @@ from pysmt.operators import (FORALL, EXISTS, AND, OR, NOT, IMPLIES, IFF,
                              BV_SLT, BV_SLE,
                              BV_COMP,
                              BV_SDIV, BV_SREM,
-                             BV_ASHR)
+                             BV_ASHR,
+                             STORE, SELECT)
 from pysmt.operators import  (BOOL_OPERATORS, THEORY_OPERATORS,
                               BV_OPERATORS, LIRA_OPERATORS,
-                              RELATIONS, CONSTANTS)
+                              RELATIONS, CONSTANTS, ARRAY_OPERATORS)
 from pysmt.typing import BOOL, REAL, INT, BVType
 from pysmt.decorators import deprecated
 from pysmt.utils import is_python_integer, is_python_rational, is_python_boolean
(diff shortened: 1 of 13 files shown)


marcogario:
Thank you very much for contributing to pySMT! We have been meaning to add arrays for a while!

Unfortunately, your current code fails the automated tests on Travis CI. You can see that there is a red cross next to your commit, and from there you can see the link to the CI result for this commit:

https://travis-ci.org/pysmt/pysmt/builds/129210704

You should also be able to run most of the tests using the command ``` $ nosetests pysmt``` from the top-level of the repo. If you have doubts on how to fix some of these errors, let us know. Thanks!

marcogario:
Could you please take a look at how the typing information was implemented in 
https://github.com/ahmedirfan1983/pysmt/blob/array/pysmt/typing.py

This was an old proposal for Arrays, but I think there is quite a bit of useful insight there. In particular, the fact that Arrays in SMT are binary types, having an index type and a value type. Notice that this allows to define nested arrays.

satyauppalapati:
Thanks Marco for providing the reference. I have actually started with the
same idea but changed the code later. I will update the code and push the
changes.

Thanks,
Satya

marcogario:
Hi,

Can you please put this activity on hold for 3-5 days? I will open a dedicated branch, and merge the contribution of @agriggio and @ahmedirfan1983. They already did quite a bit of work, and I think this could actually speed up the process.

Thanks,
Marco

https://github.com/agriggio/pysmt/tree/arrays

satyauppalapati:
Sure. I will wait for the update.

marcogario:
@satyauppalapati I am going to close this PR. Please work off the ```array``` branch (see PR #258 ) and make a new PR targeting that branch, with the additional code for cvc4 support.

Notice that tests on that branch are currently failing because we do not have array constants yet. That should not be a big problem to implement the CVC4 conversion and solving. 
Please note that the ArrayType is defined as a binary type, and the CVC4Converter needs to take this into account.

If you have any other interesting example to add, please go ahead.
Thanks!

PR #257, opened 2016-05-16
Fixed bug in Non-linear theories comparison

marcogario:
Linearity information was being incorrectly compared and propagated in logics oracles.

### DIFF
--- a/pysmt/logics.py
+++ b/pysmt/logics.py
@@ -102,15 +102,15 @@ class Theory(object):
             real_difference = False
 
         return Theory(
-            arrays=self.arrays | other.arrays,
-            bit_vectors=self.bit_vectors | other.bit_vectors,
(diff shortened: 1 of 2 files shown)


pysmt/logics.py
```diff
@@ -109,7 +109,7 @@ def combine(self, other):
             real_arithmetic=self.real_arithmetic | other.real_arithmetic,
             integer_difference=integer_difference,
             real_difference=real_difference,
-            linear=self.linear | other.linear,
+            linear=self.linear & other.linear,
```
mikand: why bitwise `&` instead of logic `and`?

mikand: Also, why using bitwise operator in the other cases?

marcogario: The answer for the first question is the second question ;)
Ok to change them to logical.

PR #258, opened 2016-05-16
Arrays Theory

marcogario:
Provide support for arrays for the following solvers:

* MathSAT
* Z3
* CVC4
* Boolector

This is based off the fork of @agriggio. Currently, some tests fail because we are missing support for constants. ~~Other tests are been skipped, but a loud warning (print) has been added, in order to re-enable them.~~

At the currrent stage, the following are missing:

- [x] Array Constants
- [x] Support for CVC
- [x] Support for Boolector
- [x] HR Parsing
- [x] Clarify wheter the logic QF_AX should be supported (how are indexes and values expressed?)


mikand:
I briefly discussed the issue of Constant Arrays with @agriggio and we came up with the following.

Internally we introduce a new constant type ARRAY_CONST that is a unary operator having as payload the index type and the only child is the constant value of each cell of the array. Alternatively, we can store as payload the ArrayType itself to vaoid reconstruction.

We can think of a constructor ```Array(index_type, constant_value)```.

mikand:
NOTE: I fixed some tests and the parsing of HR formulae and rebased the whole PR on master to avoid conflicts.

marcogario:
Ok, for these constant arrays. I guess they are needed to have a meaningful building block. I am still a bit puzzled on whether we need/want to do something smarter to represent the array in a model.

In particular, I am not clear on what happens when you get a model for the following:
```
p > 5
q = 0
a1 = Store(Array(INT, Int(0)), q, p)
```

I think that we want to be able to extract the value of a cell of a1. Consider the corresponding behavior for Int/Real: ```FNode.constant_value()``` provides us with an object that can be used within the python code. Where is the equivalent for Arrays? I expect to be able to do sth like: ```a1.constant_value()[0] == 6```

In the quantifier free fragment, can we always model an array as a default value + the cells that changed? If so, a naive approach probably needs to store the indexes that do not coincide with the default, leading to some dict-like structure.

Of course, we can also decide that arrays constants are never extracted directly, and need to be evaluated against a model: i.e., ```model.get_value(Select(a1, 0))```. However, I'd keep this as last resort option.

mikand:
@marcogario and myself discussed the constant array representation at length today. Here is a summary of the decision.

We will introduce a n-ary node ARRAY_VALUE. The node has at least one child that is the default value of the array and has the index type as payload. Then, it might have an even number of additional children. Each pair of such additional children represent a store in the array value: the first element being the index, the second being the expression. All the indexes are forced to be constants.

We imagined a constructor as follows:
```python
Array(idx_type, default_value, assigned_values=None)
```  
that allows for declaring a constant array.

For example, 
```python
a = Array(INT, Real(4), {Int(1) : Symbol(x, Real), Int(3) : Real(7)})
```
is equivalent to
```python
a = Store(Store(Array(INT, Real(4)), Int(1), Symbol(x, Real)), Int(3), Real(7))
```

The type-checking of the node is as follows. We require the children in odd-position in the children list to have the same type and all the even-positioned children to have the same type as the index type stored in the node payload. The returned type is an array type with index type as per the payload and value type as the first child.

All the walkers and utilities shall be updated to support this new node.

pysmt/fnode.py
```diff
@@ -142,6 +143,16 @@ def is_constant(self, _type=None, value=None):
         Optionally, check that the constant is of the given type and value.
         """
         if self.node_type() not in CONSTANTS:
+            if self.node_type() == ARRAY_VALUE:
+                # An array value can be a constant if all its children
+                # are constants
+                for c in self.args():
+                    if not c.is_constant():
+                        return False
+                if _type is not None or value is not None:
```
mikand: Type can be supported by means of FNode.get_type()

pysmt/logics.py
```diff
@@ -602,8 +602,13 @@ def __hash__(self):
 PYSMT_LOGICS = frozenset([QF_BOOL, QF_IDL, QF_LIA, QF_LRA, QF_RDL, QF_UF, QF_UFIDL,
                           QF_UFLIA, QF_UFLRA, QF_UFLIRA,
                           BOOL, LRA, LIA, UFLIRA, UFLRA,
-                          QF_BV, QF_UFBV])
+                          QF_BV, QF_UFBV,
+                          QF_ABV, QF_AUFBV, QF_AUFLIA, QF_ALIA
+                          #MG: QF_AX?
```
mikand: @agriggio What is QF_AX? If we only have arrays with no additional theory for indexes isn't it a SAT problem?

pysmt/solvers/yices.py
```diff
@@ -105,8 +105,13 @@ def add_assertion(self, formula, named=None):
 
     def get_model(self):
         assignment = {}
+        # MG: This iteration is probelmatic, since it assumes that all
```
mikand: @marcogario Open an issue on this...

pysmt/logics.py
agriggio: No, it's not a SAT problem. You can have uninterpreted sorts for both the index and the element. QF_AX contains SMT formulas only the theory of arrays only. Example:

```
(declare-sort Index 0)
(declare-sort Element 0)
(declare-fun a () (Array Index Element))
(declare-fun b () (Array Index Element))
(declare-fun i () Index)
(declare-fun e () Element)
(declare-fun e2 () Element)
(assert (= (store a i e) b))
(assert (not (= (select a i) e)))
(assert (= a b))
(check-sat)
```

pysmt/test/test_solving.py
```diff
@@ -239,18 +243,25 @@ def do_model(self, solver_name):
 
     @skipIfSolverNotAvailable("cvc4")
     def test_model_cvc4(self):
+        return
```
mikand: Remove this

pysmt/test/test_solving.py
```diff
@@ -173,6 +173,9 @@ def test_examples_cvc4(self):
         for (f, validity, satisfiability, logic) in get_example_formulae():
             try:
                 if not logic.quantifier_free: continue
+                if logic.theory.arrays:
```
mikand: This could be handled as a NoSolverAvailableError exception as CVC4 does not support (for now) arrays

pysmt/typing.py
```diff
@@ -280,6 +284,91 @@ def __hash__(self):
...
+        return True
+
+    def __eq__(self, other):
+        if other is None:
+            return False
+        if self.type_id != other.type_id:
+            return False
+        if id(self) == id(other):
+            return True
+        return str(self) == str(other)
```
mikand: This could be handled by recursive application of equality on the index and value types

pysmt/typing.py
```diff
@@ -280,6 +284,91 @@ def __hash__(self):
...
+
+    def __ne__(self, other):
+        if other is None:
```
mikand: This could be simply rewritten as ```not self.__eq__(other)```

marcogario:
BTOR supports only Arrays of type <BV,BV>.

Both BTOR and CVC4 do not support constant arrays. CVC4 has an operator ```STORE_ALL``` that is similar to a const array, but according to the API documentation (http://cvc4.cs.nyu.edu/builds/documentation/internals/latest-unstable/namespaceCVC4_1_1kind.html) it is not supported when creating a formula and it is used only to return models.

mikand:
I am done with the implementation part on MathSAT and Z3.
I finished the SMT/HR printing and parsing and fixed several bugs.

marcogario:
I rebased these commits!!!

Notice that I disabled one of the examples from QF_ABV, because it is incompatible with the current CVC4 wrapper #265 . The same example was causing problems with btor #264, but that was easier to fix.

I also introduced the extended Logic for Arrays+ Arrays Constants. To be discussed if we want a better name. It seems to me that we might need to refactor ```logics.py``` to deal with the combinatorial explosion of adding new theories.

marcogario:
@mikand, @agriggio can you complete the CLA at https://www.clahub.com/agreements/pysmt/pysmt ? Thanks!

@mikand please review the changes once more. I think this is ready to be merged in master.

pysmt/solvers/btor.py
```diff
@@ -125,6 +125,9 @@ def get_value(self, item):
         titem = self.converter.convert(item)
         if item.is_symbol() and item.symbol_type().is_bv_type():
             return self.mgr.BV(titem.assignment, item.bv_width())
+        elif item.is_symbol() and item.symbol_type().is_array_type():
+            # Array constant without a base value
```
mikand: Why we return None? Isn't there a way to obtain an array model in BTOR?

mikand:
@marcogario This is OK for me now, please check my latest commit and merge...

PR #260, opened 2016-05-22
Strings Theory (Z3)

marcogario:
Support for Strings Theory #245
- For now the only solver supporting this is CVC4

PR #261, opened 2016-05-23
Update README.rst

marcogario:
- Removed bit about using a patch for CVC4. This is not true anymore, since the changes have been incorporated up-stream.
- Clarified when we use wrappers that are not part of the tool itself
- Improved table with theories including Arrays
- Remove pypi download counter badge, since it is not updated anymore

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -14,10 +14,6 @@ pySMT: a Python API for SMT
            :target: https://pypi.python.org/pypi/pySMT/
            :alt: Latest PyPI version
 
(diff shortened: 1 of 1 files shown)


marcogario:
Some doubts about the reference tables remain:

- LIA will be supported by CVC4 after #252. What about LRA?
- Qelim for Mathsat should support also LIA (from one of the most recent versions), we need to update this.
- Arrays #258 and Strings #260 should also appear in the table.

PR #263, opened 2016-05-26
Other String APIs

sureshgl:
Hi Marco,

Please consider the changes, all the test were passing on y branch. I have configured only z3 and cvc4 on my machine.

Also , please not that  SmtDagPrinter.py  changes are pending, and may not be effect.

Regards,
GL.


pysmt/formula.py
```diff
@@ -897,14 +897,66 @@ def BVRepeat(self, formula, count=1):
         for _ in xrange(count-1):
             res = self.BVConcat(res, formula)
         return res
-
-    def Length(self, formula):
-        """ Creates an expression of the form:
-            Length formula
+    
+    def str_length(self, formula):
```
marcogario: For consistency, all methods of Formula should be CamelCase. This should be ```StrLength```. the same applies to all other methods.

pysmt/formula.py
```diff
@@ -897,14 +897,66 @@ def BVRepeat(self, formula, count=1):
...
+        """ Creates an expression of the form Length formula
         """
-        n = self.create_node(node_type=op.LENGTH, args=(formula,))
+        t = self.env.stc.get_type(formula)
```
marcogario: No need to do this here. The typechecker is automatically called within ```create_node```. If the rule in the type checker is correct, then this check will be performed before leaving this function.

pysmt/formula.py
```diff
@@ -897,14 +897,66 @@ def BVRepeat(self, formula, count=1):
...
+        if ( t != types.STRING):
+            raise TypeError("Cannot create a str_length other than STRING argument.")
+            
+        n = self.create_node(node_type=op.STR_LENGTH, args=(formula,))
         return n
+    
+    def str_concat(self, *args ):
```
marcogario: Please document all these functions.

pysmt/formula.py
```diff
@@ -897,14 +897,66 @@ def BVRepeat(self, formula, count=1):
...
+        tuple_args = self._polymorph_args_to_tuple(args)
 
+        if len(tuple_args) <= 1:
+            raise TypeError("Cannot create a Str_Concat without arguments.")
+        else:
```
marcogario: No need to put the else here.

pysmt/shortcuts.py
```diff
@@ -385,6 +383,52 @@ def BVAShr(left, right):
         of steps specified by the right BV."""
     return get_env().formula_manager.BVAShr(left, right)
 
+# String API
+def str_length(s):
```
marcogario: This follows the same naming conventions as Formula, including the documentation from there.

pysmt/simplifier.py
```diff
@@ -593,7 +593,7 @@ def walk_bv_ashr(self, formula, args, **kwargs):
             return ret
         return self.manager.BVAShr(l, r)
 
-    def walk_length(self, formula, args, **kwargs):
-        return self.manager.Length(args[0])
+    def walk_str_length(self, formula, args, **kwargs):
+        return self.manager.str_length(args[0])
```
marcogario: In the simplest form, it should be possible to apply simplification rules on constants. 
For example, length of a constant can by computed via python. Similarly, methods such as concatenation etc.

More complex simplifications can be left as future work, and implemented as needed.

pysmt/solvers/cvc4.py
```diff
@@ -142,6 +144,16 @@ def get_value(self, item):
 
     def _exit(self):
         del self.cvc4
+    
+    def set_option(self, name, value):
```
marcogario: Good! Can you use this within ```reset_assertions``` ?

pysmt/solvers/solver.py
```diff
@@ -241,6 +240,15 @@ def set_options(self, options):
         :type options: Dictionary
         """
         raise NotImplementedError
+    
+    def set_option(self, name, value):
```
marcogario: This should not be here, please remove it. Currently, we do not have a uniform way of passing options to the solver *after* it was created. This needs much more discussion to identify a way of doing it in a uniform way among all solvers.

pysmt/test/examples.py
```diff
@@ -560,20 +560,18 @@ def get_example_formulae(environment=None):
...
                     logic=pysmt.logics.QF_SLIA),
 
             # ( 5 < len(s)  & "mystr" == str1)
-            Example(expr=And(LT(Int(5), Length(str1)),
+            Example(expr=And(LT(Int(5), str_length(str1)),
                              Equals(String("mystr"), str1)),
                     is_valid=False,
                     is_sat=False,
                     logic=pysmt.logics.QF_SLIA),
-
```
marcogario: Each operator should appear in at least one example.

pysmt/test/test_string.py
```diff
@@ -0,0 +1,309 @@
...
+        s2 = Symbol("s2",STRING)
+        solver = Solver(logic=QF_SLIA)
+        solver.add_assertion(  Not( Implies( Equals(s1,s2),
+                                             Equals(str_length(s2), str_length(s1))
+                                            )
+                                   )
+                             )
+        res = solver.solve()
+        if ( res):
+            print solver.get_model()
```
marcogario: We support both python 2 and python 3. In python 3, ```print``` is a function. E.g., rewrite as ```print(solver.get_model())``` (similarly for the other examples below)

pysmt/test/test_string.py
```diff
@@ -0,0 +1,309 @@
...
+        assert not res, "Was expecting  FALSE"
```
marcogario: ```self.assertFalse(res)```

pysmt/test/test_string.py
```diff
@@ -0,0 +1,309 @@
...
+        assert not res, "Was expecting  FALSE" 
+        return
+    
+    
+    @skipIfNoSolverForLogic(QF_SLIA)
+    def test_str_contains(self):
+        s1 = Symbol("s1",STRING)
+        s2 = Symbol("s2",STRING)
+        solver = Solver(logic=QF_SLIA)
+        solver.set_option('strings-exp','true')
```
marcogario: This option is really specific only to CVC4 and, btw, it is already set within the solver. No need to have this here.

pysmt/type_checker.py
```diff
@@ -61,7 +60,25 @@ def __init__(self, env=None):
         self.set_function(self.walk_bv_rotate, op.BV_ROL, op.BV_ROR)
         self.set_function(self.walk_bv_extend, op.BV_ZEXT, op.BV_SEXT)
         self.set_function(self.walk_bv_comp, op.BV_COMP)
-        self.set_function(self.walk_string_length, op.LENGTH)
+        
+        self.set_function(self.walk_identity_string, op.STRING_CONSTANT)
+        self.set_function(self.walk_str_length, op.STR_LENGTH)
```
marcogario: The defaults mappings are defined in ```walkers/generic.py``` and inherited here. Therefore, you do not need to re-state them here. The idea here is that you might be to group operators that have the same type-checking rules together (see for example ```walk_bv_to_bool```).

marcogario:
This looks like a lot of progress! Thank you very much!
I added some comments from a preliminary review, once those are fixed, I think we will be in a very good shape! :smile:

marcogario:
Hi, 

I'll be reviewing this, and merging int into our ```strings``` branch in the next 3-4 days.

Before this goes into master a few things need to happen:
- More tests concerning the SMT-LIB format (although experimental) need to be added.
- Arrays branch needs to be merged into master. This will create a few conflicts with this branch, but I'll take care of addressing them. This merge should happen at some point during the upcoming week. 
- Finally, I'll need you to confirm that you have read and agreed with the APACHE license under which pySMT is released, by signing the CLA at https://www.clahub.com/agreements/pysmt/pysmt

Thanks,
Marco

sureshgl:
Hi Marco,

Thank you for accepting the changes.

Please note that I have signed the CLA.

Eagerly waiting for the master copy to start our prototype.


Regards,
GL.


On Fri, Jun 3, 2016 at 5:41 PM, Marco Gario <notifications@github.com>
wrote:

> Hi,
>
> I'll be reviewing this, and merging int into our strings branch in the
> next 3-4 days.
>
> Before this goes into master a few things need to happen:
>
>    - More tests concerning the SMT-LIB format (although experimental)
>    need to be added.
>    - Arrays branch needs to be merged into master. This will create a few
>    conflicts with this branch, but I'll take care of addressing them. This
>    merge should happen at some point during the upcoming week.
>    - Finally, I'll need you to confirm that you have read and agreed with
>    the APACHE license under which pySMT is released, by signing the CLA at
>    https://www.clahub.com/agreements/pysmt/pysmt
>
> Thanks,
> Marco
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/pysmt/pysmt/pull/263#issuecomment-223562780>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AB5h1J-f0Pys4drHd9Yo9cfjiNNj4Dorks5qIBn3gaJpZM4InN2D>
> .
>



-- 
Suresh GL

marcogario:
Could you please clarify what is the expected difference btw ```StrToInt```, ```StrToInt16``` and ```StrToInt32```? The theory of INTegers is infinite precision. Therefore they should be equivalent if we are talking about SMT-LIB Integers.  I can understand that some numbers cannot be represented in 16 or 32 bits integers, but I think this is mixing different things.

The opposite direction does not actually make sense, since you cannot actually be distinguished between a "Int16" and "Int32" in SMT-LIB, i.e., they are both elements of type INT. 
Can you comment on this? 

Thanks!

pysmt/rewritings.py
```diff
@@ -543,7 +546,7 @@ def walk_quantifier(self, formula, args, **kwargs):
 
     def walk_theory_op(self, formula, **kwargs):
         #pylint: disable=unused-argument
-        return None
+        return [],None
```
marcogario: FYI, @sureshgl,  this is way the prenex rewriter was not working on the examples you commented out.
In python it is ok to have different types for the return value. It is not beautiful, but in selected cases it can make your life easier. In this case, we have a special case to handle (theory formula), that we represent with None. If we return always a tuple, and only set the second element to None, every time we want to check the arguments of the call, we need to do unpacking. Here we know that the first argument is useless if the second is None, and simplify this to None.

I think that you changed this, because you were receiving some error on ```StrContains```, ```StrPrefixOf```, ```StrSuffixOf```. This is because these are not *only* theory operators, but predicates, i.e., they have a Boolean return value. This means that they need to be taken into account when manipulating the boolean structure of the formula.

I fixed it. We probably need to clarify this in the documentation of the rewriter.

marcogario:
Changes were merged to branch ```strings```. Some clean-up will follow. Leaving this open to track discussion.

marcogario:
@sureshgl : I am going to leave only ```StrToInt``` and ```IntToStr``` as operators. The other two cases, are described in the wiki only at the syntax level, but their semantics is not clear. They are also not mentioned within the papers in which the theory is defined (as far as I can see). 

If these are needed, please open a ticket, so that we can try to clarify their semantics.

marcogario:
See #260 for the latest version of this feature

pysmt/rewritings.py
sureshgl: I agree & thank you.

PR #273, opened 2016-06-17
fixed bug in Z3Converter.walk_array_value

agriggio:
The code was broken: `enumerate(args[1::2])` returns `(0, args[1]), (1, args[3]), ...`, and not
`(1, args[1]), (3, args[3]), ...` as the code was expecting


mikand:
Thanks for the patch!

PR #274, opened 2016-06-21
New Yices Wrapper

mikand:
Using yicespy as the new API for accessing the Yices Solver


mikand:
TODO:
- [ ] GMP support
- [ ] Fix wrapper version from github in the installers
- [ ] Test Python 3.5

PR #279, opened 2016-07-08
Compressed all SMT2 test files

marcogario:
The parser recognizes compressed smt2 files if they end with .gz
In this way, we can shrink the directory for small_set from 30MB to 1MB


marcogario:
The problem seems to be related to python 3. The parser returns an empty script when parsing the "wrong.smt" test case.

marcogario:
Note: This disables pypy3 tests, because the version on Travis is too old (2.*) and does not support gzip properly.

pysmt/smtlib/parser.py
```diff
@@ -32,6 +33,15 @@
 from pysmt.utils import interactive_char_iterator
 
 
+def open_(fname):
+    """Transparently handle .bz2 files."""
+    if fname.endswith(".bz2"):
+        if PY2:
+            return bz2.BZ2File(fname, "r")
+        else:
+            return bz2.BZ2File(fname, "rt")
```
mikand: This seems to fail in Python 3.4... (see Travis)

Maybe we need a better distinction on the python version and put this logic in utils?

Issue #270, opened 2016-06-08
Non-Linear Test


marcogario:
I cannot find when did we remove the limitation on Non-Linear expressions, but we did. 

As a matter of fact, the following works:

```
from pysmt.shortcuts import Symbol, Times, Solver
from pysmt.typing import REAL

x = Symbol("x", REAL)
f = (x*x).Equals(x)
s = Solver(name="z3")
s.is_sat(f)
print(s.get_model())
```

Apart from model extraction (within the eager model) how far are we from supporting NRA? What about NIA?

We need to write a test ```test_non_linear.py``` to detail how a user could want to use the Non-linear functionalities of a solver, and use that to measure where the system is not ready. 

- Parts like the oracles are not supporting it, and no solver is defined as supporting non-linear (thus, one needs to explicit the solver to use).
- Operators such as abs, and mod are not available

marcogario:
Z3 can output algebraic solutions in the following format by calling ```sexpr()```:

```
>> x.sexpr()
 '(root-obj (+ (^ x 2) (- 5)) 2)'
```

This format is supported by mpmath to compute a solution:

```
>> import mpmath
>> mpmath.findroot(lambda x: x**2 -5, 2)
```

marcogario:
Solvers that do not support non-linear expressions should raise an exception when such a formula is being asserted.

marcogario:
Agreed to catch this during conversion.

marcogario:
See #282

PR #282, opened 2016-07-16
Support for NRA

marcogario:
Non-linear (polynomial) expressions are now supported by pySMT. This includes the Times, Div and Pow operators.

The only solver currently supporting QF_NRA is z3. For this reason we rely on z3 to represent algebraic solutions (e.g., sqrt(2)). This might be subject to changes in the future, and the constructor ```Algebraic``` is "hidden" in ```FormulaManager```.

This addresses #270.


pysmt/formula.py
```diff
@@ -231,74 +212,77 @@ def Times(self, left, right):
...
         """
-        if not right.is_constant(types.REAL):
-            raise NonLinearError
-        inverse = Fraction(1) / right.constant_value()
-        return self.Times(left, self.Real(inverse))
+        if not exponent.is_constant():
+            raise ValueError("The exponent of POW must be a constant.", exponent)
 
+        if base.is_constant():
+            val = base.constant_value() ** exponent.constant_value()
```
mikand: Why performing this simplification here? In other operators (e.g. Times or Plus) this is not done and is demanded to the simplifier....

pysmt/numeral.py
```diff
@@ -0,0 +1,21 @@
...
+        """Represents a Number (Algebraic)"""
+        def __hash__(self):
+            return hash(self.sexpr())
+
+else:
+    class Numeral(object):
+        """Represents a Number (Algebraic)"""
+        def __init__(self, obj):
+            self.obj = obj
+            warnings.warn("Z3 is not installed. We currently do not support stand-alone algebraic numbers. This might lead to unexpected results")
```
mikand: Why not a NotImplementedError instead of a warning?

pysmt/operators.py
```diff
@@ -55,7 +55,11 @@
 BV_ASHR,                                    # Arithmetic shift right (47)
 ARRAY_SELECT,                               # Array Select (48)
 ARRAY_STORE,                                # Array Store (49)
-ARRAY_VALUE                                 # Array Value (50)
+ARRAY_VALUE,                                # Array Value (50)
+
+DIV,                                        # Arithmetic Division (51)
```
mikand: Is there a reason to explicitly represent Div(a, b) internally instead of using Times(a, Pow(b, -1))? This would reduce the number of operators and simplify the walkers, so unless there is an efficiency or technical reason for having the explicit div I would remove it.

pysmt/printers.py
```diff
@@ -33,6 +36,34 @@ def __init__(self, stream, env=None):
...
+        self.set_function(partial(self._walk_nary, " s% "), op.BV_SREM)
+        self.set_function(partial(self._walk_nary, " s<= "), op.BV_SLE)
+        self.set_function(partial(self._walk_nary, " s< "), op.BV_SLT)
+        self.set_function(partial(self._walk_nary, " u<= "), op.BV_ULE)
+        self.set_function(partial(self._walk_nary, " u< "), op.BV_ULT)
+        self.set_function(partial(self._walk_nary, " << "), op.BV_LSHL)
+        self.set_function(partial(self._walk_nary, " >> "), op.BV_LSHR)
+        self.set_function(partial(self._walk_nary, " a>> "), op.BV_ASHR)
+        self.set_function(partial(self._walk_nary, " bvcomp "), op.BV_COMP)
+
```
mikand: :+1:

pysmt/test/examples.py
```diff
@@ -613,6 +613,31 @@ def get_example_formulae(environment=None):
...
+            Example(expr=Equals(Real(2), Times(r,r)),
+                    is_valid=False,
+                    is_sat=True,
+                    logic=pysmt.logics.QF_NRA),
+
+            # p**2 = 0
+            Example(expr=Equals(Pow(p, Int(2)), Int(0)),
+                    is_valid=False,
+                    is_sat=True,
+                    logic=pysmt.logics.QF_NIA),
```
mikand: So this PR is about NRA and NIA! :bowtie:

pysmt/test/test_nra.py
```diff
@@ -0,0 +1,85 @@
+#
```
mikand: No test for NIA....

mikand:
This looks VERY good! I added few comments in the PR, but generally I love this!

pysmt/formula.py
marcogario: The idea here (maybe implemented incorrectly) is that constant pow is linear. By doing this simplification here, we do not need to deal with walk_pow in the converters of each solver. A solver that does not support pow, will never see the pow operator if we only have constant pow. Otherwise, we need to implement walk_pow in each converter, and check whether we are in a constant case. This issue does not exist with plus or times, because we still need to handle those operators in all LIRA converters.

Does this make sense?

pysmt/numeral.py
marcogario: It seemed too harsh... You should still be able to get a model, as long as you treat the object as an opaque object. As soon as you try to do anything on it, it will break. This should never happen right now, so I am fine with converting it into ```NotImplementedError```.

pysmt/operators.py
marcogario: I'll try to use the Pow(x, -1) and see what happens. If nothing breaks, I'll remove DIV.

marcogario: Ok, I think this boils down to 2 things:

1) Support for NIA
2) Printing

If we support NIA (that I think is reasonable), then we need to understand what it means to write ```2/x``` and whether we can write ```x^-c``` (negative exponent). From a quick test, it looks like Z3 can solve the first one without problems, while not the second one.

For printing, I think that x/y is more readable than (x * (y ^-1)).

mikand: I tried the followig script (note that the exponentiation operator is ** not ^ ...)
```python
import z3

a = z3.Int("a")
b = z3.Int("b")

c = a / b
d = a * (b ** -1)

f1 = (c == 2)
f2 = (d == 2)

s = z3.Solver()
s.add(f1)
res = s.check()
print("Result for %s: %s" % (f1, res))
if str(res) == "sat":
    print "  Model: " + str(s.model())

print("")
s = z3.Solver()
s.add(f2)
res = s.check()
print("Result for %s: %s" % (f2, res))
if str(res) == "sat":
    print "  Model: " + str(s.model())
```

On my machine this produces:
```
$ python test.py
Result for a/b == 2: sat
  Model: [b = 0, a = 38]

Result for a*b**-1 == 2: unknown
```

So, both seems to be representable in z3, but the second one returns unknown. Moreover, there seems to be a bug, because the model returned by the division yields an invalid model (unless division by 0 is undefined, and therefore free). I think this needs a bit more investigation...

mikand: NOTE

on the website http://rise4fun.com/z3/tutorialcontent/guide you can read the following:

" Z3 also has support for division, integer division, modulo and remainder operators. Internally, they are all mapped to multiplication. "

marcogario: Yes, this is compatible with the tests I made. So, how do you suggest that we proceed? I think that there seems to be a case to leave Div.

Re division by 0: it is undefined (see http://smtlib.cs.uiowa.edu/theories-Reals.shtml)
```Since in SMT-LIB logic all function symbols are interpreted as total
  functions, terms of the form (/ t 0) *are* meaningful in every 
  instance of Reals. However, the declaration imposes no constraints
  on their value. ```

mikand: OK to leave the DIV.

pysmt/formula.py
mikand: OK, let's keep this even if it is not uniform wrt the other operators.

pysmt/numeral.py
mikand: I agree that this should never happen with the code in this PR, so I think an exception is better to signal future issues on this matter.

marcogario:
Ok. I changed the warning to an NonImplementedError: if this should not be possible, then lets make sure it is not! :)

I renamed the test file in test_nlira, adding (a few) tests for NIA. The only thing missing is adding some files from SMT-LIB.

marcogario:
I added examples from SMT-LIB. For me it is ok to merge this. 

@mikand  Can you take a second look and eventually merge?

PR #284, opened 2016-07-24
Portfolio Solver

marcogario:
- Created Portfolio class that uses multiprocessing to solve the problem
  using multiple solvers.
- The solver provides a naive incremental interface by using
  IncrementalTrackingSolver Interface
- get_value and get_model work after a SAT query. Other
  artifacts (unsat-core, interpolants) are not supported.
- Extend Factory.is_\* to include `portfolio` key-word, and exported as
  is_\* shortcuts. The syntax becomes:
  
  ```
   is_sat(f, portfolio=["s1", "s2"])
  ```

Exceptions are currently not handled internally. Similarly, logic testing is done only when the solvers are started (i.e., during solve()).  If at least one solver terminates, this might not be a problem, but it is not nice.

### DIFF
--- a/pysmt/solvers/btor.py
+++ b/pysmt/solvers/btor.py
@@ -76,8 +76,6 @@ class BoolectorSolver(IncrementalTrackingSolver,
     LOGICS = [QF_BV, QF_UFBV, QF_ABV, QF_AUFBV, QF_AX]
     OptionsClass = BoolectorOptions
 
-    OptionsClass = BoolectorOptions
-
     def __init__(self, environment, logic, **options):
         IncrementalTrackingSolver.__init__(self,
                                            environment=environment,
@@ -89,24 +87,6 @@ class BoolectorSolver(IncrementalTrackingSolver,
         self.mgr = environment.formula_manager
         self.declarations = {}
         return
-
-    def _prepare_config(self, options):
-        options = self.options
-        if options.generate_models:
-            self.btor.Set_opt("model_gen", 1)
-        else:
-            self.btor.Set_opt("model_gen", 0)
-
-        # Disabling Incremental usage is not allowed.
-        # This needs to be set to 1
-        self.btor.Set_opt("incremental", 1)
-
-        if options.rewrite_level is not None:
-            self.btor.Set_opt("rewrite_level", options.rewrite_level)
-        if options.dual_prop is not None:
-            self.btor.Set_opt("dual_prop", options.dual_prop)
-        if options.eliminate_slices is not None:
-            self.btor.Set_opt("eliminate_slices", options.eliminate_slices)
 
 # EOC BoolectorOptions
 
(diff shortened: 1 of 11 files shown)


marcogario:
@mikand: This is not ready to be merged, but I'd like some feedback. An open point that I am not sure how to manage is multiple instances of the same solver with different options. I cannot figure out a syntax that I like.

marcogario:
Done with the exception handling (as discussed). Missing is the handling of options for solvers, including multiple instance of the same solver with different options (btw, we should find a meaningful example for this).

pysmt/solvers/msat.py
```diff
@@ -131,20 +131,29 @@ def __contains__(self, x):
         """Returns whether the model contains a value for 'x'."""
         return x in (v for v, _ in self)
 
+
+class MSatOptions(SolverOptions):
+    """Options from MathSAT Solver. """
+    VALID_OPTIONS = SolverOptions.VALID_OPTIONS +\
+                    [("random_seed", None),
```
marcogario: randomseed is part of the SMT-LIB standard, we should move this option to SolverOptions (and incidentally, consider which other options from the standard we can support at this time).

pysmt/solvers/portfolio.py
```diff
@@ -187,15 +197,15 @@ def _run_solver(solver, options, formula, signaling_queue, ctrl_pipe):
     from pysmt.environment import get_env
 
     Solver = get_env().factory.Solver
-    with Solver(name=solver) as s:
+    with Solver(name=solver, **options) as s:
```
marcogario: Since we are not using the incremental interface of the solver, we should disable it using the options  (```incremental=False```)

pysmt/solvers/btor.py
```diff
@@ -76,6 +76,8 @@ class BoolectorSolver(IncrementalTrackingSolver,
     LOGICS = [QF_BV, QF_UFBV, QF_ABV, QF_AUFBV, QF_AX]
     OptionsClass = BoolectorOptions
 
+    OptionsClass = BoolectorOptions
```
mikand: Duplicated line (see above)

pysmt/solvers/btor.py
```diff
@@ -88,6 +90,28 @@ def __init__(self, environment, logic, **options):
         self.declarations = {}
         return
 
+    def _prepare_config(self, options):
```
mikand: Where is this function used?

pysmt/solvers/portfolio.py
```diff
@@ -0,0 +1,255 @@
...
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+import logging
+from multiprocessing import Process, Queue, Pipe
+
+from pysmt.solvers.solver import IncrementalTrackingSolver, SolverOptions
```
mikand: Does this work? SolverOptions is now in ```pysmt.solvers.options``` ...

marcogario: Only if you believe with all your heart! :wink: 

pysmt.solver imports SolverOptions, therefore it is visible from there. I left it like this, in case we want to split the files in the future.

PR #286, opened 2016-07-27
Bdd simplify

marcogario:
Simplifier based on BDD conversion. Provides additional utilities like validation and handling of SMT formulas via boolean abstraction.

### DIFF
--- a/pysmt/simplifier.py
+++ b/pysmt/simplifier.py
@@ -671,6 +671,7 @@ class Simplifier(pysmt.walkers.DagWalker):
 class BddSimplifier(Simplifier):
     def __init__(self, env=None, static_ordering=None, bool_abstraction=False):
         Simplifier.__init__(self, env=env)
+        self.super_functions = dict(self.functions)
         self._validation_sname = None
 
         Solver = self.env.factory.Solver
@@ -682,8 +683,8 @@ class BddSimplifier(Simplifier):
         self.back = self.s.converter.back
         # Set methods for boolean_abstraction
         self.bool_abstraction = bool_abstraction
-        self.set_function(self.walk_abstract, *op.RELATIONS)
-        self.set_function(self.walk_abstract, op.FUNCTION)
+        self.set_function(self.walk_simplify_and_abstract, *op.RELATIONS)
+        self.set_function(self.walk_abstract_function, op.FUNCTION)
         self.ba_map = {}
         self.get_type = self.env.stc.get_type
         self.FreshSymbol = self.env.formula_manager.FreshSymbol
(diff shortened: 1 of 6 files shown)


pysmt/simplifier.py
```diff
@@ -664,3 +666,73 @@ def walk_div(self, formula, args, **kwargs):
...
+    def _validate(self, old, new):
+        if self.validate_simplifications:
+            Iff = self.env.formula_manager.Iff
+            is_valid = self.env.factory.is_valid
+            sname = self._validation_sname
+            assert is_valid(Iff(old, new), solver_name=sname ), \
+              "Was: %s \n Obtained: %s\n" % (str(old), str(new))
+
+    def abstract_and_simplify(self, formula):
+        abs_formula = self.walk(formula)
```
mikand: Would it make sense to invoke the syntactical simplification beforehand? If an atom simplifies to TRUE or FALSE the BDD could be smaller....

mikand: Even if the Simplifier recognizes that two atoms are equivalent the BDD could be reduced...

pysmt/simplifier.py
```diff
@@ -664,3 +666,73 @@ def walk_div(self, formula, args, **kwargs):
         return self.manager.Div(sl, sr)
 
 # EOC Simplifier
+
+
+class BddSimplifier(Simplifier):
```
mikand: This class will throw an exception at construction time if pyCUDD is not installed, it is OK for now, but for the next simplifier we would need a factory class...

pysmt/walkers/dag.py
```diff
@@ -70,8 +70,11 @@ def _compute_node_result(self, formula, **kwargs):
             except KeyError:
                 f = self.walk_error
 
-            args = [self.memoization[self._get_key(s, **kwargs)] \
-                    for s in self._get_children(formula)]
+            try:
+                args = [self.memoization[self._get_key(s, **kwargs)] \
+                        for s in self._get_children(formula)]
+            except KeyError as ex:
```
mikand: This should never happen if the walker is correct, right?

pysmt/simplifier.py
```diff
@@ -664,3 +666,73 @@ def walk_div(self, formula, args, **kwargs):
...
+              "Was: %s \n Obtained: %s\n" % (str(old), str(new))
+
+    def abstract_and_simplify(self, formula):
+        abs_formula = self.walk(formula)
+        abs_res = self.back(self.convert(abs_formula))
+        res = abs_res.substitute(self.ba_map)
+        return res
+
+    def walk_abstract(self, formula, args, **kwargs):
+        type_ = self.get_type(formula)
```
marcogario: This is not needed. Here it will always be BOOL

marcogario:
I addressed the comments in the PR. Ready to be reviewed again.

PR #287, opened 2016-08-01
Python 3.5

marcogario:
This PR completes the support for python 3.5. With this, we support two versions of python: 2.7 and 3.5. All solvers have been extended/adjusted in order to work with python 3.5.

### DIFF
--- a/pysmt/cmd/install.py
+++ b/pysmt/cmd/install.py
@@ -135,8 +135,9 @@ def parse_options():
                         type=str, default="~/.smt_solvers",
                         help='The folder to use for the installation')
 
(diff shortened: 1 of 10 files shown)


pysmt/cmd/install.py
```diff
@@ -135,8 +135,9 @@ def parse_options():
                         type=str, default="~/.smt_solvers",
                         help='The folder to use for the installation')
 
+    py_bindings = "~/.smt_solvers/python-bindings-%d.%d" % sys.version_info[0:2]
```
mikand: :+1:

PR #300, opened 2016-08-20
Installers: Make dependency from pip optional

marcogario:
Picosat installer only checks pip availability when trying to install it. On systems that do not have pip, it is still possible to use install.py to check if Picosat is available.

### DIFF
--- a/pysmt/cmd/install.py
+++ b/pysmt/cmd/install.py
@@ -33,7 +33,7 @@ INSTALLERS = [Installer(MSatInstaller,    "5.3.9", {}),
               Installer(CVC4Installer,    "1.5-prerelease", {"git_version" : "c15ff43597b41ea457befecb1b0e2402e28cb523"}),
               Installer(YicesInstaller,   "2.4.2", {"yicespy_version": "22b94419522ba772a1cc1e72dbe84e01b8adc16d"}),
               Installer(BtorInstaller,    "2.2.0", {"lingeling_version": "bal"}),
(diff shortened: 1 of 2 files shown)


mikand:
I like the change, but there is a more general issue here. On ubuntu 15.04+ the pip command we issue from the picosat installer is not working due to a bug (https://github.com/pypa/pip/issues/3826).

I created a patch (https://github.com/pysmt/pysmt/commit/52e35a9b0a6b846f4eef8babe357ba854f68210e) but this is ubuntu specific.

We might want to entirely avoid pip somehow. Do you have any idea on this?

marcogario:
* What happens if you always provide ```--system```? Would it be wrong?
* How does it fail? Can we catch the exception?

The idea of using the packaged version was to avoid compilation. We can either build the package locally, or download it from pypi and unpack it. 

Nevertheless, I would still prefer to use pip, if possible (notice how the installer for Picosat is much simpler than the average installer).

mikand:
always specifying --system does not work on Travis (https://travis-ci.org/pysmt/pysmt/builds/152924032) Probably we can catch the exception thrown on ubuntu without the --system and launch the ubuntu specific version. Isn't there a way to avoid compilation without using PIP?

marcogario:
I just checked: the pip installation still performs the compilation, looks like I never got around to build the binary pkgs.

Potential solution: download the pypi pkg (https://pypi.python.org/pypi/pyPicosat ) unpack it and run 
``` setup.py build``` and copy the bits and pieces (like for mathsat).

PR #302, opened 2016-08-20
Rewritings: TimesDistributor

marcogario:
Push Times operator down through Plus and Minus.

* Added utility function 'formulas_from_smtlib_test_set' to be able to
  get all SMT-LIB examples that fit in a given logic.


pysmt/rewritings.py
```diff
@@ -618,26 +618,78 @@ def walk_ite(self, formula, args, **kwargs):
         else:
             return formula
 
+
+class TimesDistributor(IdentityDagWalker):
+    """Normalize the use of multiplication by pushing it into the leafs.
+
+    E.g., (x+1)*3 -> (x*3) + 3
```
ahmedirfan1983: will this rewriting handle the case when both lhs and rhs are plus or minus nodes.

for example: 
```
(x+1) * (x -1) -> x*x -1*x + 1*x - 1*1
```

I am afraid, it will not. 

What do you think?

marcogario: It should, but you can add it as an example to be sure! :) 

It will first distribute on the left x*(x-1)...  That formula will be process immediately after (notice that get_children is called when going down the tree) and thus it will distribute on the right, since there is a symbol  on the ledt.

marcogario: Well, it did not work... but for a different reason :)
There was a typo in walk_times (I was checking the lhs instead of the rhs).

I was on mobile, now I can try to explain the passages better. ```get_children``` puts on the stack the nodes that need to be processed. This is done during the downwards visit of the formula. For the expression ```Times(a,b)``` it would normally return ```[a,b]```.
Here, we create half of the rewriting when going down, and recombine the pieces when going up (a much more complex example of this strategy is in the NNFizer).
For your example: ```(x+1) * (x -1) ``` we return the list ``` [  x*(x-1), 1*(x-1) ] ```. Immediately after this, the element ``` x*(x-1)``` is popped from the stack, and similarly we put on the stack the elements ```[x*x, x*-1]```. This is the first part of the result. You can imagine that the same occurs on the other half.

At this point, when going back up, we are visiting expressions that were not part of the original formula. In particular, ```walk_times``` will be called on ```x*(x-1)```. At this point, we know that ```args``` will contain the rewriting of  ```[x*x, x*-1]```. We look at the original formula (```x*(x-1)```) and we decide that these nodes need to be conjoined with a Minus.

I do not know if this clarifies anything... it is much easier with a whiteboard :)

marcogario:
This depends on #312

PR #303, opened 2016-08-20
Version information:

marcogario:
Make versioning information consistent within the repo.

The version is store int pysmt/__init__.py . There we define:

* VERSION     : A tuple containing the version information
* __version__ : String representation of VERSION (following PEP 440)
* git_version : A simple function that returns the version including
                git information.

pysmt.__version__ information is used by setup.py and docs/conf.py to
obtain the version information.

install.py and shell.py gain a new --version option that uses
git_version to display the version information.


mikand:
:+1:

Issue #297, opened 2016-08-16
support for ntimes in Fnode


ahmedirfan1983:
Current Fnode for times operator is a binary node, which makes rewriting an nary times expression difficult using the Identity dag walker. Such kind of rewriting would be a lot easier if pysmt has ntimes Fnode. 

for example:
sorting the following expression x*y*x (i mean the result should x*x*y) would be easier if we have ntimes Fnode.

marcogario:
@mikand In order for this to be completed we need:

- [ ]  Support for n-ary Times (printing, oracles, solvers, etc.)
- [ ]  Flattening of n-ary Times in Simplifier

I would not bother now about sorting, and handle it in a separate issue. For your convenience, I created the branch  i297/nary_times  with the simplest examples and tests possible.

mikand:
OK, I'll do this ASAP

mikand:
See PR #304

PR #304, opened 2016-08-20
N-ary multiplication

mikand:
As explained in issue #297, this PR transforms the Times operator in a n-ary node.

All the components of the system have been ported to this change.


pysmt/oracles.py
```diff
@@ -223,14 +223,12 @@ def walk_lira(self, formula, args, **kwargs):
...
         theory_out = args[0]
         for t in args[1:]:
             theory_out = theory_out.combine(t)
-        # Check for non-linear
-        left, right = formula.args()
-        if len(left.get_free_variables()) != 0 and \
-           len(right.get_free_variables()) != 0:
+        # Check for non-linear counting the arguments having at least
+        # one free variable
+        if sum(1 for x in formula.args() if len(x.get_free_variables()) != 0) > 1:
```
marcogario: ```if any(len(x.get_free_variables()) != 0):```  
(you can avoid the len by simply checking ```if x.get_free_variables()```).

pysmt/simplifier.py
```diff
@@ -288,44 +288,37 @@ def walk_plus(self, formula, args, **kwargs):
...
-                return sl
-            elif sr.is_zero():
-                if sr.is_real_constant():
-                    return self.manager.Real(0)
-                else:
-                    assert sr.is_int_constant()
-                    return self.manager.Int(0)
+                return self.manager._Algebraic(Numeral(mul))
+
+        new_args = []
```
marcogario: I think we should unify both cases. In particular, we iterate on the arguments and use an accumulator for the constants, while maintaining the non-constant objects. Currently you are doing 2 passes of the arguments. I think you can do it in one.

* If it is zero: return 0
* If it is one: continue
* If is is a constant: add to accumulator
* If is is not a constant, add to new_args

Do you have anther reason for this?

marcogario:
:+1: Ok to merge once you handle the two comments above.

Note: I added one example with 0 in the multiplication, just to be sure that it is handled correctly.

pysmt/solvers/cvc4.py
```diff
@@ -288,9 +288,16 @@ def walk_equals(self, formula, args, **kwargs):
         return self.mkExpr(CVC4.EQUAL, args[0], args[1])
 
     def walk_times(self, formula, args, **kwargs):
-        if not args[0].isConst() and not args[1].isConst():
-            raise NonLinearError(formula)
-        return self.mkExpr(CVC4.MULT, args[0], args[1])
+        res = args[0]
+        nl_count = 0 if res.isConst() else 1
```
marcogario: Looks like isConst() returns whether you have a constant term and not a constant expression.

https://travis-ci.org/pysmt/pysmt/jobs/153900213

marcogario:
This seems to fail also in the SMT-LIB wrappers. Maybe times is only binary in SMT-LIB?

pysmt/oracles.py
mikand: ```if any(len(x.get_free_variables()) != 0):```
is not correct.

Here we need to check if at least two arguments are non-constant (hence I check if the count is > 1). Surely we can avoid the len, but I feel that leaving it is clearer.

pysmt/simplifier.py
mikand: :+1:

PR #309, opened 2016-08-24
Add support for gmpy2

marcogario:
gmpy2 and Python's Fraction can now both be used in pySMT.

Setting PYSMT_GMPY env variable modifies the behavior of pysmt.constants module. In particular:

* If it is not set, the default behavior is to use GMPY2 if they are installed, or fall-back to Python's Fraction

* If set to True, GMPY2 are required and exception will be thrown if they are not available

* If set to False, Python's Fraction module will be used instead (regardless of the availability of gmpy2)

The pysmt.constants module also provides methods for testing and converting constants. This applies both for fractions (mpq and Fractions) but also for Integers (mpz, long, or int):

* is_pysmt_fraction
* is_pysmt_integer

both match the internal pysmt type.

* is_python_integer
* is_python_rational
* is_python_boolean

consider multiple ways to represent the same thing.

* pysmt_fraction_from_rational
* pysmt_integer_from_integer

convert a generic rational or integer object into a pysmt fraction or integer.

* to_python_integer: handle long/int py2/py3 mismatch

THE CHANGE IN THE NAMESPACES MAKE THIS CHANGE BACKWARDS INCOMPATIBLE. In particular:

* Methods from pysmt.utils moved to pysmt.constants
* Numerals ( moved from pysmt.numeral )


marcogario:
@mikand Not sure if this is ready to be merged, but feedback will be welcome.

.travis.yml
```diff
@@ -47,7 +48,8 @@ python:
 #
 env:
   matrix:
-    - PYSMT_SOLVER="all"
+    - PYSMT_SOLVER="all" PYSMT_GMPY="TRUE"
+    - PYSMT_SOLVER="all" PYSMT_GMPY="FALSE"
```
mikand: I I understand this right, we do not test each solver with and without gmpy, but only all the solvers with and without. Right?

Are the single-solver builds with or without gmpy?

pysmt/constants.py
```diff
@@ -0,0 +1,198 @@
...
+#
+# If set to True, Python's Fraction module will be used instead.
+#
+
+ENV_USE_GMPY = os.environ.get("PYSMT_GMPY")
+if ENV_USE_GMPY is not None:
+    ENV_USE_GMPY = ENV_USE_GMPY.lower() in ["true", "1"]
+
+HAS_GMPY = False
+try:
```
mikand: If ENV_USE_GMPY is False you can avoid the import

pysmt/constants.py
```diff
@@ -0,0 +1,198 @@
...
+    USE_GMPY = True
+else:
+    # Enable GMPY if they are available
+    USE_GMPY = HAS_GMPY
+
+#
+# Fractions using GMPY2
+#
+
+from fractions import Fraction as pyFraction
```
mikand: You can avoid this import if USE_GMPY is false

pysmt/constants.py
```diff
@@ -0,0 +1,198 @@
...
+mpq_type = None
```
mikand: You use ```mpq_type``` to check for compatibility in the ```is_python_rational function```,  but if we GMPY is not used as the backend library for pysmt, then we do not accept an mpq as input, because mpq_type is set to None!

pysmt/constants.py
```diff
@@ -0,0 +1,198 @@
...
+    """Tests whether var is a Fraction.
+
+    This takes into account the class being used to represent the Fraction.
+    """
+    return type(var) == FractionClass
+
+#
+# Integers using GMPY2
+#
+mpz_type = None
```
mikand: Same issue as for ```mpq_type```

pysmt/smtlib/parser.py
```diff
@@ -532,7 +532,7 @@ def atom(self, token, mgr):
                             if "." in token:
                                 res = mgr.Real(frac)
                             else:
-                                res = mgr.Int(frac.numerator)
+                                res = mgr.Int(int(frac.numerator)) #MG: Shall we use MPZ for ints?
```
mikand: Why not just passing ```frac.numerator```?
If we are using gmpy, it is an mpz and we avoid the useless conversion, otherwise it is a python integer and we are happy with it...

pysmt/solvers/cvc4.py
```diff
@@ -246,11 +246,13 @@ def walk_ite(self, formula, args, **kwargs):
     def walk_real_constant(self, formula, **kwargs):
         frac = formula.constant_value()
         n,d = frac.numerator, frac.denominator
-        return self.mkConst(CVC4.Rational(n, d))
+        rep = str(n) + "/" + str(d)
+        return self.mkConst(CVC4.Rational(rep))
```
mikand: :+1:  This was an unreported bug!

pysmt/solvers/cvc4.py
```diff
@@ -246,11 +246,13 @@ def walk_ite(self, formula, args, **kwargs):
...
 
     def walk_int_constant(self, formula, **kwargs):
-        assert type(formula.constant_value()) == int
-        return self.mkConst(CVC4.Rational(formula.constant_value()))
+        assert is_pysmt_integer(formula.constant_value())
+        rep = str(formula.constant_value())
+        return self.mkConst(CVC4.Rational(rep))
```
mikand: Is ```CVC4.Rational``` ok or is it a typo for ```CVC4.Integer```?

mikand:
:+1: Really nice work, apart for the few comments it is a big improvement in the cleanness of the numeric part!

.travis.yml
marcogario: Yes, that is correct.
The single solver builds are without, because we do not install gmpy. To be explicit, we could just add ```PYSMT_GMPY="FALSE"``` in all other cases.

pysmt/constants.py
marcogario: No (see comment below): we import them to check the mpq/mpz type. You are right below, these types should probably be defined here.

pysmt/constants.py
marcogario: Same as above.

pysmt/solvers/cvc4.py
marcogario: Yes. You can build an Integer object, but the ```mkConst``` is not defined for that case. I also had the same doubt and initially wrote Integer, but it was failing.

marcogario:
Ok, I addressed these points. 

Overall, I think this will make a good addition to pySMT. It is true that we increase by one the dimensions that we need to test and that in many use-cases, GMP are not needed. However, seen the issues with Fractions in python 2.7, I think that this is the only viable alternative (as long as we support 2.7).

Finally, if we support them at the pySMT level then we can, in the future, start experimenting with tighter integration with the underlying solvers (instead of using strings).

To me what is left is:

- [ ] Add description of gmpy and PYSMT_GMPY variable in the documentation
- [ ] Add some information about versioning (which is the minimal version we need?)
- [ ] Tests: Is there any relevant case that we are not testing?

.travis.yml
mikand: :+1:

pysmt/constants.py
mikand: :+1:

pysmt/constants.py
mikand: :+1:

PR #317, opened 2016-08-28
CI: Push / PR Distinction

marcogario:
* Run tests for 3.5 during regular push, except for PYSMT_SOLVER='all'
* Run tests for 3.5, 2.7 and pypy during a PR, including PYSMT_SOLVER='all'


marcogario:
We still "waste" 30 seconds for each configuration that is "disabled". I think that combining all py27 configurations into a single one would be reasonable.

Nevertheless, this should be enough for the near future.

PR #319, opened 2016-09-05
SizeOracle: Add MEASURE_BOOL_DAG measure

marcogario:
Measure the Boolean size of the formula. This is equivalent to replacing
every theory expression with a fresh boolean variable, and measuring the
DAG size of the formula.

This can be used to estimate the Boolean complexity of the SMT formula.


pysmt/oracles.py
```diff
@@ -109,6 +115,12 @@ def walk_count_symbols(self, formula, args, measure, **kwargs):
             return frozenset([formula]) | a_res
         return a_res
 
+    def walk_count_bool_dag(self, formula, args, measure, **kwargs):
+        #pylint: disable=unused-argument
+        if formula.is_bool_op():
+            return frozenset([formula]) | frozenset([x for s in args for x in s])
```
mikand: In this way, you are only counting the Boolean connectives, right?

Boolean variables, an theory atoms count as 0, is this right?

PR #321, opened 2016-09-07
Z3 back conversion

marcogario:
* Extend Z3 back conversion to allow the introduction of new symbols. 
* Fixed a bug in the back conversion of Times

Both behaviors are possible only if using Tactics from Z3. However, this is something we want to allow in the future, and therefore it is important to have it working now.

### DIFF
--- a/pysmt/solvers/z3.py
+++ b/pysmt/solvers/z3.py
@@ -302,8 +302,6 @@ class Z3Converter(Converter, DagWalker):
             z3.Z3_OP_BSUB : lambda args, expr: self.mgr.BVSub(args[0], args[1]),
             z3.Z3_OP_EXT_ROTATE_LEFT : lambda args, expr: self.mgr.BVRol(args[0], args[1].bv_unsigned_value()),
             z3.Z3_OP_EXT_ROTATE_RIGHT: lambda args, expr: self.mgr.BVRor(args[0], args[1].bv_unsigned_value()),
-
-
             z3.Z3_OP_POWER : lambda args, expr: self.mgr.Pow(args[0], args[1]),
             z3.Z3_OP_SELECT : lambda args, expr: self.mgr.Select(args[0], args[1]),
             z3.Z3_OP_STORE : lambda args, expr: self.mgr.Store(args[0], args[1], args[2]),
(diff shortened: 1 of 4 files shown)


pysmt/solvers/z3.py
```diff
@@ -418,7 +424,14 @@ def _back_single_term(self, expr, args, model=None):
                 return self.mgr._Algebraic(Numeral(expr))
             else:
                 # it must be a symbol
-                res = self.mgr.get_symbol(str(expr))
+                try:
+                    res = self.mgr.get_symbol(str(expr))
+                except UndefinedSymbolError:
+                    import warnings
+                    symb_type = self._z3_to_type(expr.sort())
+                    res = self.mgr.Symbol("__z3_%s" % str(expr), symb_type)
```
mikand: Using FreshSymbol instead of symbol would be much better, because it GUARANTEES no name clashing

marcogario: Ok.

marcogario:
The last commit does not cover all cases. I need some time to uniform all the walk_* methods.
Currently we get 2x faster back conversion (via API) and 4x faster forward conversion (by skipping several intermediate checks that are not needed in our case).

Note: We could do much better on the Z3 back cpnversion, if we were to get rid of the intermediate AstRef objects from the Z3 python API. We actually do not need them, and they are useful only if one wants to use Z3 Python expressions directly. However, the AstRef class is responsible to manage the ref-counter of the context. It is possible to build a context that is memory managed by using mk_context (currently the library uses mk_context_rc). The only problem here is that if we want to manipulate Z3 Python Expressions directly, we need somehow to obtain an AstRef that "plays nice" with the context.

marcogario:
@ahmedirfan1983, please note that I reverted the signature of Z3Converter.back. To use the SMT-LIB back conversion, please use Z3Converter.back_via_smtlib() . This makes the interface of the Converter slightly more consistent.

pysmt/test/test_back.py
```diff
@@ -57,13 +57,20 @@ def test_msat_back_not_identical(self):
...
-    def do_back(self, solver_name):
+    def do_back(self, solver_name, z3_string_buffer=False):
         for formula, _, _, logic in get_example_formulae():
             if logic.quantifier_free:
                 try:
                     s = Solver(name=solver_name, logic=logic)
                     term = s.converter.convert(formula)
-                    res = s.converter.back(term)
+                    if solver_name == "z3" and z3_string_buffer:
+                        if logic.theory.bit_vectors: continue
```
marcogario: This should work with bitvectors, but is pending #323 .

Issue #313, opened 2016-08-27
Tests: Move serialization of examples within the examples


marcogario:
The HR Serialization of the examples (from tests/examples.py) is stored in the file ```test_printing.py```. This means that everytime we add a new example, we need to modify two files. What often happens is that we forget to modify ```test_printing.py``` and tests fail during CI.

The proposal is to extend the ```Example``` class in order to include the HR String. This would work both as a documentation for the example itself (currently done with a comment containing something close to the HR representation), and would be used to test the HR printer.

This will force us to always provide the String representation.

mikand:
Closed in #325

PR #325, opened 2016-09-19
Extend Example to include serialized formula

marcogario:
* Example has a new field 'hr', to store the HR serialization of the
  formula;

* Introduced a simplified SExample namedtuple for backwards
  compatibility;

* Three methods can be used to iterate over the examples:
 * get_example_formula : Returns all SExamples
 * get_str_example_formula: Returns all String, Expr and Logic
 * get_full_example_formula: Returns all Examples


marcogario:
Fix #313

Issue #323, opened 2016-09-08
Should bv1 variable be declarable in SMT-LIB file?


marcogario:
This breaks the parser. The problem is that bv1 is a keyword in ```(_ bv1 8)```, but a variable otherwise.
Our parser breaks with this example. z3 seems to be able to parse it correctly.

Not sure which behavior is the right one.

```
(set-info :status unknown)
(declare-fun bv1 () (_ BitVec 8))
(assert
 (bvult (_ bv0 8) (bvudiv (bvmul (bvadd bv1 (_ bv1 8)) (_ bv5 8)) (_ bv5 8))))
(check-sat)
```

marcogario:
Fixed in #326

PR #326, opened 2016-09-20
I323/parsing bv x variables

mikand:
Fixed issue #323 by allowing bvX variables in the parser.

### DIFF
--- a/pysmt/test/test_regressions.py
+++ b/pysmt/test/test_regressions.py
@@ -18,9 +18,12 @@
 from six.moves import xrange
 from six.moves import cStringIO
 
+import pysmt.logics as logics
+import pysmt.smtlib.commands as smtcmd
 from pysmt.shortcuts import (Real, Plus, Symbol, Equals, And, Bool, Or,
                              Div, LT, LE, Int, ToReal, Iff, Exists, Times, FALSE,
-                             BVLShr, BVLShl, BVAShr, BV, BVAdd, Select)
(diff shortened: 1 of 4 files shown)


marcogario:
Memo: Tests on z3 back conversion using smtlib need to be reenabled for bv (didn't check yet)

mikand:
@marcogario The back conversion on bv has been re-enabled (see https://github.com/pysmt/pysmt/pull/326/commits/223ea43cd316a51f28edf0feb840ab44475f5d23) and it is the issue breaking the tests...

marcogario:
Ok, assign it to me if you want me to look into this.

pysmt/test/test_regressions.py
```diff
@@ -289,6 +289,16 @@ def test_parse_define_fun_bind(self):
...
 
+    def test_parse_bvx_var(self):
+        smtlib_input = """
+        (declare-fun bv1 () (_ BitVec 8))
+        (assert (bvult (_ bv0 8) (bvudiv (bvmul (bvadd bv1 (_ bv1 8)) (_ bv5 8)) (_ bv5 8))))
+        (check-sat)"""
+        parser = SmtLibParser()
+        buffer_ = cStringIO(smtlib_input)
+        parser.get_script(buffer_)
+
```
marcogario: This is only testing that the code does not crash. We can do much better, and actually test that we obtain the expression defined above.

PR #331, opened 2016-09-29
MSat: Check if solver is configured for unsat-core

marcogario:
If the solver was not configured for unsat core, a call to get_unsat_core would lead to an infinite recursion between get_unsat_core and get_named_unsat_core.

This is fixed by checking the configuration of the solver as first thing in those functions.


marcogario:
@ahmedirfan1983, FYI

PR #336, opened 2016-10-03
CVC4: Use on python-config in ./configure

marcogario:
Call CVC4 configure using the default path of the 'python' executable 
to compile the python bindings.

On systems where multiple versions of python are installed, running
install.py with a different version of python (e.g., python3.5) was
leading to the creation of invalid bindings.


marcogario:
Integrated within PR #338 .

PR #338, opened 2016-10-04
Use solver_options to specify solver-dependent options

marcogario:
Options that are shared by multiple solvers are handled as keywords
arguments, but solver-dependent options are defined in the
solver_options dictionary:

  Solver(name='picosat', generate_models=True,
         solver_options={'propagation_limit': 100})

The goal is to clearly separate these two type of configurations.
The use of dictionary for solver_options, makes it simple to store
multiple configurations:

  Solver(name=sname, solver_options=myoptions[sname])

All solvers have been ported to this new strategy, by using a custom
SolverOptions object that is responsible for the definition, checking
and setting of the options. This object is used (for example) to
correctly setup the mathsat context.

Option "random_seed" is now supported by all the solvers (that do
support it).

Backwards Incompatible Changes:

* MathSAT5Solver option 'debugFile' has been removed. Use the
  solver option: "debug_api_call_trace_filename".

* BddSolver used to have the options as keyword
  arguments (static_ordering, dynamic_reordering etc). This is not
  supported anymore.

### DIFF
--- a/pysmt/smtlib/solver.py
+++ b/pysmt/smtlib/solver.py
@@ -22,9 +22,46 @@ import pysmt.smtlib.commands as smtcmd
 from pysmt.solvers.eager import EagerModel
 from pysmt.smtlib.parser import SmtLibParser
 from pysmt.smtlib.script import SmtLibCommand
-from pysmt.solvers.solver import Solver
+from pysmt.solvers.solver import Solver, SolverOptions
 from pysmt.exceptions import (SolverReturnedUnknownResultError,
                               UnknownSolverAnswerError)
+
+
+class SmtLibOptions(SolverOptions):
+    """Options for the SmtLib Solver.
+
+    * debug_interaction: True, False
+      Print the communication between pySMT and the wrapped executable
+    """
+    def __init__(self, **base_options):
+        SolverOptions.__init__(self, **base_options)
+        if self.unsat_cores_mode is not None:
+            raise ValueError("'unsat_cores_mode' option not supported.")
+        self.debug_interaction = False
+
+        if 'debug_interaction' in self.solver_options:
+            self.debug_interaction = self.solver_options
+
+    def __call__(self, solver):
+        # These options are needed for the wrapper to work
+        solver.set_option(":print-success", "true")
+        solver.set_option(":diagnostic-output-channel", '"stdout"')
+
+        if self.generate_models:
+            solver.set_option(":produce-models", "true")
+        else:
+            solver.set_option(":produce-models", "false")
+
+        if self.random_seed is not None:
+            solver.set_option(":random-seed", str(self.random_seed))
+
+        for k,v in self.solver_options.items():
+            if k in (':print-success', 'diagnostic-output-channel'):
+                raise ValueError("Cannot override %s." % k)
+            solver.set_option(k, str(v))
+
+# EOC SmtLibOptions
+
 
 class SmtLibSolver(Solver):
     """Wrapper for using a solver via textual SMT-LIB interface.
@@ -33,13 +70,13 @@ class SmtLibSolver(Solver):
     the executable. Interaction with the solver occurs via pipe.
     """
 
+    OptionsClass = SmtLibOptions
+
     def __init__(self, args, environment, logic, LOGICS=None, **options):
         Solver.__init__(self,
                         environment,
-                        logic=logic)
-
-        # Flag used to debug interaction with the solver
-        self.dbg = False
+                        logic=logic,
+                        **options)
 
         if LOGICS is not None: self.LOGICS = LOGICS
         self.args = args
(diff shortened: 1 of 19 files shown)


marcogario:
This PR includes two additional changes:

* CVC4 installer is more robust when installing on a system with multiple version of python
* pyPicosat has been extended to better deal with flushing of the output log.

pysmt/smtlib/solver.py
```diff
@@ -22,24 +22,61 @@
...
+        if self.generate_models:
+            solver.set_option(":produce-models", "true")
+        else:
+            solver.set_option(":produce-models", "false")
+
+        if self.random_seed is not None:
+            solver.set_option(":random-seed", str(self.random_seed))
+
+        for k,v in self.solver_options.items():
+            if k in (':print-success', 'diagnostic-output-channel'):
```
mikand: What if the user pusts ```:produce-models``` in the solver_options? Maybe we should avoid overriding all the pysmt-interpreted options.

pysmt/smtlib/solver.py
```diff
@@ -86,13 +123,13 @@ def _send_silent_command(self, cmd):
     def _get_answer(self):
         """Reads a line from STDOUT pipe"""
         res = self.solver_stdout.readline().strip()
-        if self.dbg: print("Read: " + str(res))
+        if self.options.debug_interaction: print("Read: " + str(res))
```
mikand: You could use ```_debug```

pysmt/smtlib/solver.py
```diff
@@ -71,9 +104,13 @@ def set_option(self, name, value):
     def set_logic(self, logic):
         self._send_silent_command(SmtLibCommand(smtcmd.SET_LOGIC, [logic]))
 
+    def _debug(self, msg):
```
mikand: To save a little bit of time you could change ```_debug``` as follows.

```python
def _debug(self, msg, *format_args):
    if self.options.debug_interaction:
        print(msg % format_args)
```

In this way, we can avoid the construction of the debug string if the debugging is not enabled.

pysmt/smtlib/solver.py
```diff
@@ -86,13 +123,13 @@ def _send_silent_command(self, cmd):
...
         res = self.solver_stdout.readline().strip()
-        if self.dbg: print("Read: " + str(res))
+        if self.options.debug_interaction: print("Read: " + str(res))
         return res
 
     def _get_value_answer(self):
         """Reads and parses an assignment from the STDOUT pipe"""
         lst = self.parser.get_assignment_list(self.solver_stdout)
-        if self.dbg: print("Read: " + str(lst))
+        if self.options.debug_interaction: print("Read: " + str(lst))
```
mikand: You should use ```_debug```

pysmt/solvers/cvc4.py
```diff
@@ -38,9 +38,41 @@
 from pysmt.constants import Fraction, is_pysmt_integer, to_python_integer
 
 
+class CVC4Options(SolverOptions):
+
+    def __init__(self, **base_options):
+        SolverOptions.__init__(self, **base_options)
+        # TODO: CVC4 Supports UnsatCore extraction
```
mikand: Add reference to #349

pysmt/solvers/cvc4.py
```diff
@@ -98,7 +125,7 @@ def solve(self, assumptions=None):
         else:
             try:
                 res = self.cvc4.checkSat()
-            except CVC4.LogicException as ex:
+            except:
                 raise InternalSolverError(ex.toString())
```
mikand: ```ex``` variable is now unassigned!

pysmt/solvers/yices.py
```diff
@@ -61,10 +61,68 @@ def cleanup():
 STATUS_UNSAT = 4
 
 
+class YicesOptions(SolverOptions):
+    def __init__(self, **base_options):
+        SolverOptions.__init__(self, **base_options)
+        # TODO: Yices Supports UnsatCore extraction
```
mikand: AFAIK Yices does NOT support Unsat Core extraction

pysmt/solvers/yices.py
```diff
@@ -73,18 +131,29 @@ def __init__(self, environment, logic, **options):
...
+        self.yices = None
+        self.yices_params = None
+        self._create_yices_context()
         self.converter = YicesConverter(environment)
         self.mgr = environment.formula_manager
         self.model = None
         self.failed_pushes = 0
         return
 
+    def _create_yices_context(self):
```
mikand: Why is this in a separate method? All the other solvers have this in their constructor, and this method is only called from there....

pysmt/test/test_rewritings.py
```diff
@@ -15,6 +15,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 #
+from nose.plugins.attrib import attr
```
mikand: Is this used?

pysmt/solvers/yices.py
marcogario: Mostly because there is a bit of additional code to handle 1) the logic selection 2) setting of parameters (set_params) 3) freeing of configuration. 

We can also probably just isolate the bit about the logic conversion.

pysmt/smtlib/solver.py
marcogario: Not sure if I understand this point.

pysmt/solvers/yices.py
marcogario: You are right. The manual explicitly says so.

PR #341, opened 2016-10-05
Fix array values map bug

mikand:
This PR is the extension of #340.

There is a bug in  ```FNode.array_value_assigned_values_map``` that returns the wromg map of initial values for an array. This PR fixes it.

Thanks to @daniel-rs for noticing the problem

### DIFF
--- a/pysmt/factory.py
+++ b/pysmt/factory.py
@@ -511,7 +511,7 @@ class Factory(object):
                         if solver.get_value(a).is_true():
                             res.append(a)
                         else:
-                            assert solver.get_value(a).is_false(), solver_name
+                            assert solver.get_value(a).is_false()
                             res.append(mgr.Not(a))
                 return mgr.And(res)
 
--- a/pysmt/factory.py
+++ b/pysmt/factory.py
@@ -511,6 +511,7 @@ class Factory(object):
                         if solver.get_value(a).is_true():
                             res.append(a)
                         else:
+                            assert solver.get_value(a).is_false(), solver_name
                             res.append(mgr.Not(a))
                 return mgr.And(res)
 
--- a/pysmt/fnode.py
+++ b/pysmt/fnode.py
@@ -592,9 +592,9 @@ class FNode(object):
         assert index.is_constant()
         args = self.args()
         start = 0
-        end = (len(args) - 1) / 2
+        end = (len(args) - 1) // 2
         while end - start > 0:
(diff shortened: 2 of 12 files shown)


pysmt/factory.py
```diff
@@ -511,6 +511,7 @@ def get_implicant(self, formula, solver_name=None,
                         if solver.get_value(a).is_true():
                             res.append(a)
                         else:
+                            assert solver.get_value(a).is_false(), solver_name
```
marcogario: The assertion should probably raise the result of get_value, and not the solver name.

pysmt/fnode.py
```diff
@@ -576,27 +576,40 @@ def array_value_index_type(self):
         return self._content.payload
 
     def array_value_get(self, index):
+        """Returns the value of this Array Value at the given index. The
+        index must be a constant of the correct type.
+
+        This function is equivalent (but possibly faster) than teh
```
marcogario: teh

pysmt/fnode.py
```diff
@@ -576,27 +576,40 @@ def array_value_index_type(self):
...
             else:
-                e = p
+                start = pivot + 1
         return self.array_value_default()
 
+
     def array_value_assigned_values_map(self):
         res = {}
         args = self.args()
         for i,c in enumerate(args[1::2]):
```
marcogario: So, this is basically building the pairs, right?
```
dict(zip(args[1::2], args[2::2]))
```

pysmt/test/examples.py
```diff
@@ -632,6 +632,13 @@ def get_full_example_formulae(environment=None):
                     logic=pysmt.logics.get_logic_by_name("QF_ALIA*")
                 ),
 
+            Example(hr="((Array{Int, Int}(0)[1 := 3] = aii) & (aii[1] = 3))",
+                    expr=And(Equals(Array(INT, Int(0), {Int(1) : Int(3)}), aii), Equals(Select(aii, Int(1)), Int(3))),
+                    is_valid=False,
+                    is_sat=True,
+                    logic=pysmt.logics.get_logic_by_name("QF_ALIA*")
+                ),
+
```
marcogario: More tests are needed. In particular, cases in which : 
* Array Value does not have additional values
* Array Value has an even number of additional values (n > 3)
* Array Value has an odd number of additional Values (n > 3)

This are not necessarily Example's, but we can make them for regression.

PR #352, opened 2016-10-10
Created Apache Image and link to Apache 2.0

zenbhang:
Adds aesthetic improvement to the license portion of the readme.
Also provides a quick way to learn more about apache licensing, with a direct link to the site.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -198,6 +198,9 @@ The following table summarizes the features supported via pySMT for each of the 
 
 License
 =======
+.. image:: https://www.apache.org/img/asf_logo.png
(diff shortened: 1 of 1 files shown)


marcogario:
Hi! Thanks for taking the time of working on pySMT.

The licensing information is provided at the top of the README, with a link to the license hosted in the repo (that is the one that counts). Moreover, the Logo you suggest to use is quite big, and it might lead to think that we are under the umbrella of the APACHE Software fundation (but we are not).
For these reasons, I think I will not accept the PR.

@mikand, if you want to weight in, feel free to reopen this.

PR #353, opened 2016-10-10
Tests for FNode

marcogario:
This increases the coverage of tests for FNode.

**Note**: This also changes the implementation of FNode.array_value_get, since the current implementation was incorrect.

### DIFF
--- a/pysmt/test/test_array.py
+++ b/pysmt/test/test_array.py
@@ -103,6 +103,22 @@ class TestArray(TestCase):
         select_ = Select(store_, Int(100))
         self.assertTrue(store_.is_array_op())
         self.assertTrue(select_.is_array_op())
+        self.assertTrue(store_.is_store())
+        self.assertTrue(select_.is_select())
+        self.assertFalse(select_.is_store())
+        self.assertFalse(store_.is_select())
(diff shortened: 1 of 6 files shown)


marcogario:
This depends on PR #351

pysmt/test/test_array.py
```diff
@@ -103,6 +103,37 @@ def test_is_array_op(self):
...
+        ax = Array(REAL, Real(0),
+                   {Real(1): Real(2),
+                    Real(2): Real(3),
+                    Real(3): Real(4),
+                    Real(4): Real(5),
+                })
+        self.assertEqual(ax.array_value_get(Real(1)), Real(2))
+        self.assertEqual(ax.array_value_get(Real(2)), Real(3))
+        self.assertEqual(ax.array_value_get(Real(3)), Real(4))
+        self.assertEqual(ax.array_value_get(Real(4)), Real(5))
```
mikand: You can also add few more cases:

```python
self.assertEqual(ax.array_value_get(Real(-1)), Real(0))
self.assertEqual(ax.array_value_get(Real(5)), Real(0))
```

PR #359, opened 2016-10-11
Walker Refactoring

marcogario:
*\* This is outdated: See discussion below **

Recover the behavior of the super Walker when inheriting, and easily
compose it.

  self.save_super_functions()

must be called during `__init__` **before** overriding the
functions (with set_function). Inside any walk, it is then possible to
call self._super() to execute the original walk, e.g.:

  self._super(formula, args, **kwargs)

in case of a DagWalker.

Examples usage using both a DagWalker (BddSimplifier) and
TreeWalker (SmartPrinter) are given.

### DIFF
--- a/pysmt/oracles.py
+++ b/pysmt/oracles.py
@@ -32,7 +32,7 @@ from pysmt import typing as types
 from pysmt.logics import Logic, Theory, get_closer_pysmt_logic
 
 
-class SizeOracle(pysmt.walkers.DagWalker):
+class SizeOracle(walkers.DagWalker):
     """Evaluates the size of a formula"""
 
     # Counting type can be:
@@ -50,7 +50,7 @@ class SizeOracle(pysmt.walkers.DagWalker):
      MEASURE_BOOL_DAG) = range(6)
 
     def __init__(self, env=None):
-        pysmt.walkers.DagWalker.__init__(self, env=env)
+        walkers.DagWalker.__init__(self, env=env)
 
         self.measure_to_fun = \
                         {SizeOracle.MEASURE_TREE_NODES: self.walk_count_tree,
@@ -122,42 +122,28 @@ class SizeOracle(pysmt.walkers.DagWalker):
         return frozenset([formula]) | frozenset([x for s in args for x in s])
 
 
-class QuantifierOracle(pysmt.walkers.DagWalker):
-    def __init__(self, env=None):
-        pysmt.walkers.DagWalker.__init__(self, env=env)
-
-        # Propagate truth value, and force False when a Quantifier
-        # is found.
-        self.set_function(self.walk_all, *op.ALL_TYPES)
-        self.set_function(self.walk_false, op.FORALL, op.EXISTS)
-
+class QuantifierOracle(walkers.DagWalker):
     def is_qf(self, formula):
         """ Returns whether formula is Quantifier Free. """
         return self.walk(formula)
 
+    # Propagate truth value, but force False if a Quantifier is found.
+    @walkers.handles(set(op.ALL_TYPES) - op.QUANTIFIERS)
+    def walk_all(self, formula, args, **kwargs):
+        return walkers.DagWalker.walk_all(self, formula, args, **kwargs)
+
+    @walkers.handles(op.QUANTIFIERS)
+    def walk_false(self, formula, args, **kwargs):
+        return walkers.DagWalker.walk_false(self, formula, args, **kwargs)
 
 # EOC QuantifierOracle
 
 
-class TheoryOracle(pysmt.walkers.DagWalker):
-    def __init__(self, env=None):
-        pysmt.walkers.DagWalker.__init__(self, env=env)
+class TheoryOracle(walkers.DagWalker):
 
-        self.set_function(self.walk_combine, op.AND, op.OR, op.NOT, op.IMPLIES,
-                          op.IFF, op.LE, op.LT, op.FORALL, op.EXISTS, op.MINUS,
-                          op.ITE, op.ARRAY_SELECT, op.ARRAY_STORE)
-        # Just propagate BV
-        self.set_function(self.walk_combine, *op.BV_OPERATORS)
-
-        self.set_function(self.walk_constant, op.REAL_CONSTANT, op.BOOL_CONSTANT,
-                          op.INT_CONSTANT, op.BV_CONSTANT)
-        self.set_function(self.walk_symbol, op.SYMBOL)
-        self.set_function(self.walk_function, op.FUNCTION)
-        self.set_function(self.walk_lira, op.TOREAL)
-        self.set_function(self.walk_times, op.TIMES)
-        self.set_function(self.walk_plus, op.PLUS)
-        self.set_function(self.walk_equals, op.EQUALS)
-        self.set_function(self.walk_array_value, op.ARRAY_VALUE)
+    def get_theory(self, formula):
+        """Returns the thoery for the formula."""
+        return self.walk(formula)
 
     def _theory_from_type(self, ty):
         theory = None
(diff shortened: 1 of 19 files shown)


marcogario:
@agriggio is this closer to what you had in mind?

agriggio:
Yes, but why should the user need to call `save_super_functions()` explicitly?
Can't this be done automatically? Also, does this work on multiple levels?
E.g. what happens if I inherit from a `Walker` that already uses this "trick"
internally?

marcogario:
Tests are missing.

marcogario:
This is the new proposal for class-based Walkers. Walkers are now pure classes, and the mapping between nodetype and function is done at the class level (and not at the instance level as before).
This means that accessin the implementation of a walk_* of a parent class is done as usual in python: ```Class.method(self, ...)``` . One of the key differences is that now a given nodetype is always handled by the walk_ function with a given name (e.g., op.AND is always handled by walk_and).

Some utility methods have been added to handle this process (names to be revised):

* Walker.set_ : Sets the same method for a list of nodetypes (this is the equivalent of Walkers.set_functions at the class level)

* Walker._super : calls the correct walk_* of the given class for the formula in input. This is useful in the case you want to call the method of a super class but you do not know which nodetype you are currently handling.

The BddSimplifier is a good example of usage of these functions. See also ```test_walker_super```. 

Additionally, this branch shows how to make the dispatch based on nodetype much more powerful, by dynamically associating a walk_ method to each node type. This is done in different steps:

1. The name of the walk_ function is generated from the name of the nodetype (e.g., op.AND becomes walk_and)
2. To avoid a dynamic lookup every time (```getattr```), the association nodetype-function is computed during the ```__init___``` of the walker
3. Additionally, during the ```__init__``` we are in a position to handle "custom" node types, i.e., node introduced by external tools (e.g., LTL nodes). See ```test_walker_new_operators_complete```

This makes the DynamicWalkerFunction hack obsolete.

Finally, this implementation does not break existing instance-based (i.e., old) Walkers, and provides a gradual upgrade path.

Open Points:
* There are now too many ways to define a Walker method. See HRPrinter for an example. We should define a best-practice
* Dynamic handling of new nodes needs discussion. We should publish a more complex LTL example
* Names are always hard! :)
* Due to the size of the change, it might be a good moment to list additional features/desiderata for the walkers

mikand:
Crazy idea that just came to my mind: would it be possible to use decorators instead of calling the ```Walker.set_``` function at the end of the class? Would there be a performance penalty?

marcogario:
0afceec shows how to do this with a decorator. Syntactically, it looks nicer, although the machinery behind is a bit more complex, it does not have a performance impact, since this is done once during class definition. 

See the BddSimplifier and the Walker for examples of usage.
(This is intended as a demo: Names and where to put the decorators is to be decided).

pysmt/walkers/generic.py
```diff
@@ -16,12 +16,39 @@
...
+
+class nodetypes(object):
+    def __init__(self, nodetypes):
+        self.nodetypes = nodetypes
+
+    def __call__(self, func):
+        func.nodetypes = self.nodetypes
+        return func
+
+class MetaNodeTypeHandler(type):
```
mikand: Why is the metaclass needed? Can't we simply call the ```set_``` method in the decorator?

marcogario: The current form does not allow chaining of the decorator. We need to test if the attribute is set and extend it if necessary.

marcogario:
@mikand  Can you revise the design of the Walker refactoring? If you think this makes sense, I would like to merge it in the next couple of weeks.

pysmt/oracles.py
```diff
@@ -122,42 +122,28 @@ def walk_count_bool_dag(self, formula, args, measure, **kwargs):
...
-        self.set_function(self.walk_all, *op.ALL_TYPES)
-        self.set_function(self.walk_false, op.FORALL, op.EXISTS)
-
+class QuantifierOracle(walkers.DagWalker):
     def is_qf(self, formula):
         """ Returns whether formula is Quantifier Free. """
         return self.walk(formula)
 
+    # Propagate truth value, but force False if a Quantifier is found.
+    @walkers.handles(set(op.ALL_TYPES) - op.QUANTIFIERS)
```
mikand: :heart: this!

pysmt/oracles.py
```diff
@@ -221,10 +212,10 @@ def walk_function(self, formula, args, **kwargs):
                 theory_out = theory_out.combine(t)
         else:
             theory_out = Theory()
-
         theory_out.uninterpreted = True
         return theory_out
 
+    @walkers.handles(op.TOREAL)
     def walk_lira(self, formula, args, **kwargs):
```
mikand: Can't we simply call this function `walk_toreal` without decorator?

pysmt/oracles.py
```diff
@@ -420,10 +386,12 @@ def walk_ite(self, formula, args, **kwargs):
         else:
             return frozenset(x for a in args for x in a)
 
+# EOC AtomsOracle
 
 
 def get_logic(formula, env=None):
     if env is None:
+        import pysmt.environment
```
mikand: Why this import here instead of importing pysmt at the top of the file?

pysmt/rewritings.py
```diff
@@ -97,6 +81,7 @@ def convert_as_formula(self, formula):
             conj.append(self.mgr.Or(clause))
         return self.mgr.And(conj)
 
+    @deprecated("No Alternative")
```
mikand: Right, this is a debug leftover. Nobody should use this.

pysmt/substituter.py
```diff
@@ -20,11 +20,12 @@
 from six import iteritems
 
 import pysmt.walkers
+from pysmt.walkers.generic import handles
 import pysmt.operators as op
 from pysmt.exceptions import PysmtTypeError
 
 
-class Substituter(pysmt.walkers.DagWalker):
+class Substituter(pysmt.walkers.IdentityDagWalker):
```
mikand: Right now (and possibly even in master) a user can instantiate a Substituter class and optains an identity walker. Can we prevent this by throwing a `NotImplementedError` in the walk?
What do you think?

pysmt/walkers/__init__.py
```diff
@@ -33,7 +33,7 @@
 
 """
 
-__all__ = ["DagWalker", "TreeWalker", "IdentityDagWalker"]
+__all__ = ["DagWalker", "TreeWalker", "IdentityDagWalker", "handles"]
```
mikand: Is this needed? All these names are defined in this `__init__.py` so they should be already importable with `from ... import *` (https://docs.python.org/2/tutorial/modules.html#importing-from-a-package)

pysmt/walkers/generic.py
```diff
@@ -30,80 +74,44 @@ def __init__(self, env=None):
...
-        undefined_types = set(op.ALL_TYPES) - set(self.functions.keys())
-        assert len(undefined_types) == 0, \
-            "The following types are not defined in the generic walker: {%s}" % \
-            (", ".join(op.op_to_str(u) for u in undefined_types))
-
+        for o in op.all_types():
+            try:
+                # getattr will raise an AttributeError exception if a
+                # method does not exist
+                self.functions[o] = getattr(self, nt_to_fun(o))
```
mikand: So, we are still storing the nodetype-to-function information in an instance member dictionary? Ideally this dictionary should be identical among all the instances, so (apart for backward compatibility) is there a reason not for not putting this at class level?

pysmt/walkers/identitydag.py
```diff
@@ -73,11 +73,13 @@ def walk_lt(self, formula, args, **kwargs):
         return self.mgr.LT(args[0], args[1])
 
     def walk_forall(self, formula, args, **kwargs):
-        qvars = [self.walk_symbol(v) for v in formula.quantifier_vars()]
```
mikand: Why this change (and the ones below)? Doesn't this prevent proper sub-classing of IdentityDagWalker?
For example, if I define my own walk_symbol (that for example rewrites all the integer variables as real variables) this code would create the wrong formula unless I explicitly override `walk_forall`...

pysmt/walkers/__init__.py
marcogario: I think this is a left-over of when all walkers were defined in a single file. Ok with removing it.

pysmt/walkers/generic.py
marcogario: The ``functions`` dictionary is there **only** for backwards compatibility, for this reason it might not be identical (due to the nature of the instance-based walkers).
 
If you use it (via ``set_function``), you will get a deprecation warning. In a future version (0.8.0?) we will remove it, and the class-based walker will be the only way to define a walker.

pysmt/walkers/identitydag.py
marcogario: I think this is a matter of preference. I am preparing a commit with the opposite approach.

marcogario: See how substituter changed in the last commit. I am ok with either solution.

pysmt/substituter.py
marcogario: See proposed solution in the last commit.

pysmt/walkers/identitydag.py
mikand: I prefer the solution you implemented in the last commit, it is more in the spirit of general-purpose extensibility.

pysmt/walkers/generic.py
mikand: :+1:

pysmt/substituter.py
mikand: I like the solution in the last commit :+1:

mikand:
@marcogario This looks perfect to me now. You can go ahead and merge if it is OK also for you (since the issue is not assigned to me, I do not know if you wanted to do something else)

PR #360, opened 2016-10-12
Allow the user to specify a custom logic for smtlibscript_from_formula()

agriggio:
This is useful for generating scripts for solvers that only work for specific logics.


marcogario:
I added some tests, and extended the logic parameter to accept a string. This should give you more flexibility, and allow you to use logics that cannot be represented in pySMT yet (Floating Points, Strings etc.).

PR #368, opened 2016-10-15
Enforce a single call to is_sat if not incremental

marcogario:
Non incremental is_sat modifies the stack of assertions. For this
reason, it should be possible to call it only once.

This achieves this by raising an error in solve() if called after a call
to is_sat.

This also clarifies the documentation of Solver.get_model.

Thanks to @colinmorris for the feedback on Solver interface.

### DIFF
--- a/pysmt/test/test_solving.py
+++ b/pysmt/test/test_solving.py
@@ -563,6 +563,14 @@ class TestBasic(TestCase):
                 s.add_assertion(And(x,y))
                 s.solve()
 
+    @skipIfNoSolverForLogic(QF_BOOL)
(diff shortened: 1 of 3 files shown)


pysmt/test/test_solving.py
```diff
@@ -563,6 +563,14 @@ def test_picosat_options(self):
                 s.add_assertion(And(x,y))
                 s.solve()
 
+    @skipIfNoSolverForLogic(QF_BOOL)
+    def test_incrementali_is_sat(self):
```
mikand: Typo

PR #374, opened 2016-10-18
Fixed annotations parsing according to SMTLib rules 

### LINKED ISSUE #370: SMT-LIB: Interpreting Symbols in annotations leads to unexpected results
The branch ```annotations_parsing``` shows that annotations are interpreted. If an annotation contains the name of a symbol, it is treated as a symbol, otherwise as a string. The test in the branch shows this behavior. 

The main issue is that the symbol is being escaped when treated as an annotation, therefore the name of the symbol contains spurious ```\```. The expected behavior is that ```get_symbol``` at the end  of the test succeeds.

mikand:
The parsing of SMTLib2 annotations was incorrect because it interpreted annotation values as expressions. This PR resolves this issue by properly reading the annotation values as uninterpreted strings.

This PR fixes #370

### DIFF
--- a/pysmt/smtlib/parser.py
+++ b/pysmt/smtlib/parser.py
@@ -43,6 +43,7 @@ def open_(fname):
             return bz2.open(fname, "rt")
     return open(fname)
 
+
 def get_formula(script_stream, environment=None):
     """
     Returns the formula asserted at the end of the given script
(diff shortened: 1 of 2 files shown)


pysmt/smtlib/parser.py
```diff
@@ -222,6 +218,35 @@ def tokenizer(handle, interactive=False):
...
+        self.generator = tokenizer_generator(self.reader)
+        self.extra_queue = []
+        self.consume = self.consume_token
+
+    def add_extra_token(self, token):
+        self.extra_queue.append(token)
+        self.consume = self.consume_token_queue
+
+    def consume_token_queue(self):
+        if len(self.extra_queue):
```
marcogario: This can be simplified to ```if self.extra_queue```

marcogario:
I refactored the function tokenizer_generator, so that it lives within the Tokenizer. I also added the documentation to the class. If you are ok with this, go ahead and merge.

PR #375, opened 2016-10-19
Handling of Sorts

marcogario:
Introduce the support for custom sorts (PySMTTypes).

Two new classes are introduced: _Type and PartialType
- PartialType is basically a partial function, that is used to represent
  the concept of define-sort. Binding to the actual sort is post-poned,
  by using a lambda expression, internally.
- _TypeDecl: This class represents a Type declaration, and as such
  cannot be used directly to instantiate a Symbol. This capture the
  semantics of declare-sort. A wrapper Type() is given to simplify its
  use, and making 0-arity sorts a special case.

Examples:

The following three statements are equivalent:
- Type("Colors")
- Type("Colors", 0)
- _TypeDecl("Colors", 0)()

In particular, the last case shows that instantiating a TypeDeclaration
returns a PySMTType:

  Type(Pair, 2)(Int, Int)

Type declarations and Type instances are memoized.

This commit also simplifies the handling of base types (Bool, Int,
Real), by making them **instances** of PySMTType (and not subclasses).

Other changes:
- formula.py: typename for Symbols must by an instance of PySMTTypes
- type_checker.py: Simplified handling of Equals. This also simplifies
  the handling of custom types
- environment.py: Introduce the TypeManager object
- PySMTType.type_id was removed. 
- Types are now accessible from pysmt.shortcuts


marcogario:
@mikand , @agriggio : Is this API for sorts reasonable? My understanding is that declare-sort is more important than define-sort. In practice, we rarely need a define-sort, since we can build the more complex type directly.

marcogario:
This implementation is not thread-safe anymore (due to the use of auto-generated ids for comparison). This can be probably fixed, by changing the equality check.

The use of custom sorts probably requires moving (part of) the typing information within the environment, and extending contextualization methods accordingly.

marcogario:
The latest commit fixes the issue of thread safety. A Type Manager class was added to the environment, in order to handle types. Utility methods were added in typing.py in order to keep the interface backwards compatible.

This removes also the attribute type_id for types.

pysmt/formula.py
```diff
@@ -939,9 +941,14 @@ def normalize(self, formula):
               obtain f_b that is the formula f_a expressed on the
               FormulaManager b : f_b = b.normalize(f_a)
         """
-        # TODO: name clash with formula normalization
-        # TODO: Move this out of the manager and into ad-hoc function
-        normalizer = IdentityDagWalker(self.env)
+        # TODO: Move this to environment.py
+        class FormulaContextualizer(IdentityDagWalker):
+            def walk_symbol(self, formula, **kwargs):
```
mikand: Actually, also ArrayValues have the index type as payloads and need to be normalized

pysmt/typing.py
```diff
@@ -219,11 +220,21 @@ class _FunctionType(PySMTType):
...
-        self._param_types = param_types
+        self._param_types = tuple(param_types)
         self._hash = hash(return_type) + sum(hash(p) for p in param_types)
+        # Note:
+        # To simplify handling of types, we need to treat FunctionType
+        # in a special way. We do so by setting its arity to a
+        # negative value and creating an args that is the concatenation of
+        # return_type and param_types. In this way, we can iterate on args.
+        self.arity = -1
+        self.args = (self._return_type,) + self.param_types
```
mikand: This is quite ugly, why can't we set the arity to len(self.param_types) and the return type as either an additional field or an extra cell of args?

mikand:
LGTM apart for the few notes I added. 

In addition, I had the following idea:
why can't we import the type constructors (```INT```, ```REAL```, ...) in the ```shortcuts``` module? I was experimenting with ipython on the command line and is quite inconvenient to import both the ```shortcuts``` and the ```typing``` modules all the times.

mikand:
@marcogario What is the status of this PR?

The note concerning the arity of FunctionType is unanswered: why setting -1 as arity instead of n+1 where n is the number of function argument? (It seems to me that you never use the -1 anyway)

If you have good reasons to keep it like this we can merge like this, otherwise I can change this...

marcogario:
I fixed the commend on normalization of ArrayValue, and added the associated test.

pysmt/typing.py
marcogario: I clarified the comment there, because it took me some time to recall what I did :)

The idea is that types/sorts create a tree similar to FNode. In FNode we create a distinction btw payload and args. In order to capture the difference between return value and parameters in FunctionType as you suggest, we would need to make the same distinction here. I do not see a big benefit in this. On the contrary, treating all arguments together allows us to iterate over them more easily, and use PySMTType as key in hash functions.

I have no idea why I used self.arity = -1 . I changed it to be consistent with the other cases and be ```len(self.args)```. In fact, we might consider converting this into a property.

PR #379, opened 2016-10-23
MSat: Back conversion using TAGs

marcogario:
Refactor the walk_back in order to replace big `if...elif` with a map. This relies on the possibility of using the TAG for a msat_decl, and use that directly to select the function to perform back conversion. This is now similar to what we do for Z3. 

At the same time, we use the TAGs to perform the same sort of refactoring for the function _get_signature.

### DIFF
--- a/pysmt/solvers/msat.py
+++ b/pysmt/solvers/msat.py
@@ -531,10 +531,6 @@ class MSatConverter(Converter, DagWalker):
         t = self.env.stc.get_type(args[0])
         return types.FunctionType(t, [t, t])
 
-    def _sig_binary(self, term, args):
-        t = self.env.stc.get_type(args[0])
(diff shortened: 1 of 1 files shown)


pysmt/solvers/msat.py
```diff
@@ -423,13 +519,78 @@ def _get_signature(self, term, args):
...
+            return self.term_sig[tag](term, args)
+        except KeyError:
+            raise PysmtTypeError("Unsupported expression:",
+                                 mathsat.msat_term_repr(term))
 
+    def _sig_binary(self, term, args):
+        t = self.env.stc.get_type(args[0])
+        return types.FunctionType(t, [t, t])
+
+    def _sig_binary(self, term, args):
```
mikand: Duplicated Function!

PR #385, opened 2016-11-06
Fix pysmt for z3 solver on windows

fontealpina:
This PR does:  
- fix default `smt_solvers` directory in order to be properly handled also on Win  
- fix the message for the option `--env` for Win
- fix `installers\z3.py` in order to install z3 python bindings on Win.

Check done on my win 10 - 64 bit machine:
- calling `install.py  --check`, z3 solver is 'True'
- run `nose .` and the result is 'OK'

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,18 +9,44 @@ environment:
     - PYTHON: "C:\\Python27"
       PYTHON_VERSION: "2.7.8"
       PYTHON_ARCH: "32"
+      PYSMT_SOLVERS: "None"
 
-    # - PYTHON: "C:\\Python27-x64"
-    #   PYTHON_VERSION: "2.7.8"
-    #   PYTHON_ARCH: "64"
+    - PYTHON: "C:\\Python27-x64"
+      PYTHON_VERSION: "2.7.8"
+      PYTHON_ARCH: "64"
+      PYSMT_SOLVERS: "None"
 
-    # - PYTHON: "C:\\Python34"
-    #   PYTHON_VERSION: "3.4.1"
-    #   PYTHON_ARCH: "32"
+    - PYTHON: "C:\\Python34"
+      PYTHON_VERSION: "3.4.1"
+      PYTHON_ARCH: "32"
+      PYSMT_SOLVERS: "None"
 
     - PYTHON: "C:\\Python34-x64"
       PYTHON_VERSION: "3.4.1"
       PYTHON_ARCH: "64"
+      PYSMT_SOLVERS: "None"
+
+
+    - PYTHON: "C:\\Python27"
+      PYTHON_VERSION: "2.7.8"
+      PYTHON_ARCH: "32"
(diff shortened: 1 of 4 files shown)


mikand:
Great!

Thanks for taking the time to look into this!

@marcogario This seems perfect to me, do we want to add Appveyor support immediately or we do it in another PR?

fontealpina:
To me is better to modify Appveyor stuff here, so that if there are some issues with the z3 solver, I can fix them directly here.
@marcogario I guess is necessary to update the Appveyor configuration in order to run something like `install.py --z3` and also add the path to pythonpath. 
Plese tell me how can I do it?

fontealpina:
@marcogario also I would suggest to add a `install.py --check` in the job before the execution of the `nosetests` in order to check which solvers have been installed.

mikand:
I managed to make the tests with z3 run on appveyor.
@marcogario I think this PR is ready for acceptance, can you have a look and merge?

marcogario:
@fontealpina thank you very much for doing this. I'll review the changes now.

marcogario:
Note: If I connect to appveyor, I do not see the tests as passed. The tests were executed under mikand user (you can see that if you click on the appveyor link in this page pointing).

pysmt/cmd/installers/z3.py
```diff
@@ -27,14 +27,16 @@ class Z3Installer(SolverInstaller):
     def __init__(self, install_dir, bindings_dir, solver_version,
                  mirror_link=None, osx=None):
         arch = self.architecture
-        if arch == "x86_64":
+        if arch == "x86_64" or arch == "AMD64":
```
mikand: :+1:

appveyor.yml
```diff
@@ -9,18 +9,45 @@ environment:
...
+    - PYTHON: "C:\\Python27-x64"
+      PYTHON_VERSION: "2.7.8"
+      PYTHON_ARCH: "64"
+      PYSMT_SOLVER: "None"
 
-    # - PYTHON: "C:\\Python34"
-    #   PYTHON_VERSION: "3.4.1"
-    #   PYTHON_ARCH: "32"
+    - PYTHON: "C:\\Python34"
+      PYTHON_VERSION: "3.4.1"
```
mikand: It was there already, I think we can try to use 3.5.2 by simply changing the version string

mikand:
@marcogario for some reason, Appveyor was not allowing me to stop the builds, but I could add a pysmt repo to my account that started to be the one used. Can you manage the builds on my account? Are you aware of an "organization mode"?

pysmt/cmd/installers/z3.py
fontealpina: Done. See below.

mikand:
Fixed both the appveyor access issue and the python version

mikand:
@marcogario Any other action required before merging?

Issue #397, opened 2017-02-06
serialization of formulas should use `daggify=True` by default


agriggio:
rationale: if you want exponential blow-up, you should ask for it

mikand:
Are you referring to the serialization of SmtCommands? (https://github.com/pysmt/pysmt/blob/175b9e0d504a8c546fea75a4563752da0b196b36/pysmt/smtlib/script.py#L45)

agriggio:
yes, sorry if it was unclear. we noticed it when dumping scripts

marcogario:
See PR # 399.

Issue #396, opened 2017-02-02
printing of terms in SMT-LIBv2 should be easier


agriggio:
Currently, it seems that printing of individual terms/formulas in SMT-LIB syntax is rather complicated. There is a `write_smtlib` shortcut, but it has several problems:

- it only works for formulas, not other kind of expressions (e.g. `Plus(Symbol("x"), Int(4))`)
- takes a filename as parameter instead of returning a string
- it prints also declarations and other commands, not just an SMT-LIB representation of the term

We would like to have a `to_smtlib()` function that takes a term and returns a string representation of it. Example:

`assert to_smtlib(Plus(Symbol("x"), Int(4))) == "(+ x 4)"`

mikand:
Here is a snippet to support the feature now, I will create a PR for a proper shortcut ASAP.

```python
from pysmt.smtlib.printers import SmtPrinter
from six.moves import cStringIO

def to_smtlib(f):
    buf = cStringIO()
    p = SmtPrinter(buf)
    p.printer(f)
    res = buf.getvalue()
    buf.close()
    return res


if __name__ == "__main__":
    from pysmt.shortcuts import *
    from pysmt.typing import INT
    assert to_smtlib(Plus(Symbol("x", INT), Int(4))) == "(+ x 4)"
```

marcogario:
Shall this be a shortcut, an option of fnode.serialize(), or a new method
of fnode?

mikand:
I'm leaning towards the second option: ```fnode.serialize(smtlib=True)```.

What do you (both) think?

agriggio:
I think it should be the most obvious thing to do, so +1 for the shortcut version. If you want it to be a method, why not having both? E.g:

```
def to_smtlib(f):
    return f.serialize(smtlib=True)
```

marcogario:
We do not have a thresholded printer for smtlib. I think that overloading
.serialize has no advantage, and makes the statement verbose.
More over, we would need to add the parameter in shortcuts.serialize; if we
add the shortcut to_smtlib, this would mean that we would end up with 2
ways of doing the same thing within shortcuts.

I would go with fnode.to_smtlib.  For symmetry we can add the shortcut too
(since we already have it for serialize), although I think it won't be used
much.

Since this method does not depend on the environment, I see no reason not
to implement it within fnode.

@mikand Assign the ticket to whoever should do this.

agriggio:
[...]

PR #399, opened 2017-02-14
Print SmtLib term as string shortcut and default to dag SmtLib representation

mikand:
This PR addresses issues #396 and #397.

I introduced a shortcut ```smtlib_serialize``` and a method of FNode with the same name that return a string representation of the given formula. The ```daggify``` parameter is used to enable or disable the use of ```let``` operators to represent the formulae linearly.

### DIFF
--- a/pysmt/fnode.py
+++ b/pysmt/fnode.py
@@ -17,8 +17,8 @@
 #
 """FNode are the building blocks of formulae."""
 import collections
-
 import pysmt.environment
+import pysmt.smtlib
 from pysmt.operators import (FORALL, EXISTS, AND, OR, NOT, IMPLIES, IFF,
                              SYMBOL, FUNCTION,
                              REAL_CONSTANT, BOOL_CONSTANT, INT_CONSTANT,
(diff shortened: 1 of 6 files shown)


pysmt/fnode.py
```diff
@@ -48,7 +48,7 @@
 from pysmt.constants import (Fraction, is_python_integer,
                              is_python_rational, is_python_boolean)
 from pysmt.exceptions import PysmtValueError, PysmtModeError
-
+import pysmt.smtlib
```
marcogario: Module import should be above elements imports.

pysmt/fnode.py
```diff
@@ -508,6 +508,18 @@ def serialize(self, threshold=None):
         """
         return _env().serializer.serialize(self, threshold=threshold)
 
+    def smtlib_serialize(self, daggify=True):
```
marcogario: Do we really want this mouthful of a name? I thought we mentioned 'to_smtlib' as name.

pysmt/fnode.py
```diff
@@ -508,6 +508,18 @@ def serialize(self, threshold=None):
...
+        """Returns a Smt-Lib string representation of the formula.
+
+        The daggify parameter can be used to switch from a linear-size
+        representation that uses 'let' operators to represnt the
```
marcogario: s/represnt/represent

pysmt/fnode.py
```diff
@@ -508,6 +508,18 @@ def serialize(self, threshold=None):
...
+        formula as a dag or a simpler (but possibly exponential)
+        representation that expalnds the formula as a tree.
```
marcogario: s/expalnds/expands

marcogario: both comments apply to all docstrings of this function.

mikand:
I am done implementing the requested changes.

PR #400, opened 2017-02-14
Suspend Type-Checking Context

mikand:
This PR introduces a context class to suspend the type checking of formulae during construction.

This could be useful to avoid costly type-checking when creating huge formulae that are guaranteed to be correct.

### DIFF
--- a/pysmt/test/test_typechecker.py
+++ b/pysmt/test/test_typechecker.py
@@ -23,7 +23,7 @@ from pysmt.type_checker import (assert_no_boolean_in_args,
 from pysmt.shortcuts import (Symbol, And, Plus, Minus, Times, Equals, Or, Iff,
                              LE, LT, Not, GE, GT, Ite, Bool, Int, Real, Div,
                              Function)
-from pysmt.environment import get_env
+from pysmt.environment import get_env, SuspendTypeChecking
 from pysmt.exceptions import PysmtTypeError
 from pysmt.test import TestCase, main
 from pysmt.test.examples import get_example_formulae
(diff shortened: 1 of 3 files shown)


pysmt/test/test_typechecker.py
```diff
@@ -192,5 +192,16 @@ def test_examples(self):
...
             self.assertIs(f.get_type(), BOOL)
 
+    def test_suspend(self):
+        with self.assertRaises(PysmtTypeError):
+            And(self.x, self.r)
+
+        r = None
+        with SuspendTypeChecking():
+            r = And(self.x, self.r)
+
```
marcogario: Adding this test here would fail:
```python
        with self.assertRaises(PysmtTypeError):
            And(self.x, self.r)
```
Although, I think we agree that it should not.

mikand: This is a very good point indeed...

mikand:
In general I'm also against this hack. Let me explain why I came up with this in the first place.

@ahmedirfan1983 had to parse a HUGE (~100 Mb) smt2 file. The parser takes ~69 seconds on my machine, but if the type-checking is disabled, we go down to ~45 seconds (some more performance, I estimate it in less than 10 seconds, can be squeezed by using MathSAT or Z3 to parse the smt2 file and then use the back-conversion).

Overall, I am also not too happy in having this class around, but I think that in some circumstances it could be useful if you are absolutely sure that you are building type-valid formulae.

What do you think? We could also move the class in the documentation as an hack without including it in the core of pysmt...

marcogario:
That is interesting, but I do not think that it justifies the inclusion in the code. Lets move this into the docs. We can open a new subsection under  http://pysmt.readthedocs.io/en/latest/development.html   called something like "Squeezing for Performance", and include this snippet there with a bit of explanation. I can probably do this in the next few days.

marcogario:
As discussed, this will not be included in the core. However, the relevant code snippet is now part of the documentation (see PR #402).

@ahmedirfan1983 , I'll not delete the branch immediately, in case you are basing of it. Nevertheless,  you should change your code in order to have this fix local to it.

Issue #405, opened 2017-04-28
Circular dependency in pysmt.typing?


rene-rex:
Hello

I get an error on import of the typing module:

```python3 -c "import pysmt.typing"```
```python
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/rex/software/venvs/opt/lib/python3.5/site-packages/pysmt/typing.py", line 500, in <module>
    import pysmt.environment
  File "/home/rex/software/venvs/opt/lib/python3.5/site-packages/pysmt/environment.py", line 24, in <module>
    import pysmt.simplifier
  File "/home/rex/software/venvs/opt/lib/python3.5/site-packages/pysmt/simplifier.py", line 22, in <module>
    import pysmt.typing as types
AttributeError: module 'pysmt' has no attribute 'typing'
```
Maybe this is circular dependency? `types` imports `environment`, which imports `simplifier`, which in turn import `types` again.

mikand:
Hi,

thanks for the report. Indeed this is a circular dependency that needs to be resolved.

We never realized this issue because it suffices to import `pysmt.environment` or `pysmt.shortcuts` before importing  `pysmt.types` for the problem to go away.

marcogario:
Fixed in #406.

PR #406, opened 2017-05-02
Fixed cyclic import errors

mikand:
This PR addresses issue #405 .

I created a test that invokes the python interpreter and checks that each pysmt module (except the testing ones and the solvers) can be imported as a standalone (i.e. without previously importing something else).

In order for this test to work I had to move some imports of the environment module inside some functions, but I think this trade-off is acceptable.

### DIFF
--- a/pysmt/fnode.py
+++ b/pysmt/fnode.py
@@ -17,7 +17,6 @@
 #
 """FNode are the building blocks of formulae."""
 import collections
-import pysmt.environment
 import pysmt.smtlib
 from pysmt.operators import (FORALL, EXISTS, AND, OR, NOT, IMPLIES, IFF,
                              SYMBOL, FUNCTION,
@@ -852,10 +851,12 @@ class FNode(object):
 
 def _env():
     """Aux function to obtain the environment."""
+    import pysmt.environment
     return pysmt.environment.get_env()
 
 def _mgr():
     """Aux function to obtain the formula manager."""
+    import pysmt.environment
     return pysmt.environment.get_env().formula_manager
 
 def _is_bv(node):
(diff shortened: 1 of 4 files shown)


marcogario:
I am not sure that the issue is particularly critical, and I fear that the fix is polluting other functions without a real gain.

The circular dependency mentioned in #405 was introduced after we made the type information become part of the environment. The import of the environment occurs at Line 500 and impacts only 4 types. This was done for backwards compatibility, but if we want to get rid of this, then we should move those into shortcuts.py (indeed, if you look at the implementation they are shortcuts) [1]. 

The other point is whether there are other cyclic imports and whether we want to deal with them. Note that this fix does not remove the dependency itself, but only allows the import to complete. Arguably, one could ask whether the dependency is necessary in the first place or could be refactored away.

The one that I do not like is the one on ``fnode.py:_env`` this function gets called everytime we perform an infix notation operation. Nevertheless, this is needed since we want to access the global environment.

In summary, I would accept the change in ``oracle.py`` and ``fnode.py`` if we do not find another solution. I would prefer to find a better solution for ``typing.py``.

 [1] Since we are waiting to release 0.7.0 this is the moment to break backwards compatibility!

mikand:
@marcogario  You are right.

Check out the changes now, there is a much simpler solution to import ```pysmt.enviornment``` without causing cyclic dependencies, that is importing ```pysmt```. This will make all the modules (not the packages) within the ```pysmt``` package immediately available without executing them.

Do you think this solution is acceptable?

marcogario:
Not 100% sure of why this works, but it looks much better. 

To be sure, I tried generating and installing the pkg and it also works after installing it via pip.

Issue #393, opened 2016-12-14
Upgrade z3 to 4.5.0


marcogario:

marcogario:
This will be more difficult than usual. The latest version changed the way the library is distributed, and I am having problems on my laptop importing it (although, Travis seems to work fine).

Moreover, logics names are not recognized anymore, and therefore we need to understand how to map logics to z3.

marcogario:
Using the nightly build seems to work better ( https://github.com/Z3Prover/bin/commit/542946bc5615a789e9de49316fd4d29a13fb40e6 ) 

1.  libz3.so has the usual problem that we had with other .so . It must be in the LD_LIBRARY_PATH for this to work. Surprisingly, if it is not, the error is rather obscure:
```
ArgumentError: argument 1: <type 'exceptions.TypeError'>: unbound method from_param() must be called with ContextObj instance as first argument (got int instance instead)
```
2. Debian and Ubuntu packages are indeed different. On my Debian I cannot use the Ubuntu anymore. We need to upgrade the installer accordingly.

3. There seems to be a bug with UnsatCore. The following example does not work:
```
from pysmt.shortcuts import *
from pysmt.typing import INT, REAL, FunctionType


p = Symbol("p", INT)
q = Symbol("q", INT)
r = Symbol("r", REAL)
s = Symbol("s", REAL)

ih = Symbol("ih", FunctionType(INT, [REAL, INT]))

f = Not(Implies(And(Equals(ToReal(Minus(p, Int(3))), r),
                    Equals(ToReal(q), r)),
            Or(GT(Function(ih, (ToReal(Minus(p, Int(3))),
                            Plus(q, Int(3)))), p),
               LE(Function(ih, (r, p)), p))))

with Solver(unsat_cores_mode="named") as solver:
    clauses = [f]

    for i,c in enumerate(clauses):
        solver.add_assertion(c, "a%d" % i)

        r = solver.solve()
        print(r) # This should be false
```
4. The issue with unrecognized logics persists. It seems like ```SolverFor``` only accepts a valid SMT-LIB logic.

PR #407, opened 2017-05-04
Z3: Upgrade to 4.5.1 (dev)

mikand:
This PR brings the pysmt Z3 version to 4.5.0, addressing issue #393.

Unfortunately, right now the tests fail (and Z3 cannot be updated) due to the following Z3 issue  https://github.com/Z3Prover/z3/issues/1005.

### DIFF
--- a/pysmt/cmd/installers/base.py
+++ b/pysmt/cmd/installers/base.py
@@ -90,9 +90,9 @@ class SolverInstaller(object):
 
     def download_links(self):
         if self.mirror_link is not None:
-            yield self.mirror_link.format(archive_name=self.archive_name)
+            yield self.mirror_link.format(archive_name=self.archive_name, solver_version=self.solver_version)
         if self.native_link is not None:
-            yield self.native_link.format(archive_name=self.archive_name)
+            yield self.native_link.format(archive_name=self.archive_name, solver_version=self.solver_version)
 
     def download(self):
         """Downloads the archive from one of the mirrors"""
--- a/pysmt/solvers/z3.py
+++ b/pysmt/solvers/z3.py
@@ -143,7 +143,10 @@ class Z3Solver(IncrementalTrackingSolver, UnsatCoreSolver,
                                            environment=environment,
                                            logic=logic,
                                            **options)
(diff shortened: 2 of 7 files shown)


marcogario:
Z3 Closed it in commit https://github.com/Z3Prover/z3/commit/7e1fae418a766532f99cea06a1c6021268661864

Maybe we can move forward with this?

mikand:
The issue is fixed, but unless we change pysmt-install to compile z3 from git master, we have to wait for a new release....

marcogario:
Can we try with the nightly build, to make sure that there is no other blocking issue?

marcogario:
We are using a frozen copy of the bin repo from Z3. https://github.com/pysmt/z3bin
In this way we can use a fairly recent version of Z3, while guaranteeing that we are using always the same version.

Note: This commit depends on PR #431 .

pysmt/cmd/check_module_version.py
```diff
@@ -0,0 +1,17 @@
+import sys
```
mikand: This file should not be needed anymore after #431

pysmt/cmd/installers/base.py
```diff
@@ -37,6 +41,36 @@ def TemporaryPath(path):
         sys.path = old_path
 
 
+def delete_module(modname, paranoid=None):
```
mikand: This function is a leftover from my previous experiments before subprocess, I guess it can be removed

pysmt/solvers/z3.py
```diff
@@ -905,14 +908,14 @@ def eliminate_quantifiers(self, formula):
...
-        simplifier = z3.Tactic('simplify')
+        #simplifier = z3.Tactic('simplify')
         eliminator = z3.Tactic('qe')
 
         f = self.converter.convert(formula)
-        s = simplifier(f, elim_and=True,
-                       pull_cheap_ite=True,
-                       ite_extra_rules=True).as_expr()
-        res = eliminator(s, eliminate_variables_as_block=True).as_expr()
+        # s = simplifier(f, elim_and=True,
```
mikand: This change is another leftover experiment, it should not be needed here

Issue #216, opened 2015-11-12
SMT-LIB Compliant Interface


marcogario:
pySMT can simulate an SMT-LIB solver, by reading an SMT-LIB script and execute it on one of the solvers. This however, is limited to the most commonly used commands.

Evaluation of all commands (including SMT-LIB 2.5) would be desirable.

marcogario:
In the parser by Griggio there are 6 test files. We currently only pass 2, and should aim for 100%. Nevertheless, we are able to parse most of the files of the SMT-LIB. This somehow suggests that the SMT-LIB benchmark uses a tiny fraction of the SMT-LIB language.

https://es-static.fbk.eu/people/griggio/misc/smtlib2parser.html

marcogario:
See branch: griggio_smtlib_tests 
This depends on the support for sorts.

marcogario:
Support for Sorts is not dependent on PR #411 .

We are currently able to parse all examples from Griggio.

PR #411, opened 2017-05-25
Extend parser to support declare-sort/define-sort

marcogario:
I added the test files from Griggio's parser. The goal is to get closer to 100% support of SMT-LIB (see #216).

I think a few things need to change in the parser:
1. The types are now local to the environment, therefore the parser should not use the shortcuts for the types (DONE)
2. Parsing of types is currently fragmented and performed based on string matching. A type is a recursive structure, and should be parsed in a similar way as to what we do with formulae. I got stuck here. @mikand, can you take a look and see what needs to change?

I already wrote the tests, and they should be failing.


pysmt/smtlib/parser.py
```diff
@@ -855,44 +834,90 @@ def parse_atoms(self, tokens, command, min_size, max_size=None):
...
+                        for _ in range(base_type.arity):
+                            ty = self.parse_type(tokens, command, type_params=type_params)
+                            pparams.append(ty)
+                            if isinstance(ty, tuple):
+                                has_free_params = True
+                        if has_free_params:
+                            def definition(*args):
+                                params = []
+                                for x in pparams:
+                                    if isinstance(x, tuple):
```
mikand: This not super clean, but it works. A principle dolution would be to represent the type-tree originating from a `define-sort` and provide a replace method once the types are instantiated

mikand:
@marcogario It should work now, I added only minimal tests.

pysmt/smtlib/parser.py
marcogario: I think this is good for now. In my experience the types do not get very complex.

marcogario:
@mikand I added an example with sorts, and this now requires adding support for sorts in the solvers. Are sorts part of UF? Shall we include the information of whether sorts are present in the theory description?

I was trying to extend MathSAT, but I cannot find how to create a sort that has arity > 0. If this is not possible, then we need to extend the theory descriptions.

mikand:
I really do not know how this whole uninterpreted sort thing works, maybe @agriggio can answer this?

marcogario:
I thought more about the subject. I think i have reached the conclusion that you only need simple sorts (arity 0) when doing solving. If you have a sort of arity 1,you still need to instantiate it in order to be able to use it (eg to declare a function). When you instantiate it, however you get a sort of arity 0. 

Therefore, the only thing that we need to do is to have a unique name of the instantiated sorts, and then we can create simple sorts with the given name.

marcogario:
@mikand In order to complete this, we need to address the following issue. Boolector (and maybe other solvers) do not support the declaration of sorts (if you try to feed an SMT-LIB file with declare-sort, you get an error saying that it is not supported). Apart from this, the SAT solvers will not support this. 

My suggestion is that we extend the Theory class to include a ``custom_type`` field. This is similar to what we did for ``arrays_const``. I think this is the cleanest solution, but I would like some feedback.

marcogario:
The rewriting of the parsing with the freshsymbols in quantified expressions seems to have slowed down the overall parsing performance. Shall we make this rewriting optional?

In most of the cases, it is probably sane to assume that symbols have distinct names, and no rewriting needs to occur.

mikand:
@marcogario I like the idea of the `custom_type` field.

Concerning the slowdown, do you have some numbers? In case, I can try and see if there is something to optimize...

marcogario:
@mikand , The problem seems to be in the serialization and not the parsing See #416 . No need to optimize the parser for now. Thanks.

marcogario:
@mikand this branch is ready for review and potentially merge, when you have time.

mikand:
@marcogario Wow! This is a terrific amount of work! :+1:

PR #413, opened 2017-06-05
Examples: Equivalence checking of XOR rewriting

marcogario:
A simple example of equivalence checking using BitVectors.


mikand:
@marcogario This is a very interesting encoding of XOR and a very useful example for pysmt. For me this can be merged right now, but since the PR is not assigned to me I am not sure whether this is complete or not.

PR #415, opened 2017-06-18
Improved simplification effectiveness on conjunctions and disjunctions

mikand:
This PR changes the simplification rules for conjunction and disjunction operators.

The simplification now does three things
- Flattens the And of Ands and the Or of Ors: And(x, And(y, z)) -> And(x, y, z)
- Recognizes the conjunction of x and Not(x) yielding False and the disjunction of x and Not(x) yielding True
- Transforms the argument in a set, thus removing the duplicates: And(x, x, y) -> And(x, y)

### DIFF
--- a/pysmt/simplifier.py
+++ b/pysmt/simplifier.py
@@ -79,28 +79,54 @@ class Simplifier(pysmt.walkers.DagWalker):
         return res
 
     def walk_and(self, formula, args, **kwargs):
-        args = [x for x in args if not x.is_true()]
-        if len(args) == 0:
-            return self.manager.TRUE()
-        elif len(args) == 1:
-            return args[0]
-        else:
-            if any(x.is_false() for x in args):
+        new_args = set()
+        for a in args:
+            if a.is_true():
+                continue
+            if a.is_false():
                 return self.manager.FALSE()
 
-        return self.manager.And(args)
+            if a.is_and():
+                for s in a.args():
+                    if self.walk_not(self.manager.Not(s), [s]) in new_args:
+                        return self.manager.FALSE()
+                    new_args.add(s)
+            else:
+                if self.walk_not(self.manager.Not(a), [a]) in new_args:
+                    return self.manager.FALSE()
+                new_args.add(a)
+
+        if len(new_args) == 0:
+            return self.manager.TRUE()
+        elif len(new_args) == 1:
+            return next(iter(new_args))
+        else:
+            return self.manager.And(new_args)
 
     def walk_or(self, formula, args, **kwargs):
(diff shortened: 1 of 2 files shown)


marcogario:
LGTM. The main question is how big is the performance hit on big formulae, due to the negated term search.
The worst case this is O(n^2), right? Can we get some number, for example the ones of the SMT-LIB?

mikand:
Well, actually the amortized  complexity is just O(n) because you check membership in set that uses hashes... However, I can try to set up a simple scalable test.

mikand:
Here are the results.

The test script I used is the following, it stresses the worst case for the simplifier, i.e. when no simplifications are possible.
```python
import sys

from pysmt.shortcuts import And, FreshSymbol, BOOL

size = int(sys.argv[1])
f = And(FreshSymbol(BOOL) for _ in range(size))

g = f.simplify()

assert len(g.args()) == size
```

This is the timing.
```shell
$ git checkout master 
$ for i in 1000 10000 100000 500000 1000000; do echo ""; echo "Testing with arity: $i"; time python3 simp_test.py $i; done

Testing with arity: 1000

real	0m0.124s
user	0m0.120s
sys	0m0.004s

Testing with arity: 10000

real	0m0.258s
user	0m0.256s
sys	0m0.000s

Testing with arity: 100000

real	0m2.077s
user	0m2.028s
sys	0m0.048s

Testing with arity: 500000

real	0m10.503s
user	0m10.388s
sys	0m0.108s

Testing with arity: 1000000

real	0m21.201s
user	0m21.060s
sys	0m0.136s
$ git checkout better-and-or-simplifications 
$ for i in 1000 10000 100000 500000 1000000; do echo ""; echo "Testing with arity: $i"; time python3 simp_test.py $i; done

Testing with arity: 1000

real	0m0.131s
user	0m0.120s
sys	0m0.004s

Testing with arity: 10000

real	0m0.415s
user	0m0.408s
sys	0m0.004s

Testing with arity: 100000

real	0m3.713s
user	0m3.680s
sys	0m0.032s

Testing with arity: 500000

real	0m18.588s
user	0m18.464s
sys	0m0.124s

Testing with arity: 1000000

real	0m38.021s
user	0m37.784s
sys	0m0.224s
```

As you can see, there is a slowdown due to the "manual" passing of the arguemnt list, but the complexity clearly grows linearly.
Do you think that this is acceptable?

marcogario:
You are right on the complexity. The slowdown is a 2x in general, since you need to create and check Not(x)  for each x. 

I think this is ok. We might consider to make this simplification optional if we run into performance problems.

PR #419, opened 2017-07-24
CVC4: Free C++ SmtEngine when Solver is destroyed

marcogario:
The SWIG wrapper is not handling properly the ownership of either SmtEngine or ExprManager, resulting in a memory leak: When the python objects are release the underlying C++ are not destroyed.

This commit works-around the issue by forcing the ownership of the SmtEngine to the python wrapper. When the wrapper is destroyed, the C++ is also destroyed.

A proper patch for CVC4 should be designed in order to properly account also for the release of the ExprManager, and the fact that ExprManager is ref'd within SmtEngine (but SWIG has no way of knowing this).

See http://www.swig.org/Doc1.3/Python.html#Python_nn30 for an explanation of the memory management features in SWIG.

### DIFF
--- a/pysmt/solvers/cvc4.py
+++ b/pysmt/solvers/cvc4.py
@@ -97,7 +97,9 @@ class CVC4Solver(Solver, SmtLibBasicSolver, SmtLibIgnoreMixin):
 
     def reset_assertions(self):
         del self.cvc4
-        self.cvc4 = CVC4.SmtEngine(self.em)
+        # CVC4's SWIG interface is not acquiring ownership of the
+        # SmtEngine object. Forcing it here.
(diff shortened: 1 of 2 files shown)


pysmt/solvers/cvc4.py
```diff
@@ -97,7 +97,9 @@ def __init__(self, environment, logic, **options):
 
     def reset_assertions(self):
         del self.cvc4
-        self.cvc4 = CVC4.SmtEngine(self.em)
+        # CVC4's SWIG interface is not acquiring ownership of the
+        # SmtEngine object. Forcing it here.
+        self.cvc4 = CVC4.SmtEngine(self.em); self.cvc4.thisown=1
```
mikand: I didn't know this feature of SWIG :+1:

PR #429, opened 2017-08-10
SMTLIB: Fixed bug in declare-const parsing

marcogario:
Thanks to @ahmedirfan1983 for reporting this.


mikand:
@marcogario Perfect!

PR #430, opened 2017-08-10
Parser debugging info

marcogario:
Add line and column number to parser exceptions.


marcogario:
@mikand  what do you think about this? I am stll unclean on when do we use the interactive mode.  

We could also extene this to store the  10 preceeding  characters to provide more context.

mikand:
@marcogario I like the idea and the implementation. We have to do some benchmarking to see if there is a significant performance penalty in doing this.

Concerning the interactive mode, we use it in the SMTLibSolver interface (https://github.com/pysmt/pysmt/blob/175b9e0d504a8c546fea75a4563752da0b196b36/pysmt/smtlib/solver.py#L89)

I think that a reasonable solution would be to simply avoid the debug info (line and col) in interactive mode.

marcogario:
debug python 2

The mean execution time was 2.19 seconds
The max execution time was 33.15 seconds

debug python 3.5

The mean execution time was 3.41 seconds
The max execution time was 84.54 seconds

debug pypy

The mean execution time was 0.93 seconds
The max execution time was 19.05 seconds

master python 2

The mean execution time was 1.62 seconds
The max execution time was 33.43 seconds

master python 3.5

The mean execution time was 2.41 seconds
The max execution time was 60.57 seconds

master pypy

The mean execution time was 0.75 seconds
The max execution time was 37.29 seconds

marcogario:
As shown from the preliminary performance analysis, the debugging information does sloe down the parsing a bit.  My current suggestion would be to make this conditional to the __debug__ constant. For performance you should always run python -O, and this would disable the debugging information in the parser (plus the aasertions). Another option is to make this a flag in the parser and explain in the exception message how you can activate the flag. 

Preferences?

marcogario:
The debug information are enabled by default, however they are preconditioned by the ``__debug__`` variable, and will not be used when running python with -O.

PR #431, opened 2017-08-12
Installers: Simpler version check via subprocess

marcogario:
Separated and simplified from PR #407. 

Main changes:

- Renamed check_module_version.py into check_version.py; fixed its internal structure and changed return value from "ERROR" to "NOT INSTALLED". If there is an error, the script will exit with a non-zero status, and we can print this information for debugging
- Fixed installation for cvc4, btor and picosat. Since we are not using os.system anymore, we need to perform each command separately and we cannot rely on the shell patterns.
- Removed some left-overs functions from previous attempts

### DIFF
--- a/pysmt/cmd/check_version.py
+++ b/pysmt/cmd/check_version.py
@@ -1,0 +1,79 @@
+# Copyright 2014 Andrea Micheli and Marco Gario
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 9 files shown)


pysmt/cmd/check_version.py
```diff
@@ -0,0 +1,76 @@
...
+        elif module == "picosat":
+            import picosat
+            version = picosat.picosat_version()
+
+        elif module == "yices":
+            import yicespy
+            v = yicespy.__dict__['__YICES_VERSION']
+            m = yicespy.__dict__['__YICES_VERSION_MAJOR']
+            p = yicespy.__dict__['__YICES_VERSION_PATCHLEVEL']
+            version = "%d.%d.%d" % (v, m, p)
```
mikand: It would be nice to have an error if the module is not among the recognized ones

marcogario: Done in f0e15a4

PR #432, opened 2017-08-14
Cython/smtlib parser

marcogario:
Cython: SMT-LIB Parser compilation
The SMT-LIB parser module is now compiled using Cython behind the
scenes. By default pySMT will try to use the cython version but the
behavior can be controlled via env variables:

- PYSMT_CYTHON=False, disable Cython
- PYSMT_CYTHON=True,  force Cython: Raises an error if cython or the
                      SMT-LIB parser module are not available.
- unset PYSMT_CYTHON, defaults to Cython but silently falls back to
                      pure-python version

The API of pysmt.smtlib.parser does not change, and this should not
impact any code. Notice that currently parser.pyx is a copy of
parser_py.py . Next steps require moving common functions into
__init__.py, and explore type annotation for the pyx file.

Benchmarking on parse_all.py shows:

$ PYSMT_CYTHON=True python3.5 parse_all.py --count 500
  The mean execution time was 2.34 seconds
  The max execution time was 59.77 seconds

$ PYSMT_CYTHON=False python3.5 parse_all.py --count 500
  The mean execution time was 3.39 seconds
  The max execution time was 85.46 seconds


marcogario:
This builds on top of #430 because I wanted to understand the impact of the debug information. 
Overall, I did not do anything to help cython perform better, ans I think much more can be squeezed out. 

@ahmedirfan1983 would you mind trying this branch out and let me know if it makes a difference for your usecases?  It should be sufficient to install cython via pip.

mikand:
@marcogario This is great! only issue I see is that there is qite a lot of code duplication. One suggestion: why not duplicate in cython only the tokenizer instead of the whole parser and structures? My guess is that the perfomance penalty for python is just there....

marcogario:
Yes. A lot depends on how much cython specific features we end up using. If we are not using anything, then the pyx file can be a symlink, or copied at install time. 

I agree with keeping only a few classes in the pyx. I won't have time until next weekend to work on this. If you are interested,  please assign the issue to yourself and try.

ahmedirfan1983:
I tried the new parser on a 97MB smt instance. Here are the timings for the parsing (with type-checker disabled) the file:

* Master                      -- 48.5 secs
* cython/smtlib parser -- 25.4 secs


I would say that it is a very good improvement.

marcogario:
I worked around the standard way of pyximport, and direcly compile the .py file. In this way we avoid code duplication. However, this means that we need to keep the file in pure python and cannot annotate it. I think this is fine, unless we plan to work more on squeezing performances out of the parser. 

I tried cythonizing only the tokenizer and while it performs well, the bulk of the the time is really spent in the parser.

This depends on (includes) #430.

marcogario:
@mikand Do you think the current solution is ok, or do we need more discussion?

mikand:
I like your solution, I wanted to try to see if we could write some cython code to speed-up the parser even more (e.g. use fopen() to read char by char), but I will have no time for this until the end of september, so we can maybe merge this and then possibly refine in nother PR. You can go ahead and merge if you agree with this second option.

PR #434, opened 2017-08-19
Ported MathSAT to 5.4.1

mikand:


marcogario:
I think this has the linking problem that we discussed a couple months ago (sorry for not following up). If we cannot fix it ourselves, then we need it to be fixed upstream. 

I would not change the CI to allow this to pass.

mikand:
@agriggio will the next version of MathSAT support ubuntu 14.04 for python bindings compilation?
Right now, the `libmathsat.a` does not link on 14.04, so we either need another `.a` library or an `.so`...

PR #438, opened 2017-09-09
Infix: Add NotEquals

marcogario:
When using infix notation, it would be nice to write ```x != y```. However, this is not possible for the same reasons that we do not allow ``` x == y```. The current way to write this in infix notation would be: ```~ x.Equals(y)```, that is not particularly readable in my opinion. 

This introduces a new function in fnode so that we can write ```x.NotEquals(y)```.

### DIFF
--- a/pysmt/fnode.py
+++ b/pysmt/fnode.py
@@ -681,6 +681,9 @@ class FNode(object):
     def Equals(self, right):
         return self._apply_infix(right, _mgr().Equals)
 
+    def NotEquals(self, right):
+        return self._apply_infix(right, _mgr().NotEquals)
+
     def Ite(self, then_, else_):
         if _env().enable_infix_notation:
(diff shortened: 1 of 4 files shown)


marcogario:
I am open for discussion on this one, since there might be better solutions.

mikand:
I like the idea, and I think we could push it even further by having a method `formula_manager().NotEquals()` and a shortcut `NotEquals()`. The latter is not particularly useful because one can do `Not(Equals(a,b))`, but the former is very useful because `manager.Not(manager.Equals(a,b))` is quite ugly wrt `manager.NotEquals(a, b)`. What do you think?

marcogario:
Ok, and then the infix would be calling the one in formula manager. I think
for consistency it is better to have it also in shortcuts, but I agree on
it not being very useful. I will try to change it later today.

PR #444, opened 2017-10-12
Added first Jupyter tutorial for Boolean logic 

mikand:


mikand:
GitHub now has a viewer for these files: https://github.com/pysmt/pysmt/blob/tutorial/docs/notebooks/boolean_logic.ipynb

marcogario:
How can we integrate this with the documentation? What I do not like about jupyter notebook is that files cannot be used outside jupyter. For example, there is no obvious way to execute this notebook when we run the CI. 

For these reasons, I would keep /examples as pure python. However, we have docs/tutorials.rst that (currently) imports every file in /examples . Maybe it would make sense to have this in Jupyter format, and exported as HTML for the documentation?

mikand:
Actually, you can execute a notebook:
```
$ jupyter-nbconvert --to python boolean_logic.ipynb
$ python boolean_logic.py
```

and you can also automatically generate the markdown in the very same way:
```
$ jupyter-nbconvert --to markdown boolean_logic.ipynb
$ cat boolean_logic.md
```

I think I can include a test of this in Travis, not sure how to include this in the docuemntation on readthedocs...

marcogario:
Our documentation is in restructured text. If we can generate that, then I think we can make a list of some selected examples that would be nice to have as notebooks and manually generate the source for the docs. I do not think that we need to automate the generation from the notebooks to the doc.

mikand:
There is a converter also for rst:
```$ jupyter-nbconvert --to python boolean_logic.ipynb``` 

However, I do not know how well it performs for complex markup in the cells.

mikand:
Closing in favor of #535

PR #445, opened 2017-10-13
Added support for EUF infix notation

### LINKED ISSUE #239: Overload infix notation for  EUF
The only theory that does not currently support of infix notation is UF. This issue is to discuss possible problems and syntax issues related to extending UF with infix notation. In particular, a possible idea is to use a syntax like

```
f(3) #if f is a function symbol, this should return the application of 3 to f
```

It is unclear how much interest there is in this. Also, a few examples would be useful to better understand the implications of this.

mikand:
This closes #239

### DIFF
--- a/pysmt/fnode.py
+++ b/pysmt/fnode.py
@@ -650,26 +650,32 @@ class FNode(object):
     # Infix Notation
     @assert_infix_enabled
     def _apply_infix(self, right, function, bv_function=None):
-        mgr = _mgr()
-        # BVs
         # Default bv_function to function
-        if bv_function is None: bv_function = function
+        if bv_function is None:
+            bv_function = function
+        right = self._infix_prepare_arg(right, self.get_type())
         if self.get_type().is_bv_type():
-            if is_python_integer(right):
-                right = mgr.BV(right, width=self.bv_width())
             return bv_function(self, right)
-        # Boolean, Integer and Arithmetic
-        if is_python_boolean(right):
-            right = mgr.Bool(right)
-        elif is_python_integer(right):
-            ty = self.get_type()
-            if ty.is_real_type():
-                right = mgr.Real(right)
-            else:
-                right = mgr.Int(right)
-        elif is_python_rational(right):
-            right = mgr.Real(right)
         return function(self, right)
+
+    @assert_infix_enabled
+    def _infix_prepare_arg(self, arg, expected_type):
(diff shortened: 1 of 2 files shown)


pysmt/fnode.py
```diff
@@ -854,6 +861,11 @@ def __rshift__(self, right):
 
     def __mod__(self, right):
         return self._apply_infix(right, None, bv_function=_mgr().BVURem)
+
+    def __call__(self, *args):
```
marcogario: If args is empty ``f()`` the iterator will not execute and neither  ``_infix_prepare_args``. 

Also, don't you should not check against the type of the function, but the type of the argument. See comment below.

pysmt/test/test_euf.py
```diff
@@ -37,6 +37,22 @@ def test_euf(self):
...
+        f = Symbol("f", ftype1)
+        g = Symbol("g", ftype2)
+
+        with self.assertRaises(PysmtModeError):
+            f(1.0)
+
+        get_env().enable_infix_notation = True
+        check = Equals(f(1.0), g(2.0, 4))
+
+        self.assertSat(check, logic=UFLIRA,
```
marcogario: We do not need a solver for this. Let's just check that the infix and prefix formula yield the same result.

pysmt/test/test_euf.py
```diff
@@ -37,6 +37,22 @@ def test_euf(self):
         self.assertSat(check, logic=UFLIRA,
                        msg="Formula was expected to be sat")
 
+    @skipIfNoSolverForLogic(UFLIRA)
+    def test_infix(self):
+        ftype1 = FunctionType(REAL, [REAL])
+        ftype2 = FunctionType(REAL, [REAL, INT])
```
marcogario: Can you add a test case for f: Int x BV8 -> Real? It seems to me that the implementation that you provide should not work with BV. When you convert the args, you should look into the function type definition and for each positional argument detect the type... I am confused on why this works for INT...

pysmt/fnode.py
mikand: Done.

pysmt/test/test_euf.py
mikand: OK

pysmt/test/test_euf.py
mikand: Done.

PR #450, opened 2017-10-22
BV2NAT: Add bv2nat operator

marcogario:
Implement the bv2nat that takes a bitvector and returns the associated
integer by interpreting the bitvector as an unsigned integer.


mikand:
This is great. My only concern is about the name. `BvToNat` is not adherent with the smtlib standard where this function is called `bv2nat` and is not particularly explicit (`BvToNatural` or `BvToInteger` would be more explicit and more similar to the existing `ToReal`). 

Also have a look at this discussion: https://github.com/Z3Prover/z3/issues/1252

What do you think?

pysmt/smtlib/parser/parser.py
```diff
@@ -1304,6 +1305,8 @@ def __init__(self, environment=None, interactive=False):
                                 self._operator_adapter(self._ext_rotate_left)
         self.interpreted["ext_rotate_right"] =\
                                 self._operator_adapter(self._ext_rotate_right)
+        mgr = self.env.formula_manager
+        self.interpreted['bv2int'] = self._operator_adapter(mgr.BVToNat)
```
mikand: I think this should be
```python
self.interpreted['bv2nat`] = self._operator_adapter(mgr.BVToNat)
```
AFAIK `bv2int` is z3-specific, but we might support it as an alias of `bv2nat` (see https://github.com/Z3Prover/z3/issues/1252)

mikand:
Also, it would be great to add a couple of smtlib2 instances with this feature to test the parser.

pysmt/smtlib/parser/parser.py
marcogario: This is an extension of the main parser to deal with z3. z3 outputs bv2int when dumping SMT-LIB.

marcogario: This is an extension of the main parser to deal with z3. z3 outputs bv2int when dumping SMT-LIB.

marcogario:
:+1 for BvToNatural: I think it is important to make it clear that we consider an unsigned representation. For this reason I do not like BVToInteger. MathSAT has 2 different operators depending on the interpretation (signed or unsigned). 
**Note** that bv2nat is not part of SMT-LIB. It is only used in the BV theory description to describe the semantics of bitvectors. Not all solvers support the combination BV/Integers. I think we can be flexible here, but I would keep the SMT-LIB name as ``bv2nat``. There are no examples in the SMT-LIB using the operator (for the above reason). By adding one example (or more) example in ``examples.py`` we are stimulating both the parser and printer. I do not see the advantage of adding artificial SMT-LIB files for this. 

I can add a few more tests of the bv2nat conversion, with more complex expressions, simplification etc.

mikand:
Ah now I see that `bv2nat` is just in the semantics description.

Then, I agree that we are fine with the tests we have. Are you going to rename `BVToNat` to `BVToNatural` or can I accept the PR?

PR #452, opened 2017-10-26
Fixed type-checking of Equals, LT and LE.

mikand:
`LT` and `LE` are only allowed over (INT, INT) and (REAL, REAL), while equals is allowed for ('a, 'a) except if `a is BOOL.

The previous behavior allowed `Equals(BOOL, BOOL)` and also `LT(BV[8], BV[8])`


pysmt/test/test_regressions.py
```diff
@@ -470,6 +470,14 @@ def test_parse_bvconst_width(self):
         const = expr.args()[0]
         self.assertEqual(const.bv_width(), 8, const.bv_width())
 
+    def test_equality_typing(self):
```
marcogario: Add test for LT, LE

mikand: Done.

PR #453, opened 2017-10-31
MathSAT update to 5.5.1 and windows support

mikand:
This PR updates teh MathSAT version to 5.5.1 and adds installation support and CI under windows


pysmt/cmd/installers/msat.py
```diff
@@ -42,23 +77,60 @@ def __init__(self, install_dir, bindings_dir, solver_version,
 
         self.python_bindings_dir = os.path.join(self.extract_path, "python")
 
+
     def compile(self):
+        # Patching the swig wrapper
```
marcogario: Not a fan of patching the solver here. The first change should go away after upgrading to 5.5.0. What about the new setup.py? We have a dedicated repo to store patches for the solvers, that is the preferred way of doing this.

appveyor.yml
```diff
@@ -6,48 +6,70 @@ environment:
...
+      PYTHON_VERSION: "2.7.x"
+      PYTHON_ARCH: "32"
+      PYSMT_SOLVER: "msat"
+
+    - PYTHON: "C:\\Python27-x64"
+      PYTHON_VERSION: "2.7.x"
+      PYTHON_ARCH: "64"
+      PYSMT_SOLVER: "msat"
+
+    # There is a problem with python 3.5+ with the compilation of
```
marcogario: Did you report this issue to appveyor?

pysmt/cmd/installers/msat.py
mikand: OK, done

mikand:
Update.

* I managed to resolve the problem with AppVeyor. The issue was in the "run_with_env.ps1" script that was forcing a wrong version of Visual Studio for Python 3.6. Since the new version of setuptools is clever enough, I could simply remove this script.

* I moved the patched setup.py in the `solver_patches` repo

* I updated the version of MathSAT, but there are still some issues under Windows (need to patch the C wrapper and crash on `test_hr_parsing.py`)

mikand:
@agriggio Updated MathSAT to 5.5.1 and now the tests are passing!

@marcogario For me the PR is ready, can you have a look and give feedback?

mikand:
@marcogario The tests on travis are failing due to #465 that has nothing to do with this PR. Do you think we can merge this?

PR #454, opened 2017-11-02
Fixed indexing and simple path condition in MC example

cristian-mattarei:
The MC example was not implementing correctly: the simple path condition imposed a constraint over the entire assignments instead of each state of the system, and the indexing for unrolling and k hypothesis were out of 1

### DIFF
--- a/examples/model_checking.py
+++ b/examples/model_checking.py
@@ -10,9 +10,10 @@
 #
 from six.moves import xrange
 
-from pysmt.shortcuts import Symbol, Not, Equals, And, Times, Int, Plus, LE
+from pysmt.shortcuts import Symbol, Not, Equals, And, Times, Int, Plus, LE, Or, Iff
 from pysmt.shortcuts import is_sat, is_unsat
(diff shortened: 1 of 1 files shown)


examples/model_checking.py
```diff
@@ -62,19 +62,23 @@ def get_simple_path(system, k):
...
         subs_i = get_subs(system, i)
-        for j in xrange(i+1, k):
+        for j in xrange(i+1, k+1):
+            state = []
             subs_j = get_subs(system, j)
             for v in system.variables:
                 v_i = v.substitute(subs_i)
                 v_j = v.substitute(subs_j)
-                res.append(Not(Equals(v_i, v_j)))
+                cond = Not(Iff(v_i, v_j)) if v_i.get_type() == BOOL else \
```
mikand: You can use the function `EqualsOrIff` (see https://github.com/pysmt/pysmt/blob/master/pysmt/shortcuts.py#L361):
```python
Not(EqualsOrIff(v_i, v_j)) 
```

PR #456, opened 2017-11-14
Extended MC example with PDR

cristian-mattarei:


examples/model_checking.py
```diff
@@ -1,19 +1,20 @@
 # This example shows a more advance use of pySMT.
 #
 # It provides a simple implementation of Bounded Model Checking [1]
-# and K-Induction [2] and applies it on a simple infinite-state
+# with K-Induction [2], and PDR [3], and applies it on a simple 
 # transition system.
 #
 # [1] ...
```
marcogario: We might as well fix these :)

examples/model_checking.py
```diff
@@ -35,115 +36,205 @@ def __init__(self, variables, init, trans):
...
-            print("--> The system is safe!")
-            return
-
-
-def main():
-    # Example Transition System (SMV-like syntax)
+class PDR(object):
+    def __init__(self, system):
+        self.system = system
+        self.R = [system.init]
```
marcogario: This should be lower case. Uppercase identifiers are typically used for classes.

examples/model_checking.py
```diff
@@ -35,115 +36,205 @@ def __init__(self, variables, init, trans):
...
+        self.solver = Solver()
+        
+    def check_property(self, prop):
+        """Property Directed Reachability approach without optimizations."""
+        print("Checking property %s..." % prop)
+
+        while True:
+            cube = self.get_bad_state(prop)
+            if cube is not None:
```
marcogario: The nesting of the conditionals here is non-intuitive.
A high-level description of the idea is needed to clarify why this works.

examples/model_checking.py
```diff
@@ -35,115 +36,205 @@ def __init__(self, variables, init, trans):
...
+    def get_bad_state(self, prop):
+        """Extracts a reachable state that intersects the negation 
+        of the property and the last current frame"""
+        return self.solve(And(self.R[-1], Not(prop)))
+
+    def solve(self, formula):
+        """Provides a satisfiable assignment to the state variables that are consistent with the input formula"""
+        self.solver.reset_assertions()
+        self.solver.add_assertion(formula)
+        if self.solver.solve():
```
marcogario: You should be able to simplify the above by doing ```self.solver.solve(formula)```. This will solve under the assumption of formula.

examples/model_checking.py
```diff
@@ -35,115 +36,205 @@ def __init__(self, variables, init, trans):
...
+            return And([EqualsOrIff(v, self.solver.get_value(v)) for v in self.system.variables])
+        return None
+
+    def recursive_block(self, cube):
+        """Blocks the cube at each frame, if possible."""
+        for i in range(len(self.R)-1, 0, -1):
+            cubeprime = cube.substitute(dict([(v, next_var(v)) for v in self.system.variables]))
```
marcogario: You could create the map ```curr_to_next``` in the constructor. No need to recreate it over and over.

examples/model_checking.py
```diff
@@ -35,115 +36,205 @@ def __init__(self, variables, init, trans):
...
+                for j in range(1, i+1):
+                    self.R[j] = And(self.R[j], Not(cube))
+                return False
+            cube = cubepre
+        return True
+    
+    def inductive(self, prop):
+        """Checks if last two frames are equivalent """
+        if len(self.R) > 1 and \
+           self.solve(Not(EqualsOrIff(self.R[-1], self.R[-2]))) is None:
```
marcogario: You could use: ```self.solver.is_valid```.
self.R[-1] and self.R[-2] are boolean expressions. You can use Iff directly.

examples/model_checking.py
```diff
@@ -35,115 +36,205 @@ def __init__(self, variables, init, trans):
...
+
+            f = self.get_k_induction(prop, b)
+            print("   [K-IND]  Checking bound %d..." % (b+1))
+            if is_unsat(f):
+                print("--> The system is safe!")
+                return
+
+
+def counter(bit_count):
+    """Counter example with n bits and reset signal."""
```
marcogario: This example is pre-BV. It makes sense to convert it to BVs. Both approaches should still work.

marcogario:
I changed the example to use BV. The previous version was too complicated, and the first specification did not make sense. I think the rewriting using BVs is much cleaner.
I also removed ```prop``` from the arguments of ```inductive```, since it is not used in the implementation. 

Please add a high-level description of the main loop, and rename ```self.R``` into something else (frames?).

marcogario:
@cristian-mattarei I changed the structure of the IF statements so that every case has a print associated describing what is going on. I am a bit surprised that the blocking case always tries to block the same cube. Is this the expected behavior? Is this because there is no generalization in the blocking? 

If you can have a look and let me know if this is ok, I think this is ready to be merged.
Thanks!

PR #460, opened 2017-11-19
QE: Self-substitution

marcogario:
Implementation of Self-Substitution as described in 
Dror Fried, Lucas M. Tabajara, and Moshe Y. Vardi,, "BDD-Based Boolean Functional Synthesis", CAV 2016


marcogario:
Tests are failing due to mathsat qe. It seems that this additional QE method is triggering the execution of some tests that were typically not executed. Need to investigate further to understand if it is a bug or feature :)

PR #467, opened 2017-12-26
Infix: Add BV textual infix notation

marcogario:
Unsigned operators used to have only the infix notation with operators
(eg., bv1 & bv2), creating an asymmetry between infix operators that
are expressible with the textual notation and not.

All BV operators now have the textual infix notation (eg.,
bv1.BVAnd(bv2)). In addtion, unsigned operators maintain the usual
infix notation.


marcogario:
Included in #468

PR #469, opened 2017-12-26
Updated Travis CI to work on new images

mikand:
Travis recently upgraded their images (https://blog.travis-ci.com/2017-12-12-new-trusty-images-q4-launch) and the builds were failing. This PR fixes the config to make CI work again.

This also includes #470 that fixes an issue accidentally committed to master because of CI being broken


pysmt/cmd/installers/bdd.py
```diff
@@ -44,7 +44,7 @@ def compile(self):
             makefile = "Makefile_64bit"
 
         # Build the pycudd
-        command = ['python%s-config' % self.python_version, '--includes']
+        command = ['python-config', '--includes']
```
marcogario: This is not correct when multiple versions of python are installed (e.g., 3.5 and 3.6). python-config maps to the default interpreter (the one you would get when you call python), while python%s-config takes into account the version of python used by the interpreter calling the installer.

.travis.yml
```diff
@@ -38,7 +38,7 @@ addons:
     - libmpc-dev
 
 python:
-  - 3.5
```
marcogario: Ok. This is now aligned with what we do in appveyor.

pysmt/cmd/installers/bdd.py
mikand: Indeed, but since `python3.6-config` is not installed on travis I see 2 strategies:
1) Try to use `python{version}-config` and if it fails try with `python-config`
2) Avoid `python-config` and rely on the `sysconfig` module as it is done by the btor `configure.sh` script: 
```
  pyinc=`$PYTHON -c "import sysconfig; print(sysconfig.get_config_var('CONFINCLUDEPY'))"`
  pylib=`$PYTHON -c "import sysconfig; print(sysconfig.get_config_var('BINLIBDEST'))"`
```

marcogario: Is this a Travis issue or an Ubuntu issue?

I would go with solution 1. printing a warning message when falling back to python-config.

Solution 2. does not work within a virtual environment. When the virtual environment is constructed, the interpreter is copied in the virtual env. sysconfig uses the information from the interpreter and is not aware that it is in a virtual env. python-config lives within the virtualenv, and therefore can provide the correct path. 
You can check this by comparing the output of ``python-config --includes`` and ``python -m sysconfig | grep CONFINC`` from within a virtual environment.

mikand: On my Ubuntu machine I do have the `python{version}-config`, so I guess it is either a Travis issue or a virtual-env issue...

mikand: Let's go for solution 1 then

mikand:
@marcogario Do you have time to have a look? Since master tests are failing, both this PR and #470 need to be merged.

marcogario:
It seems that Travis still fails. I guess this is because of the other pr. Can you open a new pr that includes both changes, so that we are sure that the final result works?

mikand:
OK, I will merge #470 here and close that PR

mikand:
@marcogario You can check and merge this

PR #470, opened 2018-01-02
Fixed typing error in msat back

mikand:
The recently added theory combination example raised a typing error within the back function of the msat converter. The error was due to INT/REAL distinction made by pysmt that is ignored by mathsat for numeric constants. The fix consists in asking the type system of mathsat the relevant types instead of determining them using the pysmt type-checker.

NOTE: this bug was highlighted by the new "theory combination" example added in #451 but the PR was accepted beacuse Travis is currently broken. We need to finish with #469 to trust CI again.


mikand:
This PR has been merged in #469 to have a single PR taht fixes the issues on master

PR #476, opened 2018-03-14
floordiv and mod

FedericoAureliano:
I needed floordiv and mod for one of my projects, so I implemented them in the most naive way. If the code is useful to you, take it! Feel free to make any changes. 

Z3 supports both.

I'm also curious to find out what the right way to add operators like this is. I tried to follow these docs but not sure I was successful. Feedback would be cool.

http://pysmt.readthedocs.io/en/latest/development.html#how-to-add-a-new-theory-within-pysmt

### DIFF
--- a/pysmt/fnode.py
+++ b/pysmt/fnode.py
@@ -816,6 +816,12 @@ class FNode(object):
     def __div__(self, right):
         return self._apply_infix(right, _mgr().Div, _mgr().BVUDiv)
 
+    def __floordiv__(self, right):
+        return self._apply_infix(right, _mgr().FloorDiv, None)
+
+    def __mod__(self, right):
+        return self._apply_infix(right, _mgr().Mod, _mgr().BVURem)
+
     def __truediv__(self, right):
         return self.__div__(right)
 
@@ -880,9 +886,6 @@ class FNode(object):
 
     def __rshift__(self, right):
         return self._apply_infix(right, None, bv_function=_mgr().BVLShr)
-
-    def __mod__(self, right):
-        return self._apply_infix(right, None, bv_function=_mgr().BVURem)
 
     @assert_infix_enabled
     def __call__(self, *args):
--- a/pysmt/oracles.py
+++ b/pysmt/oracles.py
@@ -319,6 +319,37 @@ class TheoryOracle(walkers.DagWalker):
         theory_out = theory_out.set_difference_logic(False)
         return theory_out
 
+    def walk_floordiv(self, formula, args, **kwargs):
+        """Extends the Theory with Non-Linear, if needed."""
+        assert len(args) == 2
+        theory_out = args[0]
+        for t in args[1:]:
+            theory_out = theory_out.combine(t)
+        # Check for non-linear
+        left, right = formula.args()
+        if len(left.get_free_variables()) != 0 and \
+           len(right.get_free_variables()) != 0:
+            theory_out = theory_out.set_linear(False)
+        # This is  not in DL anymore
+        theory_out = theory_out.set_difference_logic(False)
+        return theory_out
+
+
+    def walk_mod(self, formula, args, **kwargs):
+        """Extends the Theory with Non-Linear, if needed."""
+        assert len(args) == 2
+        theory_out = args[0]
+        for t in args[1:]:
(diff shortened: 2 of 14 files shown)


pysmt/oracles.py
```diff
@@ -319,6 +319,37 @@ def walk_div(self, formula, args, **kwargs):
         theory_out = theory_out.set_difference_logic(False)
         return theory_out
 
+    def walk_floordiv(self, formula, args, **kwargs):
```
mikand: Technically, floordiv is non part of LIA even when the divisor is an integer constant. Consider for example `x // 5 == y` where x and y are integers, this is *not* equivalent to `x == y * 5` because `x=11` and `y=2` is a model of the first equation but not of the other. Moreover both floordiv and modulus are not part of SMTLib, but Z3-specific extensions. 

I think it would make sense to either mark the use of these operators as a NIA formula (and Z3 is the only solver that can currently deal with it) or create a special logic for LIA + floordiv + mod. 

Alternatively, for the 'quasi-linear' case, one could define the floordiv in terms of modulus:
```
(define-fun floordiv ((x Int) (y Int)) Int
  (ite (or (= (mod x y) 0) (>= x 0))
       (div x y)
       (ite (>= y 0)
            (+ (div x y) 1)
            (- (div x y) 1))))
```
This definition should be supported also by MathSAT via the modular congruence.

@marcogario @agriggio Any thoughts? What would be a clean way of including this logic for solvers that supports it?

pysmt/simplifier.py
```diff
@@ -783,6 +783,48 @@ def walk_div(self, formula, args, **kwargs):
...
 
+    def walk_floordiv(self, formula, args, **kwargs):
+        sl = args[0]
+        sr = args[1]
+
+        if sl.is_constant() and sr.is_constant():
+            l = sl.constant_value()
+            r = sr.constant_value()
+            assert sl.is_int_constant()
+            return self.manager.Int(l / r)
```
mikand: In python3 this should be `return self.manager.Int(l // r)`

pysmt/fnode.py
```diff
@@ -816,6 +816,12 @@ def __rmul__(self, right):
     def __div__(self, right):
         return self._apply_infix(right, _mgr().Div, _mgr().BVUDiv)
 
+    def __floordiv__(self, right):
+        return self._apply_infix(right, _mgr().FloorDiv, None)
```
marcogario: The none for BV is going to raise a very ugly exception. _apply_infix expects two functions, one for the Rational/Integer case and one for the Bitvector. If this does not apply to BV, we should raise a proper exception.

pysmt/oracles.py
marcogario: For what concerns logic extensions outside the SMT-LIB, we already have some precedent in the theory of array with constant element. In that case we created a custom bit in the Theory description, and thus defined logics with this extended concept. I would follow the same approach here.

marcogario:
@FedericoAureliano thanks for contributing!

@mikand I am not clear on the semantics of these operators:

- How is floordiv different from div in the integer case? If they do not differ, then we should not introduce a new operator, but extend div.
- Can we rewrite floordiv as a combination of div and floor for the rational case? If so, floordiv would become an alias for the regular div (in the integer case) or for a rewriting for the rational case.

- Is modulus with a rational on the rhs defined? Does Mathsat support it? If we only want modulus for the integers, we should enforce this in the constructor.

marcogario:
@mikand can we try to take a go/no-go decision about this? My main concern is on the semantics of the operator. Any pointer on this might help.

yoni206:
Excuse me for popping out of nowhere here, but I would just like to say that I would love to have integer modulo operator in pysmt! If mod is simpler than floor, maybe it would be useful to split this PR into two?
Anyway, if I can help to make this happen, please let me know!

marcogario:
The main concern with this is the semantics and general applicability across solvers, since there is no smtlib equivalent for this. If you want to move this forward, please make a list/table of which solver of pysmt supports these operators (you can distinguish between the two operators if necessary). If it is only z3, i would be reasonably against it. 3 solvers would be ideal.

Then we would like to have some meaningful examples that we can run through the solvers and make sure that we get the same semantics.

PR #478, opened 2018-05-14
Add support  for regular expression in smt-lib format

Phat3:
Added support for regular expression following the [SMT language partial specification](https://rise4fun.com/z3/tutorialcontent/sequences#h23).

With this pull request, the library is NOW able to correctly parse and outputs rules containing regex directive.


marcogario:
From the errors from appveyor, it seems that this patch is not python compatible, but only python 2.

pysmt/type_checker.py
```diff
@@ -97,11 +97,26 @@ def walk_str_to_str(self, formula, args, **kwargs):
         #pylint: disable=unused-argument
         return self.walk_type_to_type(formula, args, STRING, STRING)
 
+    @walkers.handles(op.RE_CONCAT, op.RE_KLEENE_PLUS, op.RE_KLEENE_STAR)
```
marcogario: This should be a walk type_to_type REGEX. See comment about types.

pysmt/typing.py
```diff
@@ -204,6 +209,30 @@ def is_array_type(self):
...
 
 
+class _RegExType(PySMTType):
+    """Internal class used to represent an Regex type.
+
+    This class should not be instantiated directly, but the factory
+    method RegExType should be used instead.
+    """
+    def __init__(self, regex_type):
+        decl = _TypeDecl("RegEx", 1)
```
marcogario: The definition of the RegLan sort has arity 0, not 1:  RegLan() and RegEx(String) are the same thing. This simplifies the handling of the type in this module, and it can be treated as other simple types like INT or REAL.

Phat3:
Perfect thanks for the review! :smile: 

I'll look into it in the next couple of days

pysmt/smtlib/printers.py
```diff
@@ -483,7 +544,7 @@ def walk_bool_constant(self, formula, **kwargs):
             return "false"
 
     def walk_str_constant(self, formula, **kwargs):
-        return '"' + formula.constant_value() + '"'
+        return '"' + formula.constant_value().replace('"', '""') + '"'
```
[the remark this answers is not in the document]
Lukas-Dresel: The problem with the old approach is that the constant value did not have the quotation marks escaped at all. E.g.
```
In [1]: from pysmt.shortcuts import String

In [2]: String('asdf"')
Out[2]: "asdf""

In [3]: String('asdf"').to_smtlib()
Out[3]: '"asdf""'
```

The pysmt specification says any double-quotes in the output must be escaped with `""`, see
``` :notes
  "SMT-LIB 2.6 has one escape sequence of its own for string literals. Two
   double quotes are used to represent the double-quote character within 
   a string literal (like the one containing this very note). That escape 
   sequence is at the level of the SMT-LIB frontend of a solver, not at the 
   level of this theory. 
  "```
as in the link posted above.

This escaping was not done before, which is what this change addresses.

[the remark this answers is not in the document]
Comment (Lukas-Dresel) [reply pysmt/smtlib/printers.py]: This was originally done before the regex PR, i guess it snuck in. If you want, we can pull it out and move it into a seperate PR.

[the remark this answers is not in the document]
Comment (marcogario) [reply pysmt/smtlib/printers.py]: Yes, please. Separate it into its own PR together with a regression test that does exactly what you show in the example above. This will probably be merged faster than the rest of the PR. Thanks.

**Comment by Phat3** [comment]
I fixed some of the issues indicated but I still need to implement the following:

- Implement abbreviations for **ReKleenePlus** and **ReOpt**
- Implement Simplification methods
- Implement the suggested tests
- Figure out which solver supports this theory

File: pysmt/smtlib/printers.py [hunk 4]
```diff
@@ -247,6 +247,67 @@ def walk_str_to_int(self,formula, **kwargs):
         self.walk(formula.arg(0))
         self.write(")")
 
+    def walk_str_to_re(self,formula, **kwargs):
```
marcogario: All of these should be implemented as a call to walk_binary or walk_nary. See examples above.

pysmt/type_checker.py
```diff
@@ -286,6 +297,28 @@ def walk_str_indexof(self, formula, args, **kwargs):
...
         return None
 
+    def walk_str_in_re(self, formula, args, **kwargs):
+        assert formula is not None
+        if len(args) == 2 and args[0].is_string_type() and \
+           args[1].is_regex_type():
+            return BOOL
+        return None
+
+    def walk_re_all(self, formula, args, **kwargs):
```
marcogario: unify these 3 cases with a walk_regex and a `@handles` property.

marcogario:
@Phat3 Do you have any SMTLIB file that uses this feature? If so, we could include them as tests in pySMT for this feature.

Issue #483, opened 2018-05-20
MathSAT / OSX: Hardcoded library path in 5.5.1


marcogario:
Incorrect dylib object in _mathsat.so on OSX.

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "mathsat.py", line 21, in <module>
    _mathsat = swig_import_helper()
  File "mathsat.py", line 20, in swig_import_helper
    return importlib.import_module('_mathsat')
  File "/usr/local/Cellar/python@2/2.7.15/Frameworks/Python.framework/Versions/2.7/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
ImportError: dlopen(./_mathsat.so, 2): Library not loaded: /Users/griggio/Documents/src/mathsat_release/build/libmathsat.dylib
  Referenced from: /Users/travis/python_bindings/msat/_mathsat.so
  Reason: image not found
```

We either need to change the installer to rewrite this (we already had this issue with CVC4 on mac, if I recall correctly), or this needs to be fixed upstream.

marcogario:
@agriggio Can you take a look?

This is how I plan to solve this: https://stackoverflow.com/questions/40325858/how-to-change-rpath-on-os-x

marcogario:
This is solved in our installer in PR #487 . However, this should be fixed upstream.

PR #484, opened 2018-05-20
Release/0.7.5

marcogario:
Release branch for 0.7.5. 

This is currently blocked by #483.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -97,7 +97,7 @@ Supported Theories and Solvers
 
 pySMT provides methods to define a formula in Linear Real Arithmetic
 (LRA), Real Difference Logic (RDL), Equalities and Uninterpreted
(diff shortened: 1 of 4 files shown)


varunpatro:
Can you also consider adding my PR #486 to this release?

My colleagues and I use mac for our work, and found that QF_LIA is best solved by Yices. For which I've now added support.

marcogario:
I would prefer to complete 0.7.5 with the items that were in scope, since there seems to be some backwards compatibility issue with OSX in #486. 
 
Do you need to pip install pysmt? Is this preferable than cloning the git repo? If so, we can do a dev release with support for yices and osx, once we address this issue.
I would also like to evaluate whether the state of their python bindings is mature enough as to replace yicespy (our own wrapper) [1].

[1] https://github.com/SRI-CSL/yices2/tree/master/src/bindings/python

Issue #412, opened 2017-06-04
Installation of yices on Mac os fails


smover:
The installation of yices on Mac os fails, since pymst dowloads the archive for linux (yices-2.5.1-x86_64-unknown-linux-gnu-static-gmp.tar.gz)

```
$> pysmt-install --yices
This script allows you to install the solvers supported by pySMT.

By executing this script, you confirm that you have read and agreed
with the licenses of each solver.

Notice: the installation process might require building tools
        (e.g., make and gcc).

Continue? [Y]es/[N]o: Y
Downloading: /Users/smover/.smt_solvers/yices/yices-2.5.1-x86_64-unknown-linux-gnu-static-gmp.tar.gz Bytes: 5791477
100.0%    
./install-yices: line 99: break: only meaningful in a `for', `while', or `until' loop
./install-yices: line 135: break: only meaningful in a `for', `while', or `until' loop

Installation failed

Can't find /sbin/ldconfig


running build_ext
building '_yicespy' extension
swigging yices_python.i to yices_python_wrap.c
swig -python -I/Users/smover/.smt_solvers/python-bindings-2.7/yices_bin/include -o yices_python_wrap.c yices_python.i
yices_python.i:216: Error: Unable to find 'yices_types.h'
yices_python.i:217: Error: Unable to find 'yices.h'
error: command 'swig' failed with exit status 1
Traceback (most recent call last):
  File "/usr/local/bin/pysmt-install", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python2.7/site-packages/pysmt/cmd/install.py", line 240, in main
    installer.install(force_redo=options.force_redo)
  File "/usr/local/lib/python2.7/site-packages/pysmt/cmd/installers/base.py", line 143, in install
    self.compile()
  File "/usr/local/lib/python2.7/site-packages/pysmt/cmd/installers/yices.py", line 68, in compile
    self.install_yicespy()
  File "/usr/local/lib/python2.7/site-packages/pysmt/cmd/installers/yices.py", line 58, in install_yicespy
    wheel_file = glob.glob(os.path.join(self.base_dir, "yicespy") + "*.whl")[0]
IndexError: list index out of range
```

mikand:
@smover Thanks for the report. Indeed we never really tested pysmt-install under OSX.

By any chance do you have some time to work on this issue? (Of course, we can provide support and guidance)

marcogario:
[...]

Issue #96, opened 2015-05-07
Pysmt install for osx and windows


marcogario:
- verify that installer works on other playforms, thus downloading yhe correct versiom of the tools

- setup ci for those platforms

PR #486, opened 2018-05-25
Upgrade Yices to 2.54 and add macOS support

varunpatro:


varunpatro:
This PR partially solves #96 and fully solves #461 and #412.

marcogario:
Thanks for the patch!
Can you confirm that the tests are passing on osx?

I get the following failure [1] from our CI both for python and python 2.

```
pysmt.test.smtlib.test_parser_qf_lia.test_generator('pysmt/test/smtlib/small_set/QF_LIA/prp-23-47.smt2.bz2', QF_LIA, False) ... /Users/travis/.travis/job_stages: line 57:  5359 Illegal instruction: 4  ${PYTHON_EXE} -mnose pysmt -v
```

[1] https://travis-ci.org/pysmt/pysmt/jobs/384114232

marcogario:
https://stackoverflow.com/questions/14268887/what-is-the-illegal-instruction-4-error-and-why-does-mmacosx-version-min-10

varunpatro:
That's really strange. Because I'm running 10.13.4 and it builds and runs fine on my computer. Let me test this a little more.

In the worst case, if the yices2 binary only works on some os versions, is there a way to halt the installation and alert this to the user?

marcogario:
The issue might be also be in the python wrapper. In particular, if the version of xcode is different than the version used for building yices. If this is the case, it should be possible to add a backwards compatibility flag when building yicespy.

If there is an issue with Yices, we should report it upstream and see how to proceed. We typically do build from source if there are these types of issues. The problem is that in those cases it takes quite a bit to install some solvers.

mikand:
@marcogario I think that this PR improves the status of the Yices solver anyway. Maybe we can merge this and then deal with the problem of making yices work under OSX on our CI separately (e.g. in #491). What do you think?

marcogario:
Ok to only include linux upgrade. Please keep a pr open for the macos support. We can maybe ask other macos users for ideas on how to debug this.

mikand:
OK, I'll merge this then I will try to see if I manage to make yices to work in #491

mikand:
BTW, I just found this: http://yices.csl.sri.com/faq.html  :shamrock: 
It seems that the Travis VM does not have the needed CPU flag enabled...

PR #489, opened 2018-05-31
Fix/doublequote escaping

Lukas-Dresel:
As discussed in PR #478 this is the fix for the double-quote parsing and printing of smt string constants.
Regression tests have been added that should currently fail on master but succeed using these changes.

While writing the testcases I noticed other places that had the issue, hopefully this should fix all of them!

### DIFF
--- a/pysmt/smtlib/parser/parser.py
+++ b/pysmt/smtlib/parser/parser.py
@@ -637,7 +637,10 @@ class SmtLibParser(object):
                 res = mgr.BV(value, width)
             elif token[0] == '"':
                 # String constant
-                res = mgr.String(token.replace('"',''))
+                val = token[1:-1]
+                val = val.replace('""', '"')
+                val = val.decode('string-escape')
+                res = mgr.String(val)
             else:
                 # it could be a number or a string
                 try:
--- a/pysmt/smtlib/parser/parser.py
+++ b/pysmt/smtlib/parser/parser.py
@@ -239,23 +239,22 @@ class Tokenizer(object):
 
                     elif c == "\"":
                         # String literals
-                        s = []
-                        c = next(reader)
-                        while c:
+                        s = c
+                        num_quotes = 0
+                        while True:
+                            c = next(reader)
+                            if not c:
+                                raise PysmtSyntaxError("Expected '\"'",
+                                                       reader.pos_info)
+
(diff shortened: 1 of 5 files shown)


pysmt/smtlib/parser/parser.py
```diff
@@ -637,7 +636,10 @@ def atom(self, token, mgr):
                 res = mgr.BV(value, width)
             elif token[0] == '"':
                 # String constant
-                res = mgr.String(token.replace('"',''))
+                val = token[1:-1]
+                val = val.replace('""', '"')
+                val = val.decode('string-escape')
```
Lukas-Dresel: This I'm not entirely sure about, this all depends on whether or not support for '\xFF' style escapes, etc. is desirable. 
Please let me know!

pysmt/smtlib/script.py
```diff
@@ -86,7 +86,7 @@ def serialize(self, outstream=None, printer=None, daggify=True):
 
         elif self.name in [smtcmd.CHECK_SAT, smtcmd.EXIT,
                            smtcmd.RESET_ASSERTIONS, smtcmd.GET_UNSAT_CORE,
-                           smtcmd.GET_ASSIGNMENT]:
+                           smtcmd.GET_ASSIGNMENT, smtcmd.GET_MODEL]:
```
Lukas-Dresel: This was lying around, hope you don't mind me sneaking it into this PR. If you do, let me know and I will remove it.

pysmt/smtlib/parser/parser.py
mikand: This `decode()` is not availabe in python3 (see: https://stackoverflow.com/questions/14820429/how-do-i-decodestring-escape-in-python3?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa) and makes the tests fail on our CI: https://travis-ci.org/pysmt/pysmt/builds/385986747?utm_source=github_status&utm_medium=notification

pysmt/smtlib/script.py
mikand: :+1:

Lukas-Dresel:
Okay, for now I took out the string unescaping, the type juggling differences between `bytes`, `unicode` and `str` between python2 and python3 is very annoying. I will look into how to solve that correctly but for now this should hopefully make the tests pass.

PR #491, opened 2018-05-31
CI with osx

mikand:
This PR uses a new feature of Travis that allows the mixture of linux and osx builds. For now, on osx we test MathSAT and Z3 only, in the future we can add the other solvers.

Note: I configured the CI so that osx builds are run ony on PRs and on master

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,250 @@
 language: python
-dist: trusty
+
+matrix:
+  include:
+  - os: linux
+    python: 3.6
+    env: PYSMT_SOLVER="all"       PYSMT_GMPY="TRUE"
+  - os: linux
+    python: 3.6
+    env: PYSMT_SOLVER="all"       PYSMT_GMPY="FALSE"
+  - os: linux
+    python: 3.6
+    env: PYSMT_SOLVER="None"      PYSMT_GMPY="FALSE"   PYSMT_CYTHON="TRUE"
+  - os: linux
+    python: 3.6
+    env: PYSMT_SOLVER="None"      PYSMT_GMPY="FALSE"   PYSMT_CYTHON="FALSE"
+  - os: linux
+    python: 3.6
+    env: PYSMT_SOLVER="msat"      PYSMT_GMPY="FALSE"
+  - os: linux
+    python: 3.6
+    env: PYSMT_SOLVER="z3"        PYSMT_GMPY="FALSE"
+  - os: linux
+    python: 3.6
+    env: PYSMT_SOLVER="cvc4"      PYSMT_GMPY="FALSE"
+  - os: linux
+    python: 3.6
+    env: PYSMT_SOLVER="yices"     PYSMT_GMPY="FALSE"
+  - os: linux
+    python: 3.6
+    env: PYSMT_SOLVER="bdd"       PYSMT_GMPY="FALSE"
+  - os: linux
+    python: 3.6
+    env: PYSMT_SOLVER="picosat"   PYSMT_GMPY="FALSE"
+  - os: linux
+    python: 3.6
+    env: PYSMT_SOLVER="btor"      PYSMT_GMPY="FALSE"
+  - os: linux
+    python: 3.6
+    env: PYSMT_SOLVER="msat_wrap" PYSMT_GMPY="FALSE"
+  - os: linux
+    python: 3.6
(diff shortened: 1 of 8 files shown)


mikand:
@marcogario what do you think? I think this solution is way better than the separate branch for MacOSX tests...

pysmt/cmd/install.py
```diff
@@ -229,10 +229,14 @@ def main():
             else:
                 print("set PYTHONPATH=" + bindings_dir + ";%PYTHONPATH%")
                 print("set PATH=" + bindings_dir + ";%PATH%")
+        elif platform.system().lower() == "darwin":
+            print("export PYTHONPATH=\"" + bindings_dir + ":${PYTHONPATH}\"")
+            print("export PATH=\"" + bindings_dir + ":${PATH}\"")
```
marcogario: why is PATH needed here?

.travis.yml
```diff
@@ -37,68 +272,34 @@ addons:
...
-      env: PYSMT_SOLVER="bdd"   PYSMT_GMPY="FALSE"
-    - python: pypy
-      env: PYSMT_SOLVER="picosat" PYSMT_GMPY="FALSE"
-    - python: pypy
-      env: PYSMT_SOLVER="btor"  PYSMT_GMPY="FALSE"
-    - python: pypy
-      env: PYSMT_SOLVER="msat_wrap" PYSMT_GMPY="FALSE"
-    - python: pypy
-      env: PYSMT_SOLVER="z3_wrap"   PYSMT_GMPY="FALSE"
+before_install: |
```
marcogario: This should go into ci/travis_before_install.sh

.travis.yml
```diff
@@ -1,5 +1,239 @@
...
+  #   language: generic
+  #   python: 3.6
+  #   env: PYSMT_SOLVER="all"       PYSMT_GMPY="TRUE"   TRAVIS_PYTHON_VERSION="3.6.5"
+  # - os: osx
+  #   osx_image: xcode9.3
+  #   language: generic
+  #   python: 3.6
+  #   env: PYSMT_SOLVER="all"       PYSMT_GMPY="FALSE"   TRAVIS_PYTHON_VERSION="3.6.5"
+  - os: osx
+    osx_image: xcode9.3
```
marcogario: I am not sure if we want to use xcode9.3 or xcode8.3. xcode8.3 runs on OSX 10.12, so that we have a bit of backwards compatibility since 10.13 has been released less than one year ago.

.travis.yml
```diff
@@ -1,5 +1,239 @@
...
+  - os: linux
+    python: pypy
+    env: PYSMT_SOLVER="None"      PYSMT_GMPY="FALSE"   PYSMT_CYTHON="FALSE"
+
+
+  # - os: osx
+  #   osx_image: xcode9.3
+  #   language: generic
+  #   python: 3.6
+  #   env: PYSMT_SOLVER="all"       PYSMT_GMPY="TRUE"   TRAVIS_PYTHON_VERSION="3.6.5"
```
marcogario: Ah, it would actually be nice to include testing for GMPY and CYTHON. I think PYSMT_SOLVER can take a list "msat;z3" to simulate the "all" value.

pysmt/cmd/install.py
mikand: For Z3, it seems that on OSX you need to have the dylib files in the PATH instead of LD_LIBRARY_PATH....

.travis.yml
mikand: :+1:

.travis.yml
mikand: I can try. For sure on the default version used by travis the tests did not work, I will try to find the lower possible config.

.travis.yml
mikand: :+1:

mikand:
@marcogario This is ready for your review.

* I managed to implement a compiler-based installer for yices that is enabled if the system detects that AVX2 instructions are not present (see http://yices.csl.sri.com/faq.html).
* The solvers supported under OSX are now yices, z3 and msat. Z3 required to use the xcode9.3 image because on older images the download itself of Z3 fails for some reason that is not easy to debug. All other configurations use xcode9 that is an OS X 10.12 image.
* Also gmpy is supported

Issue #465, opened 2017-12-12
Non-deterministic segfault of tests with Z3 in Travis


mikand:
I noted that the builds having the z3 solver can non-deterministically yield a segfault on Travis.

I tried to reproduce the problem locally even using the docker Travis image with no luck.

We could try two different strategies here:
- Update Z3 to a newer nightly build and see if the problem goes away
- Try to update the libstdc++ of the travis image (it is more a feeling than a suspect, but in my local experiment with the `quay.io/travisci/travis-python` docker image I had to update this to make Z3 work at all)

mikand:
I tried to update libstdc++ and it does not work: https://travis-ci.org/pysmt/pysmt/jobs/315236750

marcogario:
The problem seems to occur only with python 3.5, and not python 2.7 . 
The segfault might also be due to memory limits on the instance that is running the tests. Did you manage to verify how much memory is consumed during these tests?

I disabled caching in the CI and re-run the tests. Maybe that will help.

mikand:
Finger crossed, but after we switched to python 3.6 the issue never occurred

mikand:
Unfortunately, the bug is still present in #474 ....

marcogario:
See #530.

mikand:
Unfortunatey, this issue is still open and I also managed to reproduce it outside Travis: https://dev.azure.com/micheliandrea/PySMT/_build/results?buildId=17

mikand:
I think I finally isolated the issue!
On #564 I disabled the test `test_tactics_z3` in `test_solving.py` and since then I got no more segfault errors. This smells of victory because the test actually accesses the internal Z3 nodes and applies a custom tactic, so it may very well mess up the ref/deref system.

I still have no clear explanation/fix of the problem, but I'll have a look asap.

mikand:
Unfortunately, this is part of the issue, but not the whole one. The following build crashed on Z3 with OSX even with the `test_tactics_z3` test disabled.

https://dev.azure.com/micheliandrea/PySMT/_build/results?buildId=65

mikand:
Hopefully the problem should now be solved, as #564 has been merged in master. Will re-open the issue if the problem persists.

marcogario:
Was there one or two tests with tactics that were disabled? I changed only one.

mikand:
I restored and fixed both before you accepted the PR (see #564 )

mikand:
[...]

PR #494, opened 2018-06-01
Simplify shared lib usage

randomir:
Modify `z3` and `msat` installers in order to make their shared binary objects (libraries/dlls) auto-discoverable, without the need for setting `LD_LIBRARY_PATH`/`PATH`.

Tested on Linux, but not on Windows.

### DIFF
--- a/pysmt/cmd/install.py
+++ b/pysmt/cmd/install.py
@@ -185,12 +185,12 @@ def main():
         mirror_url += "/{archive_name}"
 
     # This should work on any platform
-    install_dir= os.path.expanduser(options.install_path)
+    install_dir = os.path.expanduser(options.install_path)
     if not os.path.exists(install_dir):
         os.mkdir(install_dir)
 
     # This should work on any platform
-    bindings_dir= os.path.expanduser(options.bindings_path)
+    bindings_dir = os.path.expanduser(options.bindings_path)
     if not os.path.exists(bindings_dir):
         os.mkdir(bindings_dir)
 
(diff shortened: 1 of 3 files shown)


randomir:
Please note the test failure on Linux/Python3.6 for PYSMT_SOLVER="all" PYSMT_GMPY="FALSE" existed before this PR.

mikand:
@randomir Wow! Thanks a lot for this patch! Indeed it was quite annoying to set the PATH/LD_LIBRARY_PATH and I was not aware of the -R $ORIGIN option.

I'll merge this as soon as the tests will pass (Due to #465 we sometimes have to repeat the tests)

randomir:
@mikand, I'm glad to contribute!

And while we're on the topic of annoying -- setting of `PYTHON_PATH` can be also avoided if you install `pysmt`/solvers inside `site-packages` (current environment's base dir for packages).

For example, we're doing that in this install wrapper https://github.com/dwavesystems/penaltymodel/blob/master/penaltymodel_maxgap/setup.py#L19.

The only downside is that solvers become environment-local, i.e. not project-reusable. User has to install the solver(s) in each environment it uses them.

`pysmt-install` already supports this feature through `--bindings-path`, it's just that it defaults to `~/.smt_solvers`.

I don't know what was your rationale behind `~/.smt_solvers`, but if you think this makes sense (for general use case of pysmt and/or average pysmt user), I can submit a PR that will use local `site-packages` dir by default for bindings path... That would simplify solvers install for users (get rid of env vars setting), but it could cause problems for users not using virtual environments (they could still use `pip install --user`, instead of global install). Let me know what you think.

mikand:
@randomir Thanks again! This is a very interesting point I opened an issue to discuss this: #495

PR #503, opened 2018-06-29
Fix docstring for BVAShr.

mpreiner:
BVAShr is a shift operator. I guess this is a copy&paste error from the BVRor below?


mikand:
@mpreiner Thanks a lot for noticing and for the fix!

PR #507, opened 2018-07-04
Examples: Model-checking based on PDR/IC3

marcogario:


marcogario:
This is ok for me to accept, I do not think we want to spend more time improving the description for now.

PR #510, opened 2018-07-07
CI: Upgrade pypy2.7 to pypy3.5

marcogario:
As part of the deprecation of python 2.7, this changes the version of pypy we are testing against to pypy3.5.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,9 @@ matrix:
   # - os: linux
   #   python: 3.6
   #   env: PYSMT_SOLVER="None"      PYSMT_GMPY="FALSE"   PYSMT_CYTHON="TRUE"
-  # - os: linux
-  #   python: 3.6
-  #   env: PYSMT_SOLVER="None"      PYSMT_GMPY="FALSE"   PYSMT_CYTHON="FALSE"
+  - os: linux
+    python: 3.6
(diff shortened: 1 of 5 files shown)


mikand:
It seems that the compilation of pypy failed because of the zlib.

I think we shall add `brew install xz` and check https://github.com/pyenv/pyenv/wiki/common-build-problems#build-failed-error-the-python-zlib-extension-was-not-compiled-missing-the-zlib

mikand:
I tried to follow the suggestions in https://github.com/pyenv/pyenv/issues/1073 . Let's see if this works.

mikand:
I think I managed to compile pypy3, but it seems that the tests are using the system python 2.7

marcogario:
Can you check if this is an issue only with pypy3? Maybe we were also running the pypy2 tests with the system python...

mikand:
It is. When I added the OSX platform I checked for all the cases that the python version printed by the last step of before_install was the one expected. I will try to add an automated check.

mikand:
@marcogario Done!!! Indeed the CI in master was using the system python (2.7) for all the tests on OSX. This is because (differently from linux) the virtualenv needs to be enabled in all the scripts. In the current version all the tests are passing and on OSX the python version is checked automatically in each CI script.

In my opinion this could be accepted as is, but let me know if something is missing.

PR #514, opened 2018-07-14
Boolector: Upgrade to 3.0.0

marcogario:
Updated installer for boolector 3.0.0. This does not include any new feature.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,42 +75,55 @@ matrix:
 
   # Linux - Python 2.7
   - os: linux
+    dist: xenial
     python: 2.7
     env: PYSMT_SOLVER="all"       PYSMT_GMPY="TRUE"
   - os: linux
+    dist: xenial
     python: 2.7
     env: PYSMT_SOLVER="all"       PYSMT_GMPY="FALSE"
   - os: linux
+    dist: xenial
     python: 2.7
     env: PYSMT_SOLVER="None"      PYSMT_GMPY="FALSE"   PYSMT_CYTHON="TRUE"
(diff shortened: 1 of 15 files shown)


aniemetz:
https://github.com/Boolector/boolector/pull/9 is merged in. Thanks for reporting and providing a fix!

marcogario:
The current CI fails because of the following error:
```
python -c "import pyboolector"
  File "<string>", line 1, in <module>
ImportError: /home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/pyboolector.so: undefined symbol: PyFPE_jbuf
```
I think this might be due to a mismatch between the system python and the venv python.

marcogario:
This seems to be a known issue. It mostly affects Mac users (due to the common usage of homebrew and other non-system versions of python). For us the easiest solution would be to call the cmake with the correct flags.

For Travis only, here is how another project solved it for linux.
https://github.com/SimonRit/RTK/commit/689fa97748ca8092a941daafab0fee704c0d68cf

marcogario:
I reported this to the Boolector team to identify a principled solution. What I was trying (and you can take over) was to modify the cmakecache from within our installer.

mikand:
The PR should be almost OK now, but tests are failing because it seems that CVC4 1.5 wrapper does not work on python 2.7 and swig3. My suggestion is to postpone the merge of this pr after #552 that should solve the issue with CVC4.

mikand:
@marcogario This is ready to merge for me, can you have a look and (if ok) go ahead and merge?

Issue #495, opened 2018-06-06
In install.py consider using site-packages as bindings-dir to avoid the need of PYTHONPATH extension


mikand:
This point was raised by @randomir in #494, and I think it could make a lot of sense.

Let's discuss a bit the pros and cons before taking a decision.

randomir:
@marcogario I like your proposal to just copy bindings if they already exist in the install dir. The only problem I see with that approach is making sure the compiled binaries are up to date -- consider user installs an old version of `pysmt`, installs solvers (does compilation), then upgrades `pysmt` (or installs new version in new venv), installs solvers, and gets the old binaries.

On your point of using local packages directory -- it seems doable.

So we would want:
- if venv is active, install in venv's `site-packages`
- otherwise:
  - is user has local site packages enabled (`site.ENABLE_USER_SITE`), install in user site (`~/.local/...`)
  - otherwise attempt global install (`/usr/lib/python/...`)

Right?

marcogario:
Do we need to detect the venv? Wouldn't the installation at site packages be system wide if a venv is not in use? I am just concerned that detection of the venv might be a pain. For the rest I agree on the approach.

Regarding stale bindings, we do not have this problem now, and I think we should not have it if we have venvs. Moving the bindings forces a recompilation. I would keep this as it is,and try to optimize if we really need to.

randomir:
I hope we don’t need to detect venv, I wasn’t trying to say we do that — my outline above was more of functional specification, than actual implementation steps. :)

Ok, I’ll play around with it a little bit.

marcogario:
Any progress on this? I think this would be an important improvement on usability. Let me know if I can help.

randomir:
You are right, sorry, I’ve been busy with other stuff. I’ll get back to this, hopefully have a PR soon.

Btw, unrelated, when do you plan your next release?

marcogario:
No worries, just let me know if we can help.

We do not have a strict schedule for releases, but we try to make 2 releases a year. Recently we got a few bugfixes that I think would justify making a release of 0.7.6 by the end of July.

randomir:
@marcogario, so I guess my question is, do we want to explore this idea further (if you already haven't), or to proceed with my original proposal**?

** which, btw, mutated to the following (and I'm not sure we can do better):
- if user has virtual environment activated, simply install solvers in virtualenv's site packages
- otherwise (debian's pip logic):
  - if user is root, install to system site/dist packages
[...]

PR #517, opened 2018-08-12
Install solvers to a relevant site-package by default

randomir:
Removes need for `PYTHONPATH` patching.

Bulk of work in 7d496ce, the remaining commits update docs and tests.

Addresses #495.

### DIFF
--- a/ci/travis_script.sh
+++ b/ci/travis_script.sh
@@ -44,6 +44,7 @@ if [ "${PYSMT_CUSTOM_BINDINGS_PATH}" == "TRUE" ]; then
 fi
 
 echo "${PYTHONPATH}"
+python install.py --env
 python install.py --check
 
 #
@@ -63,5 +64,8 @@ if [ "${PYSMT_SOLVER}" == "all" ];
 then
     python install.py --msat --conf --force;
     cp -v $(find ~/.smt_solvers/ -name mathsat -type f) /tmp/mathsat;
-    (for ex in `ls examples/*.py`; do echo $ex; python $ex || exit $?; done);
+    for ex in examples/*.py; do
+        echo $ex
+        python $ex
+    done
 fi
--- a/ci/travis_script.sh
+++ b/ci/travis_script.sh
@@ -35,12 +35,15 @@ fi
 echo "Check that the correct version of Python is running"
 python ${DIR}/check_python_version.py "${TRAVIS_PYTHON_VERSION}"
 
-PYSMT_SOLVER_FOLDER="${PYSMT_SOLVER}_${TRAVIS_OS_NAME}"
-PYSMT_SOLVER_FOLDER="${PYSMT_SOLVER_FOLDER//,/$'_'}"
-export BINDINGS_FOLDER=${HOME}/python_bindings/${PYSMT_SOLVER_FOLDER}
+# Update PYTHONPATH only if needed
+if [ "${PYSMT_CUSTOM_BINDINGS_PATH}" == "TRUE" ]; then
+    PYSMT_SOLVER_FOLDER="${PYSMT_SOLVER}_${TRAVIS_OS_NAME}"
+    PYSMT_SOLVER_FOLDER="${PYSMT_SOLVER_FOLDER//,/$'_'}"
+    BINDINGS_FOLDER="${HOME}/python_bindings/${PYSMT_SOLVER_FOLDER}"
+    eval "$(python install.py --env --bindings-path "${BINDINGS_FOLDER}")"
+fi
 
-eval `python install.py --env --bindings-path ${BINDINGS_FOLDER}`
-echo ${PYTHONPATH}
+echo "${PYTHONPATH}"
 python install.py --check
 
 #
--- a/ci/travis_script.sh
+++ b/ci/travis_script.sh
@@ -64,6 +64,11 @@ if [ "${PYSMT_SOLVER}" == "all" ];
 then
     python install.py --msat --conf --force;
     cp -v $(find ~/.smt_solvers/ -name mathsat -type f) /tmp/mathsat;
+
+    # since we're relying on relative `pysmt` import in examples,
+    # ensure `.` is added to `sys.path`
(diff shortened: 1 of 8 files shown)


randomir:
Heads up -- all installs pass on all platforms, and all tests succeed.

The only thing failing currently are additional examples run in "all solvers" case. PySMT import fails there for some reason. Working on it.

randomir:
All righty, all good now. The problem was with `PYTHONPATH` not including `.`, and at the same time example tests relying on `pysmt` being in path (cwd).

Since we don't need to extend `PYTHONPATH` for solvers to work anymore, `PYTHONPATH` was unset. On some systems it's enough to have `PYTHONPATH=''` for python to include cwd in `sys.path`, but I added an explicit `.`.

Btw, tests are now modified to (by default) install all solvers according to the default bindings dir heuristic in the installer. I.e. we're also testing that works.

If you want to run tests with custom/explicit bindings dir, set `PYSMT_CUSTOM_BINDINGS_PATH="TRUE"` in `.travis.yml`.

marcogario:
Great ! Thanks for this. It will take me a few days to review this properly, but on a quick look, it looks good.

ci/travis_script.sh
```diff
@@ -60,5 +64,13 @@ if [ "${PYSMT_SOLVER}" == "all" ];
...
     cp -v $(find ~/.smt_solvers/ -name mathsat -type f) /tmp/mathsat;
-    (for ex in `ls examples/*.py`; do echo $ex; python $ex || exit $?; done);
+
+    # since we're relying on relative `pysmt` import in examples,
+    # ensure `.` is added to `sys.path`
+    export PYTHONPATH=${PYTHONPATH:-.}
+
+    for ex in examples/*.py; do
+        echo $ex
+        python $ex
```
marcogario: This does not have the `exit` anymore. If the example fails, will the script terminate with an error code? If this is not the case, Travis will not know that the test failed.

ci/travis_script.sh
```diff
@@ -60,5 +64,13 @@ if [ "${PYSMT_SOLVER}" == "all" ];
 then
     python install.py --msat --conf --force;
     cp -v $(find ~/.smt_solvers/ -name mathsat -type f) /tmp/mathsat;
-    (for ex in `ls examples/*.py`; do echo $ex; python $ex || exit $?; done);
+
+    # since we're relying on relative `pysmt` import in examples,
+    # ensure `.` is added to `sys.path`
+    export PYTHONPATH=${PYTHONPATH:-.}
```
marcogario: Why wasn't this a problem before?

Would it make sense to pip install the library locally?

pysmt/cmd/installers/base.py
```diff
@@ -307,3 +308,70 @@ def find_python_config(self):
             if command is not None:
                 break
         return command
+
+
+def package_install_site(name='', user=False, plat_specific=False):
```
marcogario: `plat_specific` -> `platform_specific`

marcogario:
Looks good. Only some minor changes, I can do those.

Regarding the CI, we need to fix the caching mechanism.
Travis has an option to cache some of the build process. We do so in our travis.yml ( https://github.com/pysmt/pysmt/blob/master/.travis.yml#L246 ) for `${HOME}/python_bindings`. The idea is to avoid re-compiling the solvers every time (e.g., CVC4 requires 30 minutes to build).

What directory do we need to cache now? I guess ~/.local/bin on linux ? What about macOS?

For the same reason, I re-triggered the CI on travis after deleting the cache.

ci/travis_script.sh
randomir: Doesn't Travis use something like `set -x`, so the script/test terminates on the first failed command?

This loop should be functionally identical to the one you had -- where you had `exit`, but from a separate subshell. This: `(python $ex; exit $?)` is identical to `python $ex` (in terms of exit status).

ci/travis_script.sh
randomir: This wasn't a problem before because before `PYTHONPATH` was always explicitly set.

Absolutely, local `pip install .` is an option. I went with this solution because it required minimal changes (to stay as close to the original solution).

pysmt/cmd/installers/base.py
randomir: `plat_specific` is standard name in [`distutils`](https://docs.python.org/2/distutils/apiref.html).

randomir:
Since the current Travis build uses virtual env, python bindings for solvers are installed in site-packages inside that env. So maybe you can cache the complete virtual env for the project?

The path to the virtual env root is (for python 3.6 build): `/home/travis/virtualenv/python3.6.3/`, and path to site-packages is: `/home/travis/virtualenv/python3.6.3/lib/python3.6/site-packages`. More general, the path will be in `VIRTUAL_ENV` environment variable.

If you want me to patch something, just let me know. From the comments above, I'm not exactly sure what would you like to be changed... Goes without saying, please feel free to modify/amend my PR to your liking.

ci/travis_script.sh
randomir: Just to clarify, my motivation for refactoring this loop was the use of [`for in ls results` antipattern](http://mywiki.wooledge.org/ParsingLs). I replaced subshell invocations as well because I don't see the need for them here. Correct me if I'm wrong.

ci/travis_script.sh
marcogario: ``pip install`` might be the cleaner version, but let's keep that for another time.

randomir:
@marcogario, any idea why that z3 test keeps segfaulting? I've restarted it several times. Maybe something is cached that's not working?

marcogario:
This is a known issue that we have been trying to isolate for a while. Nothing to do with yhis particular pr. It is difficult to debug because I managed to reproduce it only once on my machine and otherwise it is non deterministic. One idea of the cause is the an incorrect use of the refcount mechanism in our z3 converter. I was reviewing it a few weeks ago, but did not find anything suspicious. We might try to upgrade z3, if there is a more recent version available.

Issue #520, opened 2018-09-04
RuntimeError: generator raised StopIteration in pysmt.smtlib.parser on python3.7


chaserhkj:
Running [the parser example in the tutorial](https://pysmt.readthedocs.io/en/latest/tutorials.html#demonstrates-how-to-perform-smt-lib-parsing-dumping-and-extension) on python 3.7 would yield the following exception:

```
Traceback (most recent call last):
  File "/usr/tmp/env/lib/python3.7/site-packages/pysmt/smtlib/parser/parser.py", line 218, in create_generator
    c = next(reader)
StopIteration

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/hkj/Projects/Research/DryadSynth-Data/t.py", line 46, in <module>
    script = parser.get_script(cStringIO(DEMO_SMTLIB))
  File "/usr/tmp/env/lib/python3.7/site-packages/pysmt/smtlib/parser/parser.py", line 834, in get_script
    for cmd in self.get_command_generator(script):
  File "/usr/tmp/env/lib/python3.7/site-packages/pysmt/smtlib/parser/parser.py", line 848, in get_command_generator
    for cmd in self.get_command(tokens):
  File "/usr/tmp/env/lib/python3.7/site-packages/pysmt/smtlib/parser/parser.py", line 1068, in get_command
    self.consume_opening(tokens, "<main>")
  File "/usr/tmp/env/lib/python3.7/site-packages/pysmt/smtlib/parser/parser.py", line 1025, in consume_opening
    p = tokens.consume()
  File "/usr/tmp/env/lib/python3.7/site-packages/pysmt/smtlib/parser/parser.py", line 189, in consume_token
    return next(self.generator)
RuntimeError: generator raised StopIteration
```
This is a python3.7-only issue, running on any previous version works fine.

I haven't got the chance to look into it further, but it seems to me that it is related to [this PEP 479](https://www.python.org/dev/peps/pep-0479/) introduced in 3.7.

marcogario:
Thanks for reporting this. Seems like this has escalated with 3.7. We had open iasue #488 about this, and you are right on the PEP479. Would you be able to provide a patch? I can support you in the process.

If not, that is totally ok, we appreciate you talking the time to report this.

marcogario:
I started working on this on the branch https://github.com/pysmt/pysmt/tree/i520_stop_iteration . The main problem is that we were relying on StopIteration bubbling up in many contexts.

[...]

PR #521, opened 2018-09-14
Fix StopIteration use in SmtLib parser

mikand:
This PR addresses #520 

Tests are passing, but we need to double-chekc the use of `consume()` vs `consume_enforce()` in all the cases.

### DIFF
--- a/pysmt/smtlib/parser/parser.py
+++ b/pysmt/smtlib/parser/parser.py
@@ -15,6 +15,8 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 #
+from __future__ import generator_stop
+
 import functools
 import itertools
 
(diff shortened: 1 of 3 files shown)


pysmt/smtlib/parser/parser.py
```diff
@@ -785,45 +802,49 @@ def get_expression(self, tokens):
...
 
-                if tk in self.interpreted:
-                    fun = self.interpreted[tk]
-                    fun(stack, tokens, tk)
-                else:
-                    stack[-1].append(self.atom(tk, mgr))
+                if tk == "(":
+                    while tk == "(":
+                        stack.append([])
+                        tk = tokens.consume()
```
mikand: Maybe we can use `consume_enforce` here as we cannot terminate with a `(`

marcogario:
Can we keep the import future and condition it to py3? Without that I was unable to get warnings for this issue. Since this issue breaks 3.7 i think forward compatibility is more important than 2.7 compatibility

mikand:
@marcogario Unfortunately, the future imports need to be the first statement of a python file, hence they cannot be conditioned.
Anyhow, I get a warning even without the future import on python 3.6:
```python
import warnings
warnings.filterwarnings('default')

def g():
    yield 1

def f():
    x = g()
    yield next(x)
    next(x)

for y in f():
    print(y)
```

```
$ python3 /tmp/t.py 
1
/tmp/t.py:15: DeprecationWarning: generator 'f' raised StopIteration
  for y in f():
```

Another thing we can do is to switch our CI to 3.7 that raises an error instead.

marcogario:
Let's change ci to 3.7.

marcogario:
If test pass on 3.7, go ahead and squash merge this

marcogario:
See #529 for testing on 3.7.

PR #531, opened 2018-10-22
Trivial improvements for bit vector simplification

hzzhang:
This pull request implements the simplification of bit vector items in some trivial cases, such as simplifying BVAdd(x, BVZero(n)) to BVZero(n). Therefore, users may not need to call the expensive BDD simplifier for those trivial cases.

### DIFF
--- a/pysmt/test/test_bv_simplification.py
+++ b/pysmt/test/test_bv_simplification.py
@@ -20,7 +20,7 @@ from six.moves import xrange
 from pysmt.shortcuts import Solver, BVAnd, BVOr, BVXor, BVConcat, BVULT, BVUGT, \
     BVULE, BVUGE, BVAdd, BVSub, BVMul, BVUDiv, BVURem, BVLShl, BVLShr, BVNot, \
     BVNeg, BVZExt, BVSExt, BVRor, BVRol, BV, BVExtract, BVSLT, BVSLE, BVComp, \
-    BVSDiv, BVSRem, BVAShr, get_env, EqualsOrIff, BVZero, BVOne, Symbol
+    BVSDiv, BVSRem, BVAShr, get_env, EqualsOrIff, BVZero, BVOne, Symbol, Bool
 from pysmt.typing import BVType
 from pysmt.test import TestCase, skipIfSolverNotAvailable, main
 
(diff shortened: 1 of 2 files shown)


hzzhang:
@marcogario 
The continuous-integration/appveyor/pr failed, and the details showed 

> AppVeyor failed to assign job to 'gce' build environment after 3 attempt(s). The envronment is either unavailable or there are no free workers, however alternate build environment was not configured by the system administrator. Please try again later.

Does it mean something goes wrong in the CI system itself?

pysmt/test/test_bv_simplification.py
```diff
@@ -94,6 +95,316 @@ def all_special(self):
...
+        f = BVMul(x, BVZero(32))
+        self.assertEqual(f.simplify(), BVZero(32))
+
+    def test_bv_mul_1(self):
+        x = Symbol("x", BVType(32))
+        f = BVMul(x, BVOne(32))
+        self.assertEqual(f.simplify(), Symbol("x", BVType(32)))
+
+    def test_bv_mul_constants(self):
+        f = BVMul(BV(10, 32), BV(12, 32))
```
marcogario: Does this consider overflow?

pysmt/test/test_bv_simplification.py
```diff
@@ -94,6 +95,316 @@ def all_special(self):
                 for op in self.special_operators:
                     self.check(op(l, n))
 
+    def test_bv_0_add(self):
+        x = Symbol("x", BVType(32))
+        f = BVAdd(BVZero(32), x)
+        self.assertEqual(f.simplify(), Symbol("x", BVType(32)))
+
```
marcogario: I would use directly x in the assert. It is more clear.

pysmt/test/test_bv_simplification.py
```diff
@@ -94,6 +95,316 @@ def all_special(self):
...
+        f = BVUDiv(x, BVZero(32))
+        self.assertEqual(f.simplify(), BV(2**32 - 1, 32))
+
+    def test_bv_0_udiv_0(self):
+        f = BVUDiv(BVZero(32), BVZero(32))
+        self.assertEqual(f.simplify(), BV(2**32 - 1, 32))
+
+    def test_bv_nonzero_udiv_0(self):
+        f = BVUDiv(BV(10, 32), BVZero(32))
+        self.assertEqual(f.simplify(), BV(2**32 - 1, 32))
```
marcogario: I don't remember what is the agreed value for this. This is undefined in arithmetic theories. Is this defined in the smtlib standard?

marcogario:
For the ci, I think mikand ia working on this in anothet branch. If travis is green, that is good for me.

pysmt/test/test_bv_simplification.py
hzzhang: No, but I'll add tests for overflow cases in the next PR.

pysmt/test/test_bv_simplification.py
hzzhang: Will change in the next PR.

pysmt/test/test_bv_simplification.py
hzzhang: Simplifier::walk_bv_udiv() before this PR does so, as well as Z3, though the bit vector division by zero seems invalid in [smtlib standard](http://smtlib.cs.uiowa.edu/theories-FixedSizeBitVectors.shtml):

>   o _ div _,  which takes an integer x ≥ 0 and an integer y > 0 and returns
>     the integer part of x divided by y (i.e., truncated integer division).
>
>   o _ rem _, which takes an integer x ≥ 0 and y > 0 and returns the
>     remainder when x is divided by y.  Note that we always have the following
>     equivalence for y > 0: (x div y) * y + (x rem y) = x.

hzzhang:
@marcogario 
I've updated the PR and CI now failed at test_tactics_z3. The test case failed due to a segment fault, which I cannot reproduce locally. It succeeded before the update, and all updates are in test_bv_simplification.py and should not affect test_tactics_z3. Does it relate to #465 which may not be completely fixed by #530 ?

mikand:
@hzzhang I just re-launched the CI test, let's see if it it deterministic or not. Interestingly, this time the segfauult happened on OSX, while in #465 it only happened on Linux...

hzzhang:
@mikand All tests passed at this time.

Issue #318, opened 2016-08-29
Extend CVC4 support for Non Linear Real Arithmetic


ghost:
I get the following error when I pass a QF_NRA benchmark to cvc4 solver.
I think CVC4 can handle QF_NRA benchmarks. 
So reporting this bug.

```
Traceback (most recent call last):
  File "/es0/irfan/pysmt_nra_smt/bin/pysmt_cvc4.py", line 18, in <module>
    main()
  File "/es0/irfan/pysmt_nra_smt/bin/pysmt_cvc4.py", line 11, in main
    cvc4.add_assertion(f)
  File "/es0/irfan/local/pysmt35/pysmt/solvers/cvc4.py", line 81, in add_assertion
    term = self.converter.convert(formula)
  File "/es0/irfan/local/pysmt35/pysmt/decorators.py", line 83, in catch_conversion_error_wrap
    res = f(*args, **kwargs)
  File "/es0/irfan/local/pysmt35/pysmt/solvers/cvc4.py", line 214, in convert
    return self.walk(formula)
  File "/es0/irfan/local/pysmt35/pysmt/walkers/dag.py", line 111, in walk
    res = self.iter_walk(formula, **kwargs)
  File "/es0/irfan/local/pysmt35/pysmt/walkers/dag.py", line 103, in iter_walk
    self._process_stack(**kwargs)
  File "/es0/irfan/local/pysmt35/pysmt/walkers/dag.py", line 96, in _process_stack
    self._compute_node_result(formula, **kwargs)
  File "/es0/irfan/local/pysmt35/pysmt/walkers/dag.py", line 80, in _compute_node_result
    self.memoization[key] = f(formula, args=args, **kwargs)
  File "/es0/irfan/local/pysmt35/pysmt/solvers/cvc4.py", line 294, in walk_times
    raise NonLinearError(formula)
pysmt.exceptions.NonLinearError: (b.speed_x__AT7 * (-1.0 * b.speed_x__AT7))
```

marcogario:
The message that you are getting is by design. If you were to instantiate the solver by specifying that you will be using QF_NRA, you would get an exception telling you that the logic is not supported by the solver. 
In order to provide a reasonable interface across solvers, we need to limit the scope of which theories are supported by whom. Which logics are currently supported is described in [1] (or you can look at Solver.LOGICS within the source code of each solver).

[...]

PR #529, opened 2018-10-13
CI: Python 3.7 on Linux

marcogario:
Upgrade CI to use python 3.7 . 
The CI environment uses an older version of SWIG (2.X), but CVC4 requires SWIG 3. We install SWIG 3.0.10 from source.

See https://github.com/travis-ci/travis-ci/issues/9069#issuecomment-425720905

PR #533, opened 2018-10-26
Nonlinear support

ahmed-irfan:
Nonlinear support: Mathsat and CVC4


ahmed-irfan:
fixes #318

mikand:
@ahmed-irfan You should also fix the LOGICS field of both MathSAT and CVC4 to allow for quantified logics, otherwise these solvers cannot be used by the automatic solver selector...

ahmed-irfan:
> The tests fail. The one for CVC4 should be fixed once #529 is merged (seems to be related to an old SWIG version), but MathSAT seems unable to solve the problem and just hangs. Can you try to reproduce the test pysmt.test.test_nlira.TestNonLinear.test_unknownresult ?

I had a look at the test "test_unknownresult ". I did not understand what we are checking in that test. Is there a particular reason for checking an SMT problem with only an irrational solution? That is same reason why Mathsat gets stuck....

marcogario:
The name of the test is very bad. There should be one test for a simple non linear expression and one for an irrational result. In the second case, I would be ok with the solver answering unknown. I think guaranteeing termination is something useful. Is there a way to configure mathsat to "give up" after a give number of iterations?

ahmed-irfan:
> The name of the test is very bad. There should be one test for a simple non linear expression and one for an irrational result. In the second case, I would be ok with the solver answering unknown. I think guaranteeing termination is something useful. Is there a way to configure mathsat to "give up" after a give number of iterations?

OK. So, I am going to modify the tests. 
And yes, there is a way in MathSAT to guarantee termination.

ahmed-irfan:
Mathsat is failing on a test containing division by a variable.........

marcogario:
The question is why is the test failing. If the logic includes division, then that is a valid example for the logic. Clearly we do not expect this to work in all cases, but then mathsat could detect this case and return an error, or return and unknown as result of the solving.

Why can't you rewrite this as multiplication with the additional assumption that the variable is non zero? If the new problem is sat, also the original is. If it is unsat, it relies on the total interpretation of div by zero, and you can answer unknown.

mikand:
@agriggio Can comment on this PR? Is the division in QF_NRA? Is the MathSAT behavior expected?

ahmed-irfan:
UPDATE:
I extended the mathsat and cvc4 wrappers with the division operator. That should fix the division tests.
Moreover, I simplified some NRA examples used in the test_solving.Basic.

Regarding if the division in QF_NRA or not, that I think is a separate issue (#537) and it is still a confusion to me.  

Info: There is still some work to be done in this PR.

ahmed-irfan:
@mikand or @marcogario :
Could you please take a look why the tests for CVC4 are getting stuck?

Thanks.

ahmed-irfan:
@mikand / @marcogario : tests are passing now on the new CI system.
PR is ready for review....

pysmt/test/examples.py
```diff
@@ -912,8 +913,8 @@ def get_full_example_formulae(environment=None):
                     logic=pysmt.logics.QF_NRA
                 ),
 
-            Example(hr="((r * r * r) = 25.0)",
-                    expr=Equals(Times(r, r, r), Real(25)),
+            Example(hr="((r * r * r) = 27.0)",
+                    expr=Equals(Times(r, r, r), Real(27)),
```
mikand: Maybe we can also add an example with powers?

pysmt/solvers/msat.py
```diff
@@ -978,16 +981,27 @@ def walk_implies(self, formula, args, **kwargs):
...
-            if not mathsat.msat_term_is_number(self.msat_env(), x):
-                nl_count += 1
-            if nl_count >= 2:
-                raise NonLinearError(formula)
-            else:
-                res = mathsat.msat_make_times(self.msat_env(), res, x)
+            res = mathsat.msat_make_times(self.msat_env(), res, x)
         return res
 
+    def walk_pow(self, formula, args, **kwargs):
```
mikand: This function assumes that the power is integer, is this the case in the general logic or something like `pow(x, 1/2)` is allowed? (I am asking for genuine curiosity and because the pysmt function `Pow` that constructs this term accepts non-integer (but constant) exponents)

pysmt/solvers/msat.py
```diff
@@ -978,16 +981,27 @@ def walk_implies(self, formula, args, **kwargs):
...
+        n = to_python_integer(formula.args()[1].constant_value())
+        if n == 0:
+            return mathsat.msat_make_number(self.msat_env(), "1")
+        is_neg = True if n < 0 else False
```
mikand: `is_neg = (n < 0)`? :smile:

pysmt/test/test_nlira.py
```diff
@@ -32,25 +32,26 @@
 
 class TestNonLinear(TestCase):
 
-    @skipIfSolverNotAvailable("z3")
     def test_times(self):
         x = FreshSymbol(REAL)
         f = Equals(Times(x, x), x)
-        with Solver(name="z3") as s:
-            self.assertTrue(s.is_sat(f))
+        for sname in self.env.factory.all_solvers():
```
mikand: You can do this as follows:
```
for sname in self.env.factory.all_solvers('QF_NRA'):
    with Solver(name=sname, logic='QF_NRA') as s:
        self.assertTrue(s.is_sat(f))
```

pysmt/test/test_nlira.py
```diff
@@ -66,35 +67,38 @@ def test_oracle(self):
...
-        # self.assertTrue(is_sat(f, solver_name="z3"))
-
-        # f = Equals(Div(Int(4), x), Int(2))
-        # self.assertTrue(is_sat(f, solver_name="z3"))
-
-        f = Equals(Times(x, x), Int(16))
-        self.assertTrue(is_sat(f, solver_name="z3"))
+        f1 = Equals(Times(x, x), Int(2))
+        f2 = Equals(Times(x, x), Int(16))
+        for sname in self.env.factory.all_solvers():
```
mikand: Same as above

mikand (changes requested):
I marked some minor stuff and a major question about the possibility of having non-integral powers

pysmt/solvers/msat.py
ahmed-irfan: I think the power node can have integer or rational exponent. 
However, to stay in NRA / NIA we should check that the exponent is a positive integer. 
Maybe it should be done for all solvers that support nonlinear arithmetic in a separate PR.

pysmt/solvers/msat.py
```diff
@@ -965,16 +968,27 @@ def walk_implies(self, formula, args, **kwargs):
...
+        res = args[0]
+        for i in range(2, n):
+            res = mathsat.msat_make_times(self.msat_env(), res, args[0])
+        if is_neg:
+            one = mathsat.msat_make_number(self.msat_env(), "1")
+            res = mathsat.msat_make_divide(self.msat_env(), one, res)
+        return res
+
+    def walk_div(self, formula, args, **kwargs):
+        return mathsat.msat_make_divide(self.msat_env(), args[0], args[1])
```
EnricoMagnago, line 990: Sorry if I intrude.    
If I understood correctly in pysmt `div` could be between integer or real division, while in msat `make_divide` is always real division.

We could rewrite a pysmt integer division in mathsat using floor and real division.
The following code should perform the rewriting.
```
    def walk_div(self, formula, args, **kwargs):
        if self.env.stc.get_type(formula).is_real_type():
            return mathsat.msat_make_divide(self.msat_env(), args[0], args[1])
        assert self.env.stc.get_type(formula).is_int_type()
        # smtlib2 semantics: den >= 0 ? floor(num / den) : ceil(num / den)
        # In the following we rewrite ceil(a) as -floor(-a)
        num = args[0]
        den = args[1]
        zero = mathsat.msat_make_number(self.msat_env(), "0")
        neg = mathsat.msat_make_number(self.msat_env(), "-1")
        # den >= 0
        cond = mathsat.msat_make_leq(self.msat_env(), zero, den)
        # floor(num / den)
        div = mathsat.msat_make_divide(self.msat_env(), num, den)
        div = mathsat.msat_make_floor(self.msat_env(), div)
        # - floor(num / -den)
        n_div = mathsat.msat_make_times(self.msat_env(), neg, den)
        n_div = mathsat.msat_make_divide(self.msat_env(), num, n_div)
        n_div = mathsat.msat_make_floor(self.msat_env(), n_div)
        n_div = mathsat.msat_make_times(self.msat_env(), neg, n_div)
        # den >= 0 ? floor(num / den) : - floor(num / -den)
        return mathsat.msat_make_term_ite(self.msat_env(), cond, div, n_div)
```

masinag:
Hi, what's the status of this PR? Do you have plans to merge it anytime soon?

mikand:
Superseded by #844

PR #544, opened 2018-12-03
Toplevel propagator

ahmed-irfan:
This adds a basic toplevel-propagation functionality.
It propagates definitions of the form:

-  variable = constant

-  variable = variable

-  constant = constant


pysmt/rewritings.py
```diff
@@ -832,3 +832,29 @@ def disjunctive_partition(formula):
...
+    if env is None:
+        import pysmt.environment
+        env = pysmt.environment.get_env()
+    mgr = env.formula_manager
+    sigma = {}
+    for c in conjunctive_partition(formula):
+        if c.is_equals():
+            l, r = sorted(c.args(), key=hash)
+            if l.is_symbol() and (r.is_symbol() or r.is_constant()) and\
+               l not in sigma:
```
mikand: What happens when we have a formula like `phi & a = b & a = c` assuming that hash(a) < hash(b) < hash(c)?

It seems to me that sigma will only contain a -> b as a substitution and not c -> b. I think you need a data structure to keep the equivalence classes instead of a simple `dict` and then you need to select a representative and substitute all the members of the same class with the representative....

pysmt/rewritings.py
```diff
@@ -832,3 +832,29 @@ def disjunctive_partition(formula):
...
+    2) variable = variable
+    """
+    if env is None:
+        import pysmt.environment
+        env = pysmt.environment.get_env()
+    mgr = env.formula_manager
+    sigma = {}
+    for c in conjunctive_partition(formula):
+        if c.is_equals():
+            l, r = sorted(c.args(), key=hash)
```
mikand: Two elements could have the same hash and so you might loose the ordering you want (This is not true with pysmt formulae, but it creates confusion). You can use the `FNode.node_id` function instead of the `hash` function. The id of two pysmt formulae are guaranteed to be different if the two objects are different.

pysmt/rewritings.py
ahmed-irfan: You are right, it will not catch the case you mentioned.

Actually, I wanted to start from a very basic one and then later improve it.

What do you think? 
Should we go for better version with some equivalence classes implementation?

pysmt/rewritings.py
ahmed-irfan: Ah, I didn't know that. 
Thanks for pointing this out.

pysmt/rewritings.py
mikand: I think it should be rather simple to use a UnionFind to construct equivalence classes and then rewrite the formula. You can have a look [https://stackoverflow.com/a/3067672](here) for a simple dictionary-based UnionFind implementation.

Instead, [http://code.activestate.com/recipes/215912-union-find-data-structure/](here) there is a proper, efficient implementation. A good idea could be to put a generic UnionFind implementation in a `util` package in pysmt for possible other uses...

marcogario: I would go with the simple implementation, and put it directly in the module where this is used. AFAIK there has no other place where we use this, and if the needs arises, we can extract that.

If performance is a problem (I would assume this is the case for expressions with a huge number of equalities), we can try to cythonize the implementation similarly to what we do for the SMT-LIB parser, or identify an external library that implements this in C.

ahmed-irfan: OK. I am going to go for the basic implementation, as @marcogario suggested.

ahmed-irfan:
@mikand can you please take a look?
The tests are failing on the Mac build.

mikand:
I re-launched the CI, probably there was a network fluke that made some tests fail.

I had a look at the code and I have two further comments:
* It would be nice to encapsulate the union-find logic into a class to avoid the nesting of functions and simplify the readability
* I am not sure that the code guarantees equi-satisfiability if the equivalence flag is set to False: consider the formula `phi & (5 = 0)`, with the existing code it will be transformed into `phi[0->5]` that might be satisfiable! I think you need to check that a constant is never inserted into a group whose leader is not the constant itself.

ahmed-irfan:
@mikand the tests are still failing on Travis for the Z3-mac build.
Can you please take a look?

pysmt/rewritings.py
```diff
@@ -832,3 +887,60 @@ def disjunctive_partition(formula):
...
+        import pysmt.environment
+        env = pysmt.environment.get_env()
+    mgr = env.formula_manager
+
+    # comparison function for ranking
+    def compare(a, b):
+        if a.node_id() == b.node_id():
+            return 0
+        if a.is_constant() and b.is_constant():
+            return (a.constant_value() > b.constant_value()) - (a.constant_value()\
```
mikand: Why not simply `a.constant_value() - b.constant_value()`? Subtracting Booleans looks a bit strange...

pysmt/rewritings.py
```diff
@@ -832,3 +887,60 @@ def disjunctive_partition(formula):
...
+    relevant = set()
+    
+    for c in conjunctive_partition(formula):
+        if c.is_equals():
+            l, r = c.args()
+            if l.is_array_value() or r.is_array_value():
+                # skipping constant arrays
+                continue
+            if (l.is_symbol() and (r.is_symbol() or r.is_constant())) or\
+               (r.is_symbol() and (l.is_symbol() or l.is_constant())):
```
mikand: Here you are skipping the "constant = constant" case, is it intentional?

ahmed-irfan:
Taken into account @mikand 's feedback....
MR is ready for merge

PR #545, opened 2018-12-04
Quick and dirty fix to make CVC4 installer work again

mikand:


mikand:
@marcogario I understand that you are working to update to the new version of CVC4, but this is a quick fix that should make the tests pass (modulo z3 segfault)

marcogario:
Lgtm. Thanks for fixing this, the upgrade of cvc4 is taking longer than expected, and I will not manage to look into that for the next few weeks.

If you want to take a look at the branch for upgrading cvc, be my guest! The issue is similar to the one with boolector. Cmake tries to be too smart and there is no way (short of changing the cmake list) for overriding the defaults. In this case, it finds travis ci swig version instead of the one ee install manually.

PR #552, opened 2019-01-15
Updated CVC4 to 1.7-prerelease

mikand:


marcogario:
Disabling the test is not a great solution. Can we detect the situation that triggers the fault and raise an exception from within the cvc solver when a user tries to pass an option in py2.7?

Otherwise it might be ok, but we need to make this very clear in the release notes.

mikand:
Unfortunately I was unable to find a way to circumvent the exception in general, the only option would be to add a set of the allowed options and raise the error in all the other cases, but the possibilities are a lot...

I opened an issue on CVC4 to see if we can fix this problem upstream:  https://github.com/CVC4/CVC4/issues/2810

mikand:
For now, I am going to merge this PR, as soon as https://github.com/CVC4/CVC4/issues/2810 we will re-enable the test and upgrade the CVC4 version

PR #553, opened 2019-01-18
Fixes for OSX installation of CVC4

alastairreid:
I ran into four problems while installing the CVC4 support in OSX.
1) A deprecation warning about a feature that will disappear in python 3.8
2) A problem due to the script assuming GNU version of sed (OSX uses BSD version by default).  (Change should work on Linux but untested)
3) A problem in running CVC4 configure script that makes it crash.  Applied the recommended change from the error message.
4) The name of the generated .so file was incorrect causing installation to fail at the last moment.  I would guess that this will break Linux builds so maybe it needs to try one filename and, if that fails, try the other filename?  I'm not in a position to test this at the moment though.

### DIFF
--- a/pysmt/cmd/installers/cvc4.py
+++ b/pysmt/cmd/installers/cvc4.py
@@ -75,7 +75,7 @@ class CVC4Installer(SolverInstaller):
 
 
         # Fix the paths of the bindings
-        SolverInstaller.mv(os.path.join(self.bin_path, "lib/pyshared/CVC4.so.4.0.0"),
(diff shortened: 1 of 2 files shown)


marcogario:
Thanks for the PR. We had not gotten around to extend the installer to OSX, but this seems to cover most of the obstacles. As you probably already guessed, we mostly develop on Linux...
Please let me know when this works for you, I can take care on fixing it to make it work on Linux, and then extend the Travis CI test matrix to include the CVC4/OSX combination.

mikand:
Superseded by support of CVC5 (CVC4 is now legacy and unmaintaned)

PR #558, opened 2019-02-12
Fix bug in deque pop in smtlib/parser

Phat3:
Collection.deque does not have a method pop that takes arguments (i.e. `pop(0)` does not exist). I replaced the method with `popleft()` which achieve the same result as the intended one.

### DIFF
--- a/pysmt/smtlib/parser/parser.py
+++ b/pysmt/smtlib/parser/parser.py
@@ -189,7 +189,7 @@ class Tokenizer(object):
         """Consumes and returns a single token from the stream.
            If the stream is empty, a PysmtSyntaxError is thrown"""
         if self.extra_queue:
-            return self.extra_queue.pop(0)
+            return self.extra_queue.popleft()
         else:
             try:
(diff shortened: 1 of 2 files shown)


mikand:
@Phat3 Thanks for this! It probably slipped into when we changed from a list to a deque for performance.

Do you have an SMT2 example we can add to the test-suite that triggers the problem with the old code? It would be great to add it as a no-regression test.

Phat3:
Yep sure! 

The SMT-LIB script that cause the problem for me is  the following:

```
(model
(define-fun STRING_cmd_line_arg_1_1000 () String "AAAAAAAAAAAA")
)
```

Should I provide a real test case or this is fine?

mikand:
If you could put this in a testcase in e.g. `pysmt/test/smtlib/test_parser_examples.py` it would be great!

Phat3:
Test case added :+1:

mikand:
@Phat3 Thanks a lot for this PR!

PR #561, opened 2019-02-24
sort the times args while simplifying

ahmed-irfan:
the args of the times args are processed using a stack. Each call
to simplify reverts the order of nonlinear multiplication. And because of that
the simplify function could not reach a fix-point if called repeatedly.
As a solution, we perform sorting of the arguments while doing simplication.

### DIFF
--- a/pysmt/simplifier.py
+++ b/pysmt/simplifier.py
@@ -23,6 +23,7 @@ import pysmt.operators as op
 import pysmt.typing as types
 from pysmt.utils import set_bit
 from pysmt.exceptions import PysmtValueError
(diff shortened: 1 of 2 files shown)


mikand:
LGTM, but i would use `FNode.node_id` instead of `hash` as a sorting function

ahmed-irfan:
> LGTM, but i would use `FNode.node_id` instead of `hash` as a sorting function

Done.

marcogario:
Accepting as is. To consider whether this is something that might make sense to do at construction, since it would also guarantee that equivalent expressions have a single representation (since they are normalized using the id).

PR #562, opened 2019-03-06
Fix import warning in python 3.7

cdonovick:
Importing `Iterable` from `collections` is deprecated in python3.7 and will be removed in python3.8

https://docs.python.org/3/library/collections.html


mikand:
@cdonovick  Thanks a lot for this contribution!

PR #564, opened 2019-03-07
Set up CI with Azure Pipelines

mikand:
Testing file


mikand:
So far, I managed to get MathSAT working uniformly on Linux, OSX and Win. Next step is Z3!

mikand:
Note: for #465 in this PR we temporarily disabled two tests involving using Z3 tactics that caused segfaults. We need to understand the issue and re-enabled the tests.

azure-pipelines.yml
```diff
@@ -0,0 +1,343 @@
+jobs:
+- job: 'Ubuntu16All'
+  condition: or(eq(variables['Build.SourceBranchName'], 'master'), eq(variables['Build.Reason'], 'PullRequest'))
```
marcogario: Why is this needed? can't we use the following syntax:
```
# Only build PRs that target master
pr:
  # Additional pushes to a PR cancel the current build
  autoCancel: true 
  branches:
    include: 
      - master
```

azure-pipelines.yml
```diff
@@ -0,0 +1,343 @@
...
+  pool:
+    vmImage: 'Ubuntu-16.04'
+  strategy:
+    matrix:
+      Py27All:
+        python.version: '2.7'
```
marcogario: Why do we need to declare this twice?

azure-pipelines.yml
```diff
@@ -0,0 +1,343 @@
...
+  - bash: |
+      curl --output pypy.tmp/get-pip.py https://bootstrap.pypa.io/get-pip.py
+      $(pypy_binary) pypy.tmp/get-pip.py
+    displayName: 'Get PIP'
+  - bash: ci/install_linux.sh
+    displayName: 'Install and Test'
+
+
+
+- job: 'Ubuntu16Py37'
```
marcogario: I would run this as first job, then the Ubuntu16All and then the Ubutnu16Py27. Failures in this one give us a better localization of the problems.

ci/install_linux.sh
```diff
@@ -0,0 +1,98 @@
...
+    cp -v $(find ~/.smt_solvers/ -name mathsat -type f) pysmt/test/smtlib/bin/mathsat;
+    mv pysmt/test/smtlib/bin/mathsat.solver.sh.template pysmt/test/smtlib/bin/mathsat.solver.sh ;
+fi
+
+# Check that the solvers are installed
+${PYTHON} install.py --check
+
+
+# Run the test suite
+${PYTHON} -m nose pysmt -v # --with-coverage --cover-package=pysmt
```
marcogario: Can we move this outside of the install script?

ci/install_linux.sh
```diff
@@ -0,0 +1,98 @@
...
+
+
+# Test examples in examples/ folder
```
marcogario: Same. Let's keep installation and testing separated.

ci/install_osx.sh
```diff
@@ -0,0 +1,95 @@
...
+    cp -v $(find ~/.smt_solvers/ -name mathsat -type f) pysmt/test/smtlib/bin/mathsat;
+    mv pysmt/test/smtlib/bin/mathsat.solver.sh.template pysmt/test/smtlib/bin/mathsat.solver.sh ;
+fi
+
+# Check that the solvers are installed
+python install.py --check
+
+
+# Run the test suite
+python -m nose pysmt -v # --with-coverage --cover-package=pysmt
```
marcogario: Same. Separate installation and testing.

pysmt/test/test_regressions.py
```diff
@@ -396,7 +396,7 @@ def test_smtlib_define_fun_serialization(self):
         self.assertEqual(res.replace('__x0', 'x'), smtlib_input)
 
     @skipIfSolverNotAvailable("z3")
-    def test_z3_nary_back(self):
+    def disabled_test_z3_nary_back(self):
```
marcogario: doesn't the @skip annotation work in this case?

marcogario:
are branches being tested or only PRs?

azure-pipelines.yml
mikand: Unfortunately, we cannot. The `pr` and `trigger` keywords operate at the build granularity, here we need to enable or disable specific jobs. The relevant examples are here: 
https://docs.microsoft.com/en-us/azure/devops/pipelines/process/conditions?view=azure-devops&tabs=yaml

mikand:
@marcogario I addressed all your comments and (assuming the windows build is fixed) the refactoring should work.

As for your questions:
* Do we need to setup the parallelism level? maxParallel keyword, or is this running with 10 parallel instances?
  No, without specifying the `maxParallel` constarint we get the maximal number of parallel jobs (and it is working in the builds)

* Do we have early termination if one of the initial jobs fails?
  No, we run everything. Do you think this is needed? Which is the dependency you envisage? (Note that I could not find an early termination keyword, so this can limit parallelism if we use the job dependencies mechanism...)

marcogario:
For early termination, I would say that we need some data. Can we at least prioritize the jobs? How long is the end to end execution right now? 

Ideally, python 3 on linux gives us 90pcnt of the confidence that the build is working. I would run that as first job and everything else afterwards. I am ok if we do this later.

mikand:
As you can see here (https://dev.azure.com/micheliandrea/pysmt/_build?definitionId=1) The build of python3 only (the branch build) takes 37 minutes (as usual CVC4 takes 35 minutes to build) while the full PR build with everything takes 47, thanks to the massive parallelism of azure pipelines.

PR #565, opened 2019-03-12
add noughts and crosses example

mattvenn:
Hi, thanks for the library.
I'm new to SAT and am learning with your library and symbiyosys for formal verification of verilog.

I created this noughts and crosses demo and then thought it might be a good example, as there aren't that many basic examples.

I'm very open to any changes you suggest, I'm fairly sure I could be doing this in a simpler way (the sudoku example seems more complicated but the program seems simpler).

Cheers!


examples/xoxo/xoxo.py
```diff
@@ -0,0 +1,183 @@
+from pysmt.shortcuts import FreshSymbol, Symbol, String, LE, GE, Int, Not, Or, And, Equals, Plus, Solver, StrContains
```
marcogario: These are very long lines, consider breaking them down.

examples/xoxo/xoxo.py
```diff
@@ -0,0 +1,183 @@
...
+
+DEBUG = False
+turns = 0
+sq_size = 3
+test = 'tests/test1.txt'
+test = 'tests/blank.txt'
+
+# create symbols for the board
+board = [[FreshSymbol(INT) for _ in xrange(sq_size)]
+    for _ in xrange(sq_size)]
```
marcogario: indent this to be within the list comprehension.

examples/xoxo/xoxo.py
```diff
@@ -0,0 +1,183 @@
...
+
+solver = Solver(name="z3")
```
marcogario: no need to specify the solver, if you can specify the logic.

examples/xoxo/xoxo.py
```diff
@@ -0,0 +1,183 @@
...
+
+display_board = [[Cell.s for _ in xrange(sq_size)] for _ in xrange(sq_size)]
+
+# load board
+with open(test) as fh:
+    for row, line in enumerate(fh.readlines()):
+        for col, cell in enumerate(line.strip().split(' ')):
+            if cell == Cell.x.name:
+                solver.add_assertion(Equals(board[row][col], Int(Cell.x.value)))
+                display_board[row][col] = Cell.x
```
marcogario: I would separate the two concerns. First, load the board and obtain an up-to-date version of the board. Then you can extract relevant constraints for the solver.

examples/xoxo/xoxo.py
```diff
@@ -0,0 +1,183 @@
...
+    s = 0
+
+DEBUG = False
+turns = 0
+sq_size = 3
+test = 'tests/test1.txt'
+test = 'tests/blank.txt'
+
+# create symbols for the board
+board = [[FreshSymbol(INT) for _ in xrange(sq_size)]
```
marcogario: Any reason why you are not using BitVectors?

examples/xoxo/xoxo.py
```diff
@@ -0,0 +1,183 @@
...
+ch = logging.StreamHandler()
+ch.setLevel(logging.INFO) # change to DEBUG for info on assertions
+logger.addHandler(ch)
+
+class Cell(Enum):
+    x = 1
+    o = 10
+    s = 0
+
+DEBUG = False
```
marcogario: You can use the python's built-in __debug__ .

examples/xoxo/xoxo.py
```diff
@@ -0,0 +1,183 @@
...
+    if(not already_played(row, col)):
+        play_move(Cell.x, row, col)
+    else:
+        logger.info("that cell is already taken")
+        continue
+
+    turns += 1
+
+    # assertions for next move
+    x_next_move_assertion = Equals(Int((turns+1)*Cell.x.value + (turns*Cell.o.value)), Plus(Plus(board[0]), Plus(board[1]), Plus(board[2])))
```
marcogario: Wrap this into a function to make it more clear.

examples/xoxo/xoxo.py
```diff
@@ -0,0 +1,183 @@
...
+    o_win_this_turn_assertions = And(Or(o_win_assertions), this_move_assertion)
+    x_win_next_turn_assertions = And(Or(x_win_assertions), x_next_move_assertion)
+    x_win_this_turn_assertions = And(Or(x_win_assertions), x_this_move_assertion)
+
+    # detect x has won
+    logger.debug("checking if x has won")
+    logger.debug(x_win_this_turn_assertions)
+    if solver.solve([x_win_this_turn_assertions]):
+        print_solver_board()
+        exit("x wins")
```
marcogario: Make a clean exit. I don't think the argument of exit() does anything useful here.

examples/xoxo/xoxo.py
```diff
@@ -0,0 +1,183 @@
...
+        result = get_a_move(Cell.x)
+        if result is not None:
+            play_move(Cell.o, result[0], result[1])
+            continue
+
+    # otherwise try and win
+    logger.debug("looking for a win")
+    played = False
+    for future_turn in range(turns, 5):
+        move_assertion = Equals(Int((future_turn)*(Cell.x.value + Cell.o.value)), Plus(Plus(board[0]), Plus(board[1]), Plus(board[2])))
```
marcogario: Break this into a function to make the expression clear.

marcogario (changes requested):
The example looks interesting. There are a few comments mostly related to making the example more understandable.

From a conceptual point of view, finding a winning strategy is actually a problem that is harder than SAT, and requires quantifiers. I.e., you would like to say that there exists a move for X s.t., for all moves of O, there is a move for X, s.t. for all move for O, there is a move for X that wins.
This is called quantified boolean formula (QBF). pySMT has support for quantifiers, but only few theories are decidable with quantifiers. I think if you use BitVectors, both Z3 and boolector should support quantifiers. Alternatively, there could be a way to encode the problem for using BDDs.

examples/xoxo/xoxo.py
```diff
@@ -0,0 +1,183 @@
+from pysmt.shortcuts import FreshSymbol, Symbol, String, LE, GE, Int, Not, Or, And, Equals, Plus, Solver, StrContains
+from pysmt.typing import INT, STRING
```
marcogario: It does not seem like you are using the theory of STRING. Consider cleaning-up these unused imports.

examples/xoxo/README.md
```diff
@@ -0,0 +1,25 @@
+# Noughts and Crosses player with SAT solver
```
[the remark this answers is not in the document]
mattvenn: Yes

examples/xoxo/xoxo.py
mattvenn: do you mean further indenting the for so it is further indented than the [[ ?

examples/xoxo/xoxo.py
mattvenn: My reasoning was to have a board that contained both players moves (X is 1, O is 10), and had an easy method to:

1. know if a win happened. So if the line adds to 3 or 30.
2. have a way to constrain the solver to look for a move on the next turn. So I can say the total value of the board should be 33, that means 3 turns have been played each.

How could I do this with bitvectors?

examples/xoxo/xoxo.py
mattvenn: I can't find any documentation for a built-in debug - can you link?

examples/xoxo/xoxo.py
mattvenn: ok

examples/xoxo/xoxo.py
mattvenn: done

examples/xoxo/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+uses pySMT and the solver of your choice to play O's and X's. The computer plays Os.
+Player X is given the value 1, and player O is given value 10. 
+This makes it easy to detect number of turns taken as well as if a player has won.
+
+# How it works
```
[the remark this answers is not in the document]
mattvenn: I need to do a bit more research to understand your comment! Might post a question if I'm still stuck.

examples/xoxo/xoxo.py
mattvenn: as in:
x_next_move_assertion = get_x_next_move_assertion(turns)
?
Not sure that makes it any clearer.

examples/xoxo/xoxo.py
mattvenn: same as above? Am I missing something? How would breaking this into a function make it clearer?

examples/xoxo/xoxo.py
marcogario: yes, sorry that was not clear.

examples/xoxo/xoxo.py
marcogario: Arithmetics in BV works the same way, so that you can do the same. Alternatively, you could optimize the representation to be more compact. This needs some more thoughts, but for example, if you have a 4 bits word, you can use to left-most for O and the right-most for X.
11XX is a win for O, XX11 is a win for X. A full game is the sum that yields 1111.
The idea is the same, but now you are thinking in base 10, and you can do the same in base 3.

To check whether base 3 is enough for the other constraints that you have.

examples/xoxo/xoxo.py
marcogario: https://docs.python.org/3/library/constants.html#__debug__

examples/xoxo/xoxo.py
marcogario: ok, I think this can be broken down to make more clear, but it is ok like this if there is no better way.
`` Plus(Plus(board[0]), Plus(board[1]), Plus(board[2])))`` this is basically the sum over the whole board, right? Maybe this can be made more clear?

``Equals(Int((turns+1)*Cell.x.value + (turns*Cell.o.value))`` what is the intuition behind this.

examples/xoxo/xoxo.py
mattvenn: thanks, used __debug__ to switch log levels

examples/xoxo/xoxo.py
mattvenn: using 4 bit length vectors now

mattvenn:
Hi Marco, did you get a chance to review the new bitvect version?

marcogario:
I should have a sometime in the next few days to get through this.

mikand:
Closing in favor of #843

PR #566, opened 2019-03-18
Install swig for yices and picosat in Travis build

makaimann:
It appears that yices and picosat need swig, because the master branch was failing travis on my fork: https://travis-ci.org/makaimann/pysmt/builds/507901966

This just installs swig for those two solvers as well.


mikand:
@makaimann Thanks for this! We probably did not realize this issue because of the Travis cache!

PR #567, opened 2019-03-20
Incremental support for boolector

makaimann:
The latest boolector api supports incremental solving, so this would just add a simple implementation of push/pop and _reset_assertions.

Thanks for all your work on pysmt and let me know if there should be any additional changes for incremental support!


pysmt/solvers/btor.py
```diff
@@ -175,7 +175,10 @@ def __init__(self, environment, logic, **options):
 
     @clear_pending_pop
     def _reset_assertions(self):
-        raise NotImplementedError
+        self.btor = pyboolector.Boolector()
```
marcogario: Can we do better here? Isn't there a way to avoid reconstructing the solver?
For example, Mathsat has a msat_reset_env method, and z3 has a z3_reset method. If not, this is ok.

makaimann: I'm not sure actually -- I don't see an obvious way at least. I'll look into it and get back to you!

makaimann: Looks like there isn't a different way currently, but it's an open issue for them: https://github.com/Boolector/boolector/issues/32.

marcogario: Ok, no action required, then.

pysmt/test/test_solving.py
```diff
@@ -155,6 +155,29 @@ def test_get_py_value(self):
...
+        for name in get_env().factory.all_solvers(logic=QF_BOOL):
+            with Solver(name) as solver:
+                solver.add_assertion(Or(a, b))
+                solver.add_assertion(Or(Not(b), c))
+                self.assertTrue(solver.solve())
+                try:
+                    solver.push(1)
+                except NotImplementedError:
+                    # if push not implemented, pop shouldn't be either
+                    self.assertRaises(NotImplementedError, solver.pop)
```
marcogario: Nice!

pysmt/test/test_solving.py
```diff
@@ -155,6 +155,29 @@ def test_get_py_value(self):
             s.solve()
             self.assertTrue(s.get_py_value(varA))
 
+    @skipIfNoSolverForLogic(QF_BOOL)
+    def test_incremental(self):
```
marcogario: the call to reset is part of the interface for incrementality

PR #568, opened 2019-03-24
Support for Regex in SMT-LIB

marcogario:
Continuation of #478, due to orphaning of source repo.

Open issues:
- Implementation of simplifier
- Testcases


marcogario:
Benchmarks related to strings: http://stringfuzz.dmitryblotsky.com/suites/generated/

Phat3:
Hi!

I should have some test cases around for the regex part. Should I provide only SMT scripts or also the integration with solvers which support that theory?

Which parts need to be done to finish the simplification part?

I got stuck in other works during these months and I forgot to finish my previous PR :sweat_smile: 

Now I have some free time again and I would be happy to help and finish this merge if possible.

marcogario:
If you are up to work on this, that would be awesome!
I did the simple part of the simplification (i.e., no simplification), and I am missing the more interesting one. However, I am not that familiar with the theory of regex as to do it accurately.

The main task for simplification would be to perform the simplification when one or more of the arguments are constants. For example, you can look at how we simplify the arithmetic theories. This would require that we have a way of representing a regex (e.g., using python regex structures, or using a string), and operate on it.

I added a few smtlib files for testing. I need to check whether these include all the new operators or only a subset. 
More detailed examples should go in pysmt/test/examples.py . Those examples are used to test different parts of the system (including simplification, printing, solving etc).

For solvers, we would need to understand which one supports this. I guess z3 and CVC4? But this can be a second step.

xblahoud:
Hi, 

I tried to use this PR to parse my files and found that there is a missing support for `re.allchar`. Also, there were missing parts for parsing the re-constants in `parsing.py`. I've added some support that works on my files in my fork. I am not sure if I implemented those parts correctly. Are you still interested in PR on this?

kubakukis14:
Hi @mikand, could you let me know how I can contribute here? Do I create a separate merge PR into this branch, or would you rather have me work here on this branch?

PR #570, opened 2019-03-31
Fix 320: Z3 segfault - missing context

marcogario:


marcogario:
Tentative fix to the segfault. Tactics would use the global context, but this is not guaranteed to exist. In the tests, we create and destroy contexts, so we might have a segfault because of this.

PR #571, opened 2019-04-06
Upload to PypPi (Test)

marcogario:
This PR introduces 2 changes:
1. The azure pipeline is extended to include uploading to PyPI test server after the build succeeds on master
2. A new way of automatically computing version numbers for dev version.

I would leave this targeting pypi-test for a while, until we are confident that it works as expected.

Notice that the credentials for the pypi test server are stored in azure. See https://docs.microsoft.com/en-us/azure/devops/pipelines/library/service-endpoints?view=azure-devops

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -48,7 +48,7 @@ jobs:
 
 
 - job: 'Ubuntu16All'
-  condition: or(eq(variables['Build.SourceBranchName'], 'master'), eq(variables['Build.Reason'], 'PullRequest'))
+  condition: or(eq(variables['build.sourceBranch'], 'refs/heads/master'), eq(variables['Build.Reason'], 'PullRequest'))
   pool:
     vmImage: 'Ubuntu-16.04'
   strategy:
@@ -98,7 +98,7 @@ jobs:
 
 
 - job: 'Ubuntu16Py27'
-  condition: or(eq(variables['Build.SourceBranchName'], 'master'), eq(variables['Build.Reason'], 'PullRequest'))
+  condition: or(eq(variables['build.sourceBranch'], 'refs/heads/master'), eq(variables['Build.Reason'], 'PullRequest'))
   pool:
     vmImage: 'Ubuntu-16.04'
   strategy:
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -311,3 +311,21 @@ jobs:
...
+- job: PyPi
+  dependsOn:
+    - 'Ubuntu16Py37'
+    - 'Ubuntu16All'
+    - 'Ubuntu16Py27'
+    - 'Ubuntu16PyPy2'
+    - 'Ubuntu16PyPy3'
+    - 'OSX10_13'
+    - 'Win2016'
+  condition: and(succeeded(), eq(variables['build.sourceBranch'], 'refs/heads/master'))
```
mikand: This comment has nothing to do with this PR, but I like your way of expressing the conditions better, so maybe we can change the other conditions in the file as:
```and(succeeded(), or(eq(variables['build.sourceBranch'], 'refs/heads/master'), eq(variables['Build.Reason'], 'PullRequest')))```

File: azure-pipelines.yml [hunk 2]
```diff
@@ -311,3 +311,21 @@ jobs:
...
+  steps:
+  - script: 'pip install twine wheel'
```
mikand: Does the python version on which `pip` is executed matter? In case you need to use `UsePythonVersion@0` task, otherwise ignore this comment.

azure-pipelines.yml
```diff
@@ -311,3 +311,21 @@ jobs:
...
+  - script: './make_distrib.sh; ls dist/'
```
mikand: Inside this script we use `python`, if a specific version is needed we need to use `UsePythonVersion@0` and/or the trick used in `ci/install_unix.sh`

marcogario: It needs to be the version that matches the pip install of wheel, but otherwise it does not matter. The wheel is a source file and should be cross py2-py3.

azure-pipelines.yml
marcogario: AFAIU the ``succeeded``` part only makes sense if we have a dependsOn. I can change the use of sourceBranch instead of sourceBranchName.

PR #573, opened 2019-04-13
Define main function for pysmt as module

marcogario:
Two commands are defined as main function of pysmt

```python -m pysmt install``` 
and 
```python -m pysmt shell```

This removes the bin alias for pysmt-shell, but preserves pysmt-install.

### DIFF
--- a/pysmt/__main__.py
+++ b/pysmt/__main__.py
@@ -1,0 +1,35 @@
+#
+# This file is part of pySMT.
+#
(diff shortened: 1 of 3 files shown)


marcogario:
I would keep as is. The idea was not to replace pysmt-install, but to make this more convenient for certain use cases. I find that doing `python -m nose` is typically more reliable than hoping that nosetest in my path is the right version of python.

PR #575, opened 2019-05-04
Static Version number in package

marcogario:
When running a dev version, we use git to compute the version number. This works ok if we are running pysmt from the git checkout.

However, if we do pip install of a dev version, this can fail. This PR provides a default value for this.
This is more of a temporary fix and a principled solution would be appreciated.

### DIFF
--- a/pysmt/__init__.py
+++ b/pysmt/__init__.py
@@ -18,34 +18,25 @@
 
 VERSION = (0, 8, 1, "dev", 1)
 
-# PEP440 Format
-__version__ = "%d.%d.%d%s%d" % VERSION if len(VERSION) == 5 else \
-              "%d.%d.%d" % VERSION
-
-def git_version():
-    """Human-readable version of latest commit.
-
-    E.g. v0.5.1-4-g49a49f2-wip
-         * 4 commits after tag v0.5.1
-         * Latest commit "49a49f2"
-         * -wip: Working tree is dirty (non committed stuff)
-    See: https://git-scm.com/docs/git-describe
-
-    If the command fails we return __version__
-    """
-    try:
-        import subprocess
-        git_version = subprocess.check_output(["git", "describe", "--dirty=-wip"],
-                                              stderr=subprocess.STDOUT)
-        return git_version.strip().decode('ascii')
-    except subprocess.CalledProcessError:
-        return __version__ # pragma: no cover
-    except OSError:
-        return __version__ # pragma: no cover
-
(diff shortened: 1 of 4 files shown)


pysmt/__init__.py
```diff
@@ -46,6 +46,9 @@ def git_version():
...
 # commits since the release
 if len(VERSION) == 5:
-    commits_from_tag = git_version().split("-")[1]
-    VERSION = VERSION[:4] + (commits_from_tag,)
-    __version__ = "%d.%d.%d%s%s" % VERSION
+    try:
+        commits_from_tag = git_version().split("-")[1]
+    except Exception as ex:
+        commits_from_tag = ""
+        VERSION = VERSION[:4] + (commits_from_tag,)
```
mikand: Doesn't this need to be un-indented?

mikand:
As a principled solution, can't we simply overwrite the `__init__.py` file with simple constants when we do the deployment on pypi?

marcogario:
As suggested. This creates a new `__init__.py` file to hardcode the version number. We could be nicer and just replace the current content, but there is little reason to do so now.

I also removed dependencies from the git_version function from the installer.

marcogario:
I merged this to fix the error on master. Still would like your opinion on the approach. In particular, I think we might be using two different version numbers.

My interpretation of the PEP 440 (https://www.python.org/dev/peps/pep-0440/#pre-release-spelling) is that the development of the version 0.8.1 should be labeled 0.8.1.devX. This is consistent with the way I first created the versioning string in init.py, but not with the git numbering, since git takes a post-release approach, and gives a number based on the previous tag.

I think using 0.8.1.devX clearly identifies that we are working on 0.8.1. I would probably say we can change the script in make_distrib.sh to replace the last digit of VERSION with the number of commits since the tag as obtained from git describe. I.e., even if git describe would provide `0.8.0-X-YYYY` as output, we only consider the value X, and create `0.8.1.devX` as version number.

marcogario:
See commit d8196b for the versioning consistent with the point above. This has not been merged, yet.

mikand:
I agree. So each time we make a release `M.m.v` we tag in git and update the default `VERSION` value in `__init__.py` by setting `M'.m'.v'` and we count the commits in master after the tag creating a version number `M'.m'.v'.devN`. The only question is how do we determine `M'`, `m'` and `v'` before we make the actual changes... So, how do I know if the next iteration will be a major, minor or version update?

Issue #24, opened 2015-03-01
Add examples in documentation and ad-hoc directory


marcogario:
Add the examples used in presentations/demos to the documentation, also as more advance tutorials.

Include the source-codes in an examples directory.

marcogario:
Created directory in examples branch. Documentation still needs to be updated.

marcogario:
Documentation will be revised later. Making pull-request for examples/ dir.

PR #577, opened 2019-05-04
BTOR: Upgrade to f689f

marcogario:
* Upgrade boolector to f689f
* Pass parameters to configure.sh through env variables (see discussion on boolector #24)


marcogario:
For some reason the version we were using does not compile anymore and makes our CI fail. Trying with latest master.

PR #579, opened 2019-05-09
Dev Version Numbering

mikand:
Follow-up of #575 
This PR aims at increasing the version number before appending the `.devX` part to have a correct version numbering in PyPI


mikand:
@marcogario This is OK for me, feel free to merge

PR #583, opened 2019-05-15
Added functions to obtain the symbols in SMTLIB script

mikand:


marcogario:
This looks good to me. I would wait for feedback from the Aina or Liana prior to merging, in case there is a good example that fails with this solution.

mikand:
:+1:

Issue #587, opened 2019-05-30
pysmt not parsing symbols with empty names


4tXJ7f:
pysmt currently does not seem to parse the following SMT-LIB benchmark: https://clc-gitlab.cs.uiowa.edu:2443/SMT-LIB-benchmarks/QF_BV/blob/master/check2/symbols.smt2 . It complains that there is an empty symbol name, presumably `||`.

mikand:
@4tXJ7f Thanks for pointing me to the `symbols.smt2` file, I was unaware of it.

The problem is in https://github.com/pysmt/pysmt/blob/master/pysmt/formula.py#L100-L101 because in pysmt we explicitly forbid empty strings to be symbol names. I honestly do not know if/which solvers support this feature, we need to make some tests.

@marcogario What do you think? should we try to allow empty symbol names in pysmt to be more SMTLIB2 compliant?

4tXJ7f:
@mikand Thanks a lot for the quick reply. Just as a datapoint: CVC4, Boolector, and Z3 all parse (and solve) the file without complaining about empty names.

mikand:
Thanks for the info. On my side I can report that MathSAT does not accept empty strings reporting error "Empty name for symbol". If we want to add support for empty symbols we need to handle this in the msat wrapper.

marcogario:
@mikand thanks for the patch in #588. I am overall ok with it, there are just a few suggestion on the PR.

Overall, I think this is a bad idea from the standard perspective, and honestly hope nobody is creating SMTLIB files with such symbols.

marcogario:
Re-opening this, because I am not happy with the impact of supporting this. Do we have any evidence of a real use case in which this is useful?

mikand:
Implemented on master by f9e27389, merged in #848, as an explicit opt-in, which
addresses the concern raised on this thread about impact:

    from pysmt.environment import get_env
    get_env().allow_empty_var_names = True

With the flag set, `(declare-fun || () Bool)` parses and the symbol works
through the walkers and the solver wrappers. With the flag unset, which is the
default, the error message now names the flag rather than failing opaquely.

This keeps the default strict while making it possible to validate pathological
models such as `QF_BV/check2/symbols.smt2`. Closing.

PR #588, opened 2019-05-31
Support for empty symbol names

mikand:
This PR adds support for symbols having empty name `''`. This is a feature supported by smtlib and requested in #587 .


pysmt/solvers/msat.py
```diff
@@ -47,6 +47,18 @@
 from pysmt.walkers.identitydag import IdentityDagWalker
 
 
+def _fix_var_name(n, back=False):
```
marcogario: Why can't we limit this to the Converter, in the transformation towards mathsat?

pysmt/solvers/msat.py
```diff
@@ -362,9 +374,7 @@ def set_preferred_var(self, var, val=None):
     def print_model(self, name_filter=None):
         if name_filter is not None:
             raise NotImplementedError
-        for v in self.converter.symbol_to_decl.keys():
```
marcogario: Why this change?

marcogario (commented):
In general, I believe giving empty string as a name to a string is non-sense, and the SMTLIB standard is wrong about this.

However, to be complaint with the standard, I agree on this change.

pysmt/test/test_back.py
```diff
@@ -63,6 +63,15 @@ def do_back(self, solver_name, z3_string_buffer=False):
                     # with the SMT-LIB. We might consider extending our
                     # parser.
                     continue
+
+                # The back conversion via SMTLIB of Z3 has a bug
```
marcogario: The issue seems to have been resolved on Z3 side. Can you check? If so, we might consider upgrading.

mikand: I upgraded to Z3 4.8.5, but the solver prints the empty-string symbol as 'null' (but at least it does not crash). I implemented a simple rewriting as for MathSAT.

pysmt/solvers/msat.py
mikand: I noted that there was a misuse of `symbol_to_decl`: sometimes it was used as a map from symbol names (strings) to decls and sometimes as a map from symbols (FNode) to decls. This uniforms the use avoiding some cache miss.

marcogario:
Tests still seem to be failing, maybe this has to do with the upgrade of z3?

pysmt/solvers/z3.py
```diff
@@ -553,14 +553,36 @@ def _back_single_term(self, expr, args, model=None):
...
+            return n
+        else:
+            if len(n) == 0:
+                n = sname
+            return n
+
+    def _substitute_empty_var(self, f):
+        '''Replaces the placeholders for empty var names'''
+        sname = '___PYSMT_EMPTY_STRING___'
+        if False and sname in self.mgr.symbols:
```
marcogario: Deadcode?

mikand: Ops... :smile:

mikand:
Mmmmh, tests are now passing, but I had to implement a rewriting for the back conversion and the new Z3 throws a segfault on my machine. I think the best is to revert the upgrade to Z3...

marcogario:
Do you think the segfault is related to the empty symbol? If so, I would not merge this feature as is. I think we have a very poor argument to allow this, and this can be enabled as an option from the environment. The breaking of the solvers would be up-to the user.

mikand:
No, the segfault seems to be related with the new version of Z3, but I still need to dig into this (I'm a bit busy right now...)

mikand:
supeseded by #848

PR #589, opened 2019-06-04
Fix assertion tracking for boolector

makaimann:
This is a very simple fix for btor to correctly support the `IncrementalTrackingSolver` interface. Currently, getting the assertions from a `BoolectorSolver` always gives a list of `None`, because `_add_assertion` doesn't return anything.

This change just returns `formula` from _add_assertion for tracking in self._assertion_stack.


mikand:
@makaimann Thanks for this fix. I'll merge immediately after #590

PR #591, opened 2019-06-11
Unsat cores from boolector

makaimann:
This pull requests adds support for getting unsat cores from boolector. Named unsat cores are not supported here yet.

### DIFF
--- a/pysmt/solvers/btor.py
+++ b/pysmt/solvers/btor.py
@@ -242,7 +242,7 @@ class BoolectorSolver(IncrementalTrackingSolver, UnsatCoreSolver,
                     unsat_core.add(a)
             return unsat_core
         else:
-            return get_named_unsat_core().values()
+            return self.get_named_unsat_core().values()
 
     def get_named_unsat_core(self):
         """After a call to solve() yielding UNSAT, returns the unsat core as a
--- a/pysmt/solvers/btor.py
+++ b/pysmt/solvers/btor.py
@@ -223,15 +223,21 @@ class BoolectorSolver(IncrementalTrackingSolver, UnsatCoreSolver,
             raise SolverReturnedUnknownResultError
 
     def get_unsat_core(self):
-        unsat_core = set()
-        # relies on this assertion stack being ordered
-        assert isinstance(self._assertion_stack, list)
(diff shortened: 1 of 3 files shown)


mikand:
@makaimann Thanks for this. It seems that there are some conflicts and I cannot merge, can you rebase/merge with the current master?

makaimann:
Yep, no problem! Done.

pysmt/solvers/btor.py
```diff
@@ -212,7 +223,21 @@ def _solve(self, assumptions=None):
...
+            assert isinstance(self._assertion_stack, list)
+            btor_assertions = [self.converter.convert(a) for a in self._assertion_stack]
+            in_unsat_core = self.btor.Failed(*btor_assertions)
+            for a, in_core in zip(self._assertion_stack, in_unsat_core):
+                if in_core:
+                    unsat_core.add(a)
+            return unsat_core
+        else:
+            # get_named_unsat_core currently unsupported
+            raise NotImplementedError("get_named_unsat_core not supported for boolector")
```
mikand: Tests are failing because of this exception (see https://dev.azure.com/micheliandrea/PySMT/_build/results?buildId=201&view=logs&jobId=684d432f-e961-5b78-851d-5ced1c1d1da7). 

We can either:

1. Relax the test accepting a `NotImplementedError` for boolector
2. Wait for https://github.com/Boolector/boolector/issues/23 to be fixed and the use this API for properly implement this
3. Implement the "named unsat core" function on top of this: we track the named assumptions and see which of these are in the "unnamed" core. I believe this should respect the contract of the SMTLIB named core (see definition on page 62 of http://smtlib.cs.uiowa.edu/papers/smt-lib-reference-v2.5-r2015-05-28.pdf)

I think the best would be to do 3 adding a comment about 2 and as soon as the python API becomes available use it instead.

mikand:
@makaimann  Looks great, thanks!

makaimann:
Thanks! I didn't get around to responding to your comments directly, but it looks like you resolved them already. The removal from `msat.py` was just to put the common function `_check_unsat_core_config` into the base class. And good suggestion on implementing `get_named_unsat_core` -- that was pretty straightforward.

PR #598, opened 2019-07-22
Fix mathsat signature for BV_CONCAT

makaimann:
Fixes an issue in the mathsat signature for concat. This was causing issues for `sequence_interpolant` when the interpolant contained `concat` because the signature wouldn't match.


mikand:
:+1: Thanks @makaimann!

PR #614, opened 2019-12-30
Add real to int conversions

shaobo-he:
Specifically, three operations are added,
RealToInt -- The SMT `to_int` function
Ceiling -- Equivalent to the Racket `ceiling` function
Truncate -- Equivalent to the Racket `truncate` function

### DIFF
--- a/pysmt/oracles.py
+++ b/pysmt/oracles.py
@@ -241,6 +241,12 @@ class TheoryOracle(walkers.DagWalker):
         return theory_out
         rtype = formula.symbol_name()
 
+    def walk_realtoint(self, formula, args, **kwargs):
+        #pylint: disable=unused-argument
+        """Extends the Theory with LIRA."""
+        theory_out = args[0].set_lira() # This makes a copy of args[0]
+        return theory_out
+
     @walkers.handles([op.STR_LENGTH, op.STR_INDEXOF, op.STR_TO_INT])
     def walk_str_int(self, formula, args, **kwargs):
         theory_out = self.walk_combine(formula, args, **kwargs)
--- a/pysmt/oracles.py
+++ b/pysmt/oracles.py
@@ -243,8 +243,8 @@ class TheoryOracle(walkers.DagWalker):
 
     def walk_realtoint(self, formula, args, **kwargs):
         #pylint: disable=unused-argument
-        """Extends the Theory with LIRA."""
-        theory_out = args[0].set_lira() # This makes a copy of args[0]
+        """Extends the Theory with NIRA."""
+        theory_out = args[0].set_nira() # This makes a copy of args[0]
         return theory_out
 
     @walkers.handles([op.STR_LENGTH, op.STR_INDEXOF, op.STR_TO_INT])
--- a/pysmt/test/test_lira.py
+++ b/pysmt/test/test_lira.py
@@ -19,6 +19,7 @@ from pysmt.shortcuts import *
 from pysmt.typing import INT, REAL, FunctionType
 from pysmt.test import TestCase, main
 from pysmt.logics import QF_UFLIRA, UFLIRA
+from pysmt.constants import Fraction
 
 class TestLIRA(TestCase):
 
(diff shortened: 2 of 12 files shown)


mikand:
@shaobo-he Thanks for the contribution.

Before accepting this PR, we need to add the new operator also to all other solvers supporting QF_LIRA (mathsat, CVC4 and yices), otherwise the tests will fail.

(note that in your test you specified UFLIRA, instead, you should specify QF_LIRA to target the right fragment and enable all relevant solvers).

Also, other changes are needed, particularly adding examples of RealToInt in tests/examples.py and adding the walk_realtoint to all the walkers (simplifier, rewritings and oracles).

Are you willing to continue the work in this direction? We can of course help you and guide you in this.

pysmt/test/test_lira.py
```diff
@@ -46,6 +47,36 @@ def test_toreal(self):
...
+        b = Symbol('b', INT)
+
+        check = Equals(RealToInt(Real(Fraction(3, 2))), Int(1))
+        check = And(check,
+                    Equals(RealToInt(Real(Fraction(-3, 2))), Int(-2)))
+        check = And(check,
+                    Equals(RealToInt(Real(1)), Int(1)))
+        check = And(check,
+                    Equals(RealToInt(ToReal(b)), b))
+        for sname in get_env().factory.all_solvers(logic=UFLIRA):
```
mikand: the logic here should be 'QF_LIRA' to enable all the relevant solvers.

mikand:
Another minor point: `RealToInt` is a strange name to me, why not calling it `Floor`?

shaobo-he:
> Another minor point: `RealToInt` is a strange name to me, why not calling it `Floor`?

`Floor` indeed makes more sense as it reflects the semantics of this operation. However, `RealToInt` is consistent with the naming of the `to_int` operation in SMTLIB. So I think it's better to keep the latter.

shaobo-he:
> @shaobo-he Thanks for the contribution.
> 
> Before accepting this PR, we need to add the new operator also to all other solvers supporting QF_LIRA (mathsat, CVC4 and yices), otherwise the tests will fail.
> 
> (note that in your test you specified UFLIRA, instead, you should specify QF_LIRA to target the right fragment and enable all relevant solvers).
> 
> Also, other changes are needed, particularly adding examples of RealToInt in tests/examples.py and adding the walk_realtoint to all the walkers (simplifier, rewritings and oracles).
> 
> Are you willing to continue the work in this direction? We can of course help you and guide you in this.

How about me making as many changes as I can and then you can take over to finish it if necessary> Hopefully you don't have to do a lot.

pysmt/test/test_lira.py
shaobo-he: It seems `QF_LIRA` is not defined in `pysmt.logics`. Should I define it?

shaobo-he:
Hi, @mikand. I updated this PR. Please take a look at it again and let me know if it needs any changes.

mikand:
Hi @shaobo-he, thanks for this work and sorry for the latency!

I am unconvinced of the addition of UF_NIRA though: UF_NIRA also contains non-linear terms and mod operators (http://smtlib.cs.uiowa.edu/logics-all.shtml#QF_NIA) that are currently not supported, so I am not comfortable in adding the theory to z3 just for the sake of having the Floor operator implemented.

Also, note that the logic addition code is incomplete: you also need to extend the TheoryOracle in oracles.py to produce non-linearity when it encounters a Floor.

Let me think a bit more on this, but maybe the right way to proceed is to add the mod operator together with the floor to fulfill the requirements of QF_NIA and have a clean implementation of the whole logic?

@marcogario any comment or feedback is welcome.

shaobo-he:
Thank you for your feedback. 

> Also, note that the logic addition code is incomplete: you also need to extend the TheoryOracle in oracles.py to produce non-linearity when it encounters a Floor.

So the changes in https://github.com/pysmt/pysmt/pull/614/files#diff-03f2e786a2b9045a041adfab13dde3a5R244-R249 are not enough? Could you give me some pointers about what should I do here?

shaobo-he:
Hello, @mikand. I'm sorry to bother you again. I was wondering if there's any feedback or decision on this PR. I'd like to see this PR merged so that I can start to work on adding operations of floating-point theory otherwise there'll be conflicts.

pysmt/oracles.py
```diff
@@ -241,6 +241,12 @@ def walk_toreal(self, formula, args, **kwargs):
         return theory_out
         rtype = formula.symbol_name()
 
+    def walk_realtoint(self, formula, args, **kwargs):
+        #pylint: disable=unused-argument
+        """Extends the Theory with NIRA."""
+        theory_out = args[0].set_nira() # This makes a copy of args[0]
```
shaobo-he, line 247: This seems problematic. The output should extend the input with nonlinearity as well as integer arithmetic.

Issue #616, opened 2019-12-31
Correctness bug using an SMTLibSolver


smover:
The SMTLibSolver instantiated with mathsat returns a wrong result when we call 'is_unsat' on the same solver.

The following code reproduces the error:
```python
"""
Example
"""
from fractions import Fraction

from pysmt.typing import REAL
from pysmt.logics import QF_LRA
from pysmt.shortcuts import (
  Solver,
  Implies, And,
  TRUE,
  LT, Equals,
  Real,
  get_env,
  Symbol
)

def get_mathsat_smtlib():
    name = "mathsat-smtlib"
    logics = [QF_LRA]

    env = get_env()
    if not env.factory.is_generic_solver(name):
        path = ["/Users/sergiomover/Tools/mathsat5/build/mathsat"]
        env.factory.add_generic_solver(name, path, logics)

    return Solver(name=name, logic=logics[0])


x1, x2 = [Symbol(var, REAL) for var in ["x1", "x2"]]

init = x1 - Real(1) + x2 < Real(Fraction(1,4))
invar = TRUE()
safe = x1 + x2 < Real(8)

smtlib1 = get_mathsat_smtlib()
smtlib2 = get_mathsat_smtlib()
z3 = Solver(logic=QF_LRA, name="z3")

invar_init = And(invar, init)
iv_imp = Implies(invar, safe)

print("smtlib1, is %s unsat? %d" % (invar_init.serialize(), smtlib1.is_unsat(invar_init)))
print("z3, is %s unsat? %d" % (invar_init.serialize(), z3.is_unsat(invar_init)))
print("smtlib1, is %s valid? %d" % (iv_imp.serialize(),smtlib1.is_valid(iv_imp)))
print("smtlib2, is %s valid? %d" % (iv_imp.serialize(),smtlib2.is_valid(iv_imp)))
print("z3, is %s valid? %d" % (iv_imp.serialize(), z3.is_valid(iv_imp)))
```

The output I get is:
```bash
smtlib1, is (True & (((x1 - 1.0) + x2) < 1/4)) unsat? 0
z3, is (True & (((x1 - 1.0) + x2) < 1/4)) unsat? 0
smtlib1, is (True -> ((x1 + x2) < 8.0)) valid? 1
smtlib2, is (True -> ((x1 + x2) < 8.0)) valid? 0
z3, is (True -> ((x1 + x2) < 8.0)) valid? 0
```

[...]

Issue #615, opened 2019-12-31
No effect setting the 'debug_interaction' option in the SmtLibSolver 


smover:
When setting the `debug_interaction` option on a SmtLibSolver I get no output of the interaction with the solver. The code I use to reporduce the issue is (substitute the path to mathsat):
```python
from pysmt.logics import QF_LRA
from pysmt.shortcuts import (
    Solver,
    LT,
    Real,
    get_env
)

name = "mathsat-smtlib"
path = [<path to mathsat>]
logics = [QF_LRA]
env = get_env()
env.factory.add_generic_solver(name, path, logics)
solver = Solver(name=name, logic=logics[0], solver_options={'debug_interaction':True})
solver.is_valid(LT(Real(3),Real(5)))
```

The test case in (test_generic_wrapper.py)[https://github.com/pysmt/pysmt/blob/451c1bb5f0162963a09298e2b1f29eb93939cdeb/pysmt/test/smtlib/test_generic_wrapper.py] does not seem to test if the debug really generates the output.

It seems that the option is set in a map but then removed immediately from the map:
https://github.com/pysmt/pysmt/blob/f9a5f113125046601d0a2c6f5c1c8755c3bea2d1/pysmt/smtlib/solver.py#L43

mikand:
See #619

PR #619, opened 2020-01-07
Fixed issue #616 and #615

mikand:
The `SmtLibSolver` class was not dealing correctly with the `pending_pop` mechanism.

And the `SmtLibOptions` class was mishandling the `debug_interaction` option.

Thanks to @smover for signaling the issues.


marcogario:
:+1:

PR #624, opened 2020-02-25
Minor Update: update to version 1.6.4

PatrickTrentin88:
NOTES:
- the distributable package of MacOS is not yet ready


mikand:
@PatrickTrentin88 Looks good to me, but pysmt fails to find the installed OptiMathSAT (https://dev.azure.com/micheliandrea/PySMT/_build/results?buildId=259&view=logs&jobId=c7ef3816-649a-5fc9-bd39-698f7c4fc42e&j=c7ef3816-649a-5fc9-bd39-698f7c4fc42e&t=c784bac3-f19d-5f9d-bcea-7b45062cd145)

The check is done by this file: https://github.com/pysmt/pysmt/blob/optimization/pysmt/cmd/check_version.py#L41

I noted that the API file is called optimathsat.py instead of mathsat.py in this PR, so maybe you need to change the imports?

PatrickTrentin88:
> I noted that the API file is called optimathsat.py instead of mathsat.py in this PR, so maybe you need to change the imports?

Quite definitively.

How do I reproduce the error of Azure locally?

mikand:
@PatrickTrentin88 It depends: when optimsat is installed, in which python package can I find the plain SMT functions (e.g. msat_solve)? If they are in the mathsat.py package, then you can simply inherit the MathSAT5Solver class and add the Optimization features, otherwise you have to make the code of  msat.py parametric on the module where to find the functions (e.g. instead of having a MathSAT5Solver class you have a function that takes a module and creates a class specialized with that module). then, you can use the class-generation method to obtain two classes taht use different modules (MathSAT5Solver and OptiMathSat5Solver) and you can inherit from the second one.

Does this make sense?

PatrickTrentin88:
> @PatrickTrentin88 It depends: when optimsat is installed, in which python package can I find the plain SMT functions (e.g. msat_solve)? If they are in the mathsat.py package, then you can simply inherit the MathSAT5Solver class and add the Optimization features, otherwise you have to make the code of msat.py parametric on the module where to find the functions (e.g. instead of having a MathSAT5Solver class you have a function that takes a module and creates a class specialized with that module). then, you can use the class-generation method to obtain two classes taht use different modules (MathSAT5Solver and OptiMathSat5Solver) and you can inherit from the second one.
> 
> Does this make sense?

I was thinking about something along the lines of the following example:

```
#!/usr/bin/env python3

import importlib

class MSATWrapper():
    def __init__(self, *args, **kwargs):
        self.__module = self.get_module()

    def get_module(self):
        return importlib.import_module("mathsat")

    def get_version(self):
        return self.__module.msat_get_version()

    def print_version(self):
        print("MathSAT version:\n\t" + self.get_version())


class OMSATWrapper(MSATWrapper):

    def __init__(self, *args, **kwargs):
        # OMSATWrapper-specific initialization
        super(OMSATWrapper, self).__init__(*args, **kwargs)

    def get_module(self):
        return importlib.import_module("optimathsat")

    def print_version(self):
        print("OptiMathSAT version:\n\t" + self.get_version())


if __name__ == '__main__':

    msat = MSATWrapper()
    msat.print_version()

    omsat = OMSATWrapper()
    omsat.print_version()
```

which yields:

```
MathSAT version:
	MathSAT5 version 5.6.0 (b'0b30bda8e66b') (Jan 22 2020 11:23:43, gmp 6.1.0, gcc 4.8.5, 64-bit)
OptiMathSAT version:
	MathSAT5 version 1.6.4 (5107e2941d30) (Feb  5 2020 18:46:44, gmp 6.1.2, gcc 9.2.1, 64-bit)
```

If the module `mathsat`, that is imported by `PySMT` at the **global** level, is not used
anywhere else, this could be a robust solution.

mikand:
@PatrickTrentin88 This is even better than my solution. The only missing part is a way to check at runtime if mathsat and/or optimathsat are present. Currently we do try-catch on the import of the module (in formula.py and in check_version.py), but we can create a more refined method somewhere and consistently use that method.

PatrickTrentin88:
> @PatrickTrentin88 This is even better than my solution. The only missing part is a way to check at runtime if mathsat and/or optimathsat are present. Currently we do try-catch on the import of the module (in formula.py and in check_version.py), but we can create a more refined method somewhere and consistently use that method.

I see no issue with loading the `mathsat` and `optimathsat` modules in the **global** scope for checking the current version. Actually, I have already changed that section of the code and dropped the `try-catch` solution since now this is no longer necessary.

I have proposed using `importlib` because I don't know any other way to associate a module to an instance variable, so as to use virtual inheritance to our advantage. *[Coming from `C++`, this was my first idea.. I don't know how much pythonic it is, though]*

PatrickTrentin88:
@mikand is the following line in `msat.py`

```
if hasattr(mathsat, "MSAT_EXIST_ELIM_ALLSMT_FM") 
```

still necessary? is it for compatibility with other `mathsat` versions?

PatrickTrentin88:
@mikand I don't understand the naming convention inside `msat.py`:

```
class MSatEnv(MSATWrapper):
class MathSAT5Model(MSATWrapper, Model):
class MathSATOptions(MSATWrapper, SolverOptions):
class MathSAT5Solver(MSATWrapper, IncrementalTrackingSolver, UnsatCoreSolver,
class MSatConverter(MSATWrapper, Converter, DagWalker):
    class MSatQuantifierEliminator(MSATWrapper, QuantifierEliminator, IdentityDagWalker):
    class MSatFMQuantifierEliminator(MSatQuantifierEliminator):
    class MSatLWQuantifierEliminator(MSatQuantifierEliminator):
class MSatInterpolator(MSATWrapper, Interpolator):
class MSatBoolUFRewriter(IdentityDagWalker):
    arguments. This class could implement different rewriting
class MSatSUAOptimizer(MathSAT5Solver, SUAOptimizerMixin):
class MSatIncrementalOptimizer(MathSAT5Solver, IncrementalOptimizerMixin):
```

Some start with `MathSAT5`, some with `MSat`, some others with `MathSAT` ..?

PatrickTrentin88:
@mikand Perhaps it is better to pass the library, or the name of the library, to the class itself as follows:

```
#!/usr/bin/env python3

import importlib

class MSatSolver():

    def __init__(self, library="mathsat"):
        self._mathsat = importlib.import_module(library)
        self._is_opti = (library == "optimathsat")

    def print_version(self):
        print(self._mathsat.msat_get_version())

    def optimize01(self):
        # Alternative 01: safety check
        if not self._is_opti:
            raise Exception("unsupported")
        else:
            print(self._mathsat.MSAT_OPTIMUM)

    def optimize02(self):
        # Alternative 02: no safety check
        print(self._mathsat.MSAT_OPTIMUM)


if __name__ == '__main__':

    msat = MSatSolver()
    msat.print_version()
    # both throw exception
    #msat.optimize01()
    #msat.optimize02()

    omsat = MSatSolver(library="optimathsat")
    omsat.print_version()
    omsat.optimize01()
    omsat.optimize02()

```

In this way, the same unit-tests can be used for both `mathsat` and `optimathsat`. The inheritance-based approach would require one to write the same example using different classes.

This would also let you choose whether you want to make `MathSAT5Solver` inherit from `Optimizer` directly, or maintain the current hierarchy tree.

However, the user is made responsible for ensuring that all `MathSAT` classes are created with the same library. Otherwise, the behavior of the solver is undefined.

PR #627, opened 2020-03-28
add clear pending pop decorator to assertions property of solvers.

EnricoMagnago:


pysmt/solvers/solver.py
```diff
@@ -333,6 +334,7 @@ def last_result(self):
         """
         return self._last_result
 
+    @clear_pending_pop
```
mikand: Test are failing because you used the wrong order for decorators: the right way is:
```python
@property
@clear_pending_pop
def assertions(self):
```

mikand:
Thanks @EnricoMagnago !

Issue #285, opened 2016-07-24
Floating Point Theory 


marcogario:
Floating point is now the only theory from the SMT-LIB that we do not support.

Mathsat supports it, we can use its API to decide which operators to support.

marcogario:
http://smtlib.cs.uiowa.edu/theories-FloatingPoint.shtml

shaobo-he:
Hello, @marcogario May I ask when floating-point theory will be supported? Anytime soon?

shaobo-he:
I'd love to but my Python programming skill only allows me to write trivial programs. I'll let you know if I will do it.

shaobo-he:
Hey @marcogario. Since I got more familiarized with pysmt through one of my recent projects, I'm willing to take the task of adding floating-point support. Please let me know if you or others are already working on it in order to avoid duplicating work.

mikand:
I can confirm nobody is currently working on this, and any contribution is very welcome!
Let us know if you need any support or help

shaobo-he:
Sounds good. I'll submit a PR once I get something working.

miriameisenhofer:
Hello @shaobo-he , if floating-point support is still being worked on, do you have a rough time estimate for when it will be available?

shaobo-he:
> Hello @shaobo-he , if floating-point support is still being worked on, do you have a rough time estimate for when it will be available?

Hello, I'm no longer working on floating point support. Please feel free to close this issue.

PR #632, opened 2020-04-06
[WIP] Add Floating-Point Support

shaobo-he:
This is a working-in-process PR that adds floating-point support to pysmt (#285 )

### DIFF
--- a/pysmt/shortcuts.py
+++ b/pysmt/shortcuts.py
@@ -40,8 +40,11 @@ import pysmt.smtlib.script
 import pysmt.smtlib.printers
 
 # Import types from shortcuts
-from pysmt.typing import INT, BOOL, REAL, BVType, FunctionType, ArrayType, Type
-assert INT or BOOL or REAL or BVType or FunctionType or ArrayType or Type
+from pysmt.typing import INT, BOOL, REAL, BVType, FunctionType, ArrayType, \
+    Type, FPType, RM
+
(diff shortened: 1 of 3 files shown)


marcogario:
> The syntax of creating a floating-point constant node. 

How is this done in SMT-LIB? How is this done in z3? These are typically things we look at to make a decision. Ideally, we want to be non-surprising, but sometimes it might be more ergonomic to change the syntax.

shaobo-he:
> > The syntax of creating a floating-point constant node.
> 
> How is this done in SMT-LIB? How is this done in z3? These are typically things we look at to make a decision. Ideally, we want to be non-surprising, but sometimes it might be more ergonomic to change the syntax.

SMT-LIB's floating-point literal construction function takes 3 bit-vectors (sign, exponent, significand) as arguments. This method allows canonical construction of floating-point values while it is generally very cumbersome for users as they have to convert a floating-point value into three bit-vectors. Therefore, I think pyz3 allows the construction to take floating-point values as arguments (https://z3prover.github.io/api/html/namespacez3py.html#a552c744c21d4d4285c8987f07587995b). Currently, I make the nodes representing floating-point constants to take three bit-vector arguments and the construction function in formula.py also take the same type of arguments. My plan is that once we decide an intuitive API such as taking floating-point values as arguments. I can implement something that normalizes these floating-point values into three bit-vectors.

Issue #631, opened 2020-04-04
Premature optimizations on real division


shaobo-he:
Division by 0 will lead to exception via this line: https://github.com/pysmt/pysmt/blob/master/pysmt/simplifier.py#L982.

shaobo-he:
I would propose we remove the optimization in the `Div` function of `formula.py` and put it in the simplifier.

mikand:
I am not sure this is a premature optimization. Indeed we could raise an exception signaling the problem, but I would avoid the possibility of even creating a formula containing a division by the zero constant.

@shaobo-he do you see any reason to allow this?

shaobo-he:
> I am not sure this is a premature optimization. Indeed we could raise an exception signaling the problem, but I would avoid the possibility of even creating a formula containing a division by the zero constant.
> 
> @shaobo-he do you see any reason to allow this?

If I understand it correctly, division by 0 is well defined by the SMT standards [1] and therefore we should be able to create such terms. For example, the constraint `Equals(Div(Real(1), Real(0)), Real(2))` should be sat instead of causing crashes.

[1] http://smtlib.cs.uiowa.edu/theories-Reals.shtml
```
 "Since in SMT-LIB logic all function symbols are interpreted as total
  functions, terms of the form (/ t 0) *are* meaningful in every 
  instance of Reals. However, the declaration imposes no constraints
  on their value. This means in particular that 
  - for every instance theory T and
  - for every value v (as defined in the :values attribute) and 
    closed term t of sort Real,
  there is a model of T that satisfies (= v (/ t 0)).
 "
```

marcogario:
This is mostly motivated from trying to normalize expressions. In LRA we cannot have general division or multiplication, except for the cases where we have constant coefficients. For those cases, however, the denominator must be non-zero. In those cases the rewriting is correct.

```
:extensions
 "Terms with _concrete_ coefficients are also allowed, that is, terms
  of the form c, (* c x), or (* x c)  where x is a free constant and 
  c is an integer or rational coefficient. 
  - An integer coefficient is a term of the form m or (- m) for some
    numeral m.
  - A rational coefficient is a term of the form d, (- d) or (/ c n) 
    for some decimal d, integer coefficient c and numeral n other than 0.
 "
```

To support non-linear logics in which there is a division by zero, I would suggest we replace the expression with a fresh symbol if the constant is zero. This would also simplify validating a model against the expression since the solver would return a satisfying value for the term.

[...]

PR #636, opened 2020-04-13
Support for division by Zero

marcogario:
* Check if we are trying to create an expression `Div(x, Real(0))`, if so, raise a User warning, if the configuration flag `enable_div_by_0`.
* Check for div by 0 in simplifier
* Extend TheoryOracle to return a non-linear theory for expressions with division by 
---
*Old Proposal*
When creating a Div expression, we check whether the right hand side is a 0. If so, we replace the entire expression with a new variable. In this way, we preserve the SMTLIB semantics of Div, while keeping the model extraction and other operations simple.

Since getting a result due to a division by zero is always surprising, we add a user warning when performing the translation and make it possible to disable it with the configuration variable `Environment.replace_div_by_0`

See discussion in https://github.com/pysmt/pysmt/issues/631

### DIFF
--- a/pysmt/environment.py
+++ b/pysmt/environment.py
@@ -74,6 +74,7 @@ class Environment(object):
         self._factory = None
         # Configurations
         self.enable_infix_notation = False
+        self.enable_div_by_0 = True
 
         # Dynamic Walker Configuration Map
         # See: add_dynamic_walker_function
--- a/pysmt/formula.py
+++ b/pysmt/formula.py
@@ -33,6 +33,7 @@ if sys.version_info >= (3, 3):
 else:
     from collections import Iterable
 
+import warnings
 from six.moves import xrange
 
 import pysmt.typing as types
@@ -261,7 +262,11 @@ class FormulaManager(object):
 
     def Div(self, left, right):
         """ Creates an expression of the form: left / right """
-        if right.is_constant(types.REAL):
+        if right.is_constant(types.REAL, 0) and self.env.enable_div_by_0:
+            # Allow division by 0 byt warn the user
+            # This can only happen in non-linear logics
+            warnings.warn("Warning: Division by 0")
+        elif right.is_constant(types.REAL):
             # If right is a constant we rewrite as left * 1/right
             inverse = Fraction(1) / right.constant_value()
             return self.Times(left, self.Real(inverse))
--- a/pysmt/test/test_solving.py
+++ b/pysmt/test/test_solving.py
@@ -240,6 +240,34 @@ class TestBasic(TestCase):
             self.assertEqual(satisfiability, s, f)
 
     def do_model(self, solver_name):
+        def assert_model(f, subs):
+            simp = f.substitute(subs).simplify()
+            if not simp.is_true() and get_env().enable_div_by_0:
+                free_vars = simp.get_free_variables()
+                self.assertTrue(len(free_vars) == 0)
+
+                # Models might not be simplified to a constant value
+                # if there is a division by zero. We find the
+                # division(s) and ask the solver for a replacement
+                # expression.
+                stack = [simp]
+                div_0 = []
+                while stack:
+                    x = stack.pop()
+                    if x.is_constant():
+                        pass
+                    elif x.is_div() and x.arg(1).is_zero():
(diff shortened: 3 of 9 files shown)


shaobo-he:
I'm afraid that replacing division by zero with a fresh variable is problematic. Consider the following Python program and its corresponding SMT-LIB program,

```Python
from pysmt.shortcuts import Symbol, And, is_sat, Div, Equals, Not, Real
from pysmt.typing import REAL

varA = Symbol('A', REAL)
varB = Symbol('B', REAL)

f = And(Equals(varA, varB),
        Not(Equals(Div(varA, Real(0)), Div(varB, Real(0)))))

print is_sat(f)
```

```SMT
(set-logic QF_NRA)
(declare-const varA Real)
(declare-const varB Real)

(assert (= varA varB))
(assert (not (= (/ varA 0) (/ varB 0))))

(check-sat)
```
`is_sat` returns `True` for the Python program while the SMT constraints are unsat.

marcogario:
You are right. This should be an UF such as `Div0(x)` . This still makes the model reading from the solver a bit more annoying but should still be doable in Z3.

marcogario:
We might consider also just leaving the term `Div(x, 0)`. The main reason I wanted a different symbol was to make it more obvious when processing. Since it becomes a UF, we can treat it uniformly as a UF. However, I would be curious to see the alternative approach and compare them.

mikand:
I like the warning idea: I find it quite strange that a user really wants to create a division by **constant** 0. I think the SMT-Lib interpretation is more for the mathematical beauty of having a total div function than for practical reasons. 

I am not sure to understand what you mean by "The main reason I wanted a different symbol was to make it more obvious when processing": if we leave a Div(x, Real(0)), after throwing a warning, I think that everything should work except for the simplifier (that will require a very simple fix). Do you see any other problematic point?

marcogario:
> Do you see any other problematic point?

Model extraction. In this PR we have a more general approach to read an assignment to a ground function. If we leave the Div(x, 0), we need to special case it. That is probably still ok, though. 

> I find it quite strange that a user really wants to create a division by constant 0.

I agree. I think the main issue is when the solver returns a value that leads to division by zero. That is currently a bug on our side. I do not think, however, that it makes sense to forbid this to the user altogether.

pysmt/formula.py
```diff
@@ -261,7 +262,11 @@ def Pow(self, base, exponent):
 
     def Div(self, left, right):
         """ Creates an expression of the form: left / right """
-        if right.is_constant(types.REAL):
+        if right.is_constant(types.REAL, 0) and self.env.enable_div_by_0:
+            # Allow division by 0 byt warn the user
```
marcogario, line 266:
```suggestion
            # Allow division by 0 but warn the user
```

marcogario:
@mikand Would you mind reviewing this again? I changed it as discussed and rebased it against the release branch. CI is taking forever, so apologies if any test fails in the meanwhile.

pysmt/test/test_solving.py
```diff
@@ -240,6 +240,34 @@ def test_examples_btor(self):
             self.assertEqual(satisfiability, s, f)
 
     def do_model(self, solver_name):
+        def assert_model(f, subs):
```
mikand, line 243: Would it make sense to have this code as a service of pysmt?

Something like `check_model(formula, dict_like_object)`?

pysmt/environment.py
```diff
@@ -74,6 +74,7 @@ def __init__(self):
         self._factory = None
         # Configurations
         self.enable_infix_notation = False
+        self.enable_div_by_0 = True
```
mikand, line 77: This flag has a strange name: if `enable_div_by_0` is True, we allow the div by zero and we throw a warning, if it is False, we allow the div by zero and we do not raise the warning...

Maybe we can call it `warn_if_div_by_zero`?

marcogario: The idea of the configuration flag was mostly to disable this feature. If this is set to false, there would not be any special handling of the division by zero, and thus we would throw a DivisionByZero exception.

I am also ok with always allowing this behavior and only control whether we throw a warning or not (change this to `warn_if_div_by_zero`.

PR #639, opened 2020-04-17
Optimization Modulo Theory improved API and support

arrighi79:


pysmt/optimization/goal.py
```diff
@@ -0,0 +1,51 @@
+from pysmt.shortcuts import Times, Int
```
mikand, line 1: Missing header: all PySMT files need to have the standard header (just as the one in `__init__.py`)

pysmt/optimization/goal.py
```diff
@@ -0,0 +1,51 @@
...
+
+"""
+Abstract class for goals
+"""
+class Goal:
```
mikand, line 6: Duck typing: there is no need for this class

pysmt/optimization/goal.py
```diff
@@ -0,0 +1,51 @@
...
+"""
+class MaximizationGoal(Goal):
+    def __init__(self, formula):
+        self.formula = formula
+
+    def term(self):
+        return self.formula
+
+    def minimize_term(self):
+        return Times(Int(-1), self.formula)
```
mikand, line 21: This needs refinement.
- It's OK if formula is an integer-type
- It would be better to use `Real(-1)` if formula is a real-type 
- If formula is a bit-vector-type you need to use `BVSub(BVZero, formula)`
- you need to accept an environment at the constructor in case a user wants to specify it, and by default use the formula_manager in the default environment (have a look at any solver for an example of this)

mikand: @PatrickTrentin88 I am not 100%sure about bit vectors: the idea here would be to transform a maximization into a minimization to simplify the implementation of solver-dependent code.

Two questions:
1. Does this transformation impose any performance penalty / do you see limitations in this?
2. Is the transformation above correct for signed and unsigned bit-vectors?

pysmt/optimization/optimizer.py
```diff
@@ -168,15 +168,15 @@ def optimize(self, cost_function, strategy='ub',
             else:
                 # Check if there is a model </<= bound
                 check_result = self._optimization_check_progress(client_data,
-                                                                 cost_function,
+                                                                 goal.minimize_term(),
```
mikand, line 171: You need to check that goal is a Minimization or Maximization goal, otherwise throw a meaningful exception

pysmt/optimization/goal.py
PatrickTrentin88: @mikand  Such transformation does not appear to be type-correct in the case of a Floating-Point or Bit-Vector goal, since it is mixing LA (`Times`, `Int`) with a `formula` of an arbitrary type. 

I don't see the need to rewrite maximization as minimization when both `z3` and `OptiMathSAT` have native support maximization.

Even if one wanted to implement their own optimization procedures on top of a regular SMT solver, supporting both requires changing only the constraint learned at each iteration. 

What am I not seeing?

pysmt/optimization/goal.py
```diff
@@ -0,0 +1,51 @@
...
+MaxSMT goal common to all solvers.
+Attention: some solvers may not support this goal
+"""
+class MaxSMT(Goal):
+
+    """Accepts soft clauses and the relative weights"""
+    def __init__(self, clause, weight):
+        self.soft = []
+        for x in range(min(len(clause), len(weight))):
+            self.soft.append((clause.pop(0), weight.pop(0)))
```
PatrickTrentin88, line 47: My understanding is that here `clauses` and `weights` are two lists of terms.

I am no python expert, but why not

```
self.soft = zip(clause, weight)
```

instead of 

```
        for x in range(min(len(clause), len(weight))):
            self.soft.append((clause.pop(0), weight.pop(0)))
```

pysmt/optimization/goal.py
```diff
@@ -0,0 +1,51 @@
...
+
+    """Accepts soft clauses and the relative weights"""
+    def add_soft_clause(self, clause, weight):
+        self.soft.append((clause.pop(0), weight.pop(0)))
```
PatrickTrentin88, line 51: My understanding is that here `clause` and `weight` are simple terms. What is `pop(0)` doing here?

pysmt/optimization/goal.py
mikand: You are missing nothing :)

I was thinking of starting by forcing the use of minimization consistently to preserve the simplicity of the optimizers "from outside" (the ones that incrementally search for better solutions) that currently support only minimization, but maybe it is better di directly support minimization **and** maximization without converting one into the other.

pysmt/optimization/goal.py
mikand: You are right (I honestly skipped this class because for now we are working with simple minimization and maximization).

pysmt/optimization/goal.py
mikand: You are correct.

pysmt/optimization/goal.py
PatrickTrentin88: @mikand @arrighi79

Unless in `pysmt` there is a universal `LT` constraint, which translates to the appropriate SMT-LIB operator given the type of its arguments, the simplicity is already lost when one handles objectives of different types. 

How about:

```
    def make_better_than(goal, value):
        """returns (goal < value) when minimizing and
            (value < goal) when maximizing"""
        ...
```

The method would also be responsible for checking for the type of `goal.formula` and use the SMT-LIB operator of the right type.

Then the incremental optimization search becomes completely agnostic about the internal details of the `Goal` instance(s) it is handling.

pysmt/optimization/optimizer.py
```diff
@@ -124,15 +124,15 @@ def _binary_bound(self, lb, ub):
             u = lb + abs(lb) + 1
         return (l + u + 1) // 2
 
-    def _bound(self, lb, ub, strategy):
+    def _bound(self, lb, ub, strategy, goal):
         if strategy == 'binary':
-            return self._binary_bound(lb, ub)
+            return self._binary_bound(lb, ub) #ok
         if strategy == 'ub':
```
mikand, line 130: After this refactoring we shall call this strategy `linear` instead of `ub` (because maximization works by improving lower bounds, not upper bounds).

pysmt/optimization/optimsat.py
```diff
@@ -101,9 +80,19 @@ def _le(self, x, y):
...
-        msat_obj = self._msat_lib.msat_make_minimize(self.msat_env(), obj_fun, None, None, False)
+
+        if goal.is_minimization_goal():
+            make_fun = self._msat_lib.msat_make_minimize
+        elif goal.is_maximization_goal():
+            make_fun = self._msat_lib.msat_make_maximize
+        else:
+            raise GoalNotSupportedError("optimathsat", goal)
+
+        msat_obj = make_fun(self.msat_env(), obj_fun, None, None, False)
```
mikand, line 94: :+1: for the use of functional assignment

pysmt/optimization/goal.py
```diff
@@ -0,0 +1,152 @@
...
+        return True
+
+
+class MinMaxGoal(Goal):
+
+    def __init__(self, terms):
+        self.term = terms
+
+    def add_terms(self, *terms):
+        self.term.extend(list(terms))
```
mikand, line 117: I think that also an `add_term` function could be useful (possibly more useful than this)

pysmt/optimization/optimizer.py
```diff
@@ -151,34 +202,190 @@ def optimize(self, cost_function, strategy='ub',
...
+            self.pop()
+        rt = None
+        self.push()
+        t = goal.terms()[0]
+        le = self._get_le(t)
+        _or = self._get_or(t)
+
+        symbol_type = self._get_symbol_type(t)
+
+        _private_symbol = Symbol("_private_symbol", symbol_type)
```
mikand, line 262: Pysmt has a `FreshSymbol` function to create special symbols safely.

pysmt/optimization/optimizer.py
```diff
@@ -151,34 +202,190 @@ def optimize(self, cost_function, strategy='ub',
...
+
+        for term in goal.terms():
+            self.push()
+            part_res.append(self._optimize_min(MinimizationGoal(term),
+                                          strategy, feasible_solution_callback,
+                                          step_size, **kwargs)[1]
+                            )
+            self.pop()
+        rt = None
+        self.push()
```
mikand, line 255: This part, should go inside the mixins, not here!

pysmt/optimization/optimizer.py
```diff
@@ -151,34 +202,190 @@ def optimize(self, cost_function, strategy='ub',
...
+
+    def _optimize_minmax(self, goal, strategy,
+                 feasible_solution_callback,
+                 step_size, **kwargs):
+
+        part_res = []
+
+        for term in goal.terms():
+            self.push()
+            part_res.append(self._optimize_min(MinimizationGoal(term),
```
mikand, line 249: Nope. This is not the intended semantics!
have a look here (http://disi.unitn.it/rseba/papers/cav15_extended.pdf, page 4):
Screenshot from 2020-05-04 10-25-47

mikand: Note also that pysmt has a `Max` and a `Min` operator that symbolically express the maximum and the minimum value among a gven, finite set. This could be useful for a simple initial implementation...

pysmt/optimization/optimizer.py
```diff
@@ -151,34 +202,190 @@ def optimize(self, cost_function, strategy='ub',
...
+        _or = self._get_or(t)
+
+        symbol_type = self._get_symbol_type(t)
+
+        _private_symbol = Symbol("_private_symbol", symbol_type)
+        f = []
+        for res in part_res:
+            f.append(le(_private_symbol, res))
+
+        self.add_assertion(_or(f))
```
mikand, line 267: Nope! you cannot have `_private_symbol` to be smaller than at least one element, you need it to be smaller than all the elements and to be equal to al least one of them to be the minimum....

pysmt/optimization/optimizer.py
```diff
@@ -87,6 +89,55 @@ def _comparation_functions(self, objective_formula):
...
+        if otype.is_int_type():
+            return INT
+        elif otype.is_real_type():
+            return REAL
+        elif otype.is_bv_type():
+            return BVType(otype.width)
+        else:
+            raise PysmtValueError("Invalid optimization function type: %s" % otype)
+
+    def _get_or(self, objective_formula):
```
mikand, line 103: There is no need for this function: the logical disjunction is always `Or`, `BVOr` is the bit-wise disjunction and you do not need it.

pysmt/optimization/optimizer.py
```diff
@@ -87,6 +89,55 @@ def _comparation_functions(self, objective_formula):
...
+            return mgr.Or
+        elif otype.is_real_type():
+            return mgr.Or
+        elif otype.is_bv_type():
+            return mgr.BVOr
+        else:
+            raise PysmtValueError("Invalid optimization function type: %s" % otype)
+
+
+    def _get_le(self, objective_formula):
```
mikand, line 116: You already have this in the mixins, the reason why you added it here is because you added the push/pop code in the wrong place

pysmt/optimization/z3.py
```diff
@@ -0,0 +1,94 @@
...
+        cost_function = goal.term()
+        obj = self.converter.convert(cost_function)
+
+        h = None
+        if goal.is_minimization_goal():
+            h = self.z3.minimize(obj)
+        elif goal.is_maximization_goal():
+            h = self.z3.maximize(obj)
+        else:
+            raise GoalNotSupportedError("z3", goal.__class__)
```
mikand, line 55: This does not support the maxmin nor minmax goals!

pysmt/test/test_optimization.py
```diff
@@ -21,45 +21,97 @@
...
+
+        f21 = GE(y, Int(11))
+        f22 = LE(y, Int(14))
+
+        f31 = GE(z, Int(15))
+        f32 = LE(z, Int(18))
+
+        minmax = MinMaxGoal([x, y, z])
+
+        optimizers = ["z3_sua", "z3_incr", "msat_sua", "msat_incr", "optimsat"]
```
mikand, line 72: Nope. If you do like this this test will fail if some of the solvers are not installed. you can do like this:
```
for oname in get_env().factory.all_optimizers(logic=QF_LIA):            
    with Optimizer(name=oname) as opt:
```
you can use an if over `oname` if you want to skip solvers, but here it should not be needed

pysmt/optimization/optimizer.py
PatrickTrentin88: > Nope. This is not the intended semantics!
> have a look here (http://disi.unitn.it/rseba/papers/cav15_extended.pdf, page 4):
> Screenshot from 2020-05-04 10-25-47

Just a comment to repeat what I stated in the email: the above definition is for Max-Min, the one for Min-Max is dual.

PatrickTrentin88: > Note also that pysmt has a `Max` and a `Min` operator that symbolically express the maximum and the minimum value among a gven, finite set. This could be useful for a simple initial implementation...

The above definition is more convenient than using a `max`/`min` operator, because the objective function is always *explicitly* upper-bounded (or lower-bounded, depending on the definition) by the terms in the given list of arguments.

pysmt/optimization/optimizer.py
```diff
@@ -138,6 +140,93 @@ def _get_lt(self, objective_formula):
...
+            t2 = x
+        else:
+            t1 = x
+            t2 = goal.term()
+        return lt(t1, t2)
+
+    def linear_search_cut(self):
+        """must be called always"""
+        if self._obj.is_minimization_goal():
+            bound = self._upper
```
PatrickTrentin88, line 179: Or it can just remember the last model value (see changes in the 2nd email).
Either is fine to me.

pysmt/optimization/optimizer.py
```diff
@@ -138,6 +140,93 @@ def _get_lt(self, objective_formula):
...
+
+    def _compute_pivot(self):
+        if self._lower is None and self._upper is None:
+            return None
+        l,u = self._lower, self._upper
+        if self._lower is None and self._upper is not None:
+            l = self._upper - (abs(self._upper) + 1)
+        elif self._lower is not None and self._upper is None:
+            u = self._lower + abs(self._lower) + 1
+        return (l + u + 1) // 2
```
PatrickTrentin88, line 193: 1. Is `(l + u + 1) // 2` type-correct in SMT-LIB, given that `l`,`u` can be of Int/Real/BV type?

2. If `obj.term` is of Int type, one should probably round the result of the division.

3. The `pivot` value should be validated before returning it. It should be discarded
    when it is _smaller or equal `self._lower` (minimization)_
    or _larger or equal `self._upper` (maximization)_.

4. `binary_search_cut()` expects a valid `pivot` all the time, but the current implementation
     of this method does not guarantee that. A safe pivoting value for minimization is `self._upper`
     and for maximization is `self._lower`. _(i.e. it becomes a linear step)_

pysmt/optimization/optimizer.py
```diff
@@ -138,6 +140,93 @@ def _get_lt(self, objective_formula):
...
+
+    def empty(self):
+        """True: remove this unit from optimization search"""
+        return self._upper <= self._lower
+
+
+    def search_is_sat(self, model):
+        self._pivot = None
+        model_value = model.eval(self._obj.term())
+        if self._obj.minimize and self.is_better_than(model_value, self._upper):
```
PatrickTrentin88, line 209: Remember to implement `is_better_than`. @mikand can probably help suggesting how to do it.

pysmt/optimization/optimizer.py
```diff
@@ -151,49 +284,118 @@ def optimize(self, cost_function, strategy='ub',
...
         self._cleanup(client_data)
-        return last_model, cast(ub)
+        if goal.is_minimization_goal():
+            t = ub
+        else:
+            t = lb
+        return last_model, cast(t)
+
+
+class SUAOptimizerMixin(ExternalOptimizerMixin):
```
mikand, line 384: Why did you duplicate this class?

pysmt/optimization/optimizer.py
```diff
@@ -151,49 +284,118 @@ def optimize(self, cost_function, strategy='ub',
...
+        first_step = True
+        while not current.empty():
+            if not first_step:
+                if strategy == "linear":
+                    lin_assertions = current.linear_search_cut(step_size)
+                elif strategy == "binary":
+                    lin_assertions = current.binary_search_cut()
+                else:
+                    raise PysmtValueError("Unknown optimization strategy '%s'" % strategy)
+                status = self.solve(assumptions=[lin_assertions])
```
mikand, line 312: If you want to maintain the distinction between SUAOptimizers and Incremental optimizers, this part needs to be delegated to a mixin function

pysmt/optimization/optimizer.py
```diff
@@ -151,49 +284,118 @@ def optimize(self, cost_function, strategy='ub',
...
+                first_step = False
+            if status:
+                model = self.get_model()
+                current.search_is_sat(model)
+            else:
+                current.search_is_unsat()
+
+
+
+    def _optimize_max_min(self, goal, strategy,
```
mikand, line 324: Is this still needed?

pysmt/optimization/optimizer.py
PatrickTrentin88: @mikand 

1. What is the concrete model value of a real variable? An infinite precision fraction or a float?

2. agreed

3./4. The previous implementation returned `None`, which was not valid. Now for Int/BV it appears to be correct. For Reals, after removing `+1`, it might depend on the type of `l,u`. I wouldn't trust the result of `(l + u)/2` to return some value in the middle for every pair of `float` values. I would trust it if `l,u` are represented as infinite precision fractions.

mikand: @PatrickTrentin88 

in pysmt we use infinite-precision fractions (mpq_t from GMP if pygmp is installed otherwise `fractions.Fraction` from the standard library) for rationals and plain integers (that in python are arbitrary-precision) for INTs (and also as internal representation of BVs).

@arrighi79 Be careful in using these types for lower and upper bounds. You can also use assertions to check them.

PatrickTrentin88: > @PatrickTrentin88
> 
> in pysmt we use infinite-precision fractions (mpq_t from GMP if gmpy2 is installed otherwise `fractions.Fraction` from the standard library) for rationals 

Out of curiosity, why do you use `mpq_t` if available and `fractions.Fraction` otherwise? It looks like they are interchangeable, since you can swap one for the other, so why bother?

mikand: Math operations with gmpy are WAY faster than with fractions.Fraction. This is negligible if you use the python representations just for representation, but if you use them in algorithms and you do computation the difference in terms of performance is huge.

pysmt/optimization/optimizer.py
arrighi79: no, but until I finish the implementation of the new classes I prefer not to remove the old ones

pysmt/optimization/optimizer.py
arrighi79: one of the side effects of the rebase

pysmt/optimization/optimizer.py
```diff
@@ -141,33 +129,63 @@ def _get_lt(self, objective_formula):
...
-    def __init__(self, goal, lt, le, client_data):
+    def __init__(self, goal, environment,client_data):
         self._obj = goal
         self._lower = None #-INF where i found this costant?
         self._upper = None #+INF
         self._pivot = None
-        self._lt = lt
-        self._le = le
+        self.environment = environment
+        self._cast, self.op_strict, self.op_ns = self._comparation_functions(goal)
```
PatrickTrentin88, line 138: `op_strict` and `op_ns` are not quite quite self-explaining names.

pysmt/optimization/goal.py
```diff
@@ -77,6 +93,9 @@ def __init__(self, formula):
         """
         self.formula = formula
 
+    def opt(self):
+        return MaximizationGoal
```
PatrickTrentin88, line 97: Are we returning a class here? Is this the pythonic way to do it? *(just curious)*

mikand:
@arrighi79 Note that I merged a patch to the pysmt parser that supports OMT format. We will need to make a bridge from that to our solvers. Right now, you probably have to rebase/marge your branch to avoid conflicts...

pysmt/test/test_optimization.py
```diff
@@ -21,61 +21,304 @@
...
+        formula = GT(x, Real(10))
+        min = MinimizationGoal(x)
+        for oname in get_env().factory.all_optimizers(logic=QF_LRA):
+            with Optimizer(name=oname) as opt:
+                if opt.can_diverge_for_unbounded_cases():
+                    continue
+                opt.add_assertion(formula)
+                with self.assertRaises(PysmtUnboundedOptimizationError):
+                    opt.optimize(min)
+
```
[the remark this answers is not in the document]
mikand: Note, that it seems to me that these function are not implemented in all the solvers...

pysmt/test/test_optimization.py
```diff
@@ -21,61 +21,328 @@
...
+        obj5 = MinimizationGoal(u)
+        for oname in get_env().factory.all_optimizers(logic=QF_LIA):
+            with Optimizer(name=oname) as opt:
+                try:
+                    opt.add_assertion(f1)
+                    opt.add_assertion(f2)
+                    model, values = opt.lexicographic_optimize([obj1, obj2, obj3, obj4, obj5])
+                    self.assertTrue(model is None)
+                    self.assertTrue(values is None)
+                except NotImplementedError:
```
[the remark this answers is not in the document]
mikand: Also in the other functions this is copy-pasted

pysmt/optimization/optimizer.py
```diff
@@ -0,0 +1,541 @@
...
+        This function can raise a PysmtUnboundedOptimizationError if
+        the solver detects that the optimum value is either positive
+        or negative infinite or if there is no optimum value because
+        one can move arbitrarily close to the optimum without reching
+        it (e.g. "x > 5" has no minimum for x, only an infimum)
+
+        """
+        raise NotImplementedError
+
+    def pareto_optimize(self, goals):
```
mikand, line 71: This function is duplicated!

pysmt/optimization/optimizer.py
```diff
@@ -0,0 +1,541 @@
...
+        it (e.g. "x > 5" has no minimum for x, only an infimum)
+
+        """
+        raise NotImplementedError
+
+
+
+    def lexicographic_optimize(self, goals):
+        """
+        This function return a pair of (model, values) where 'values' is a list of result for each goal in 'goals'
```
mikand, line 97:
```suggestion
        This function returns a pair of (model, values) where 'values' is a list containing the optimal values (as pysmt constants) for each goal in 'goals'
```

pysmt/optimization/optimizer.py
```diff
@@ -0,0 +1,541 @@
...
+        If there aren't any solution the function return a pair (None,None)
```
mikand, line 98:
```suggestion
        If there is no solution the function returns a pair (None,None)
```

pysmt/optimization/optimizer.py
```diff
@@ -0,0 +1,541 @@
...
+
+        The parameter 'goals' must be a list of 'Goals'(see file goal.py).
+
+        The lexicographic method consists of solving a sequence of single-objective optimization problems. The order of
+        problems are important because the result of previews goals become a costraint for subsequent goals
```
mikand, line 103: previews?

pysmt/optimization/optimizer.py
```diff
@@ -0,0 +1,541 @@
...
+
+        For some implemented examples see file pysmt/test/test_optimization.py
+        """
+        raise NotImplementedError
+
+
+    def boxed_optimize(self, goals):
+        """
+        This function return dictionary where the keys are the goals of optimization and the values are a pair
```
mikand, line 112:
```suggestion
        This function returns dictionary where the keys are the goals of optimization. Each goal is mapped to a pair
```

pysmt/optimization/optimizer.py
```diff
@@ -0,0 +1,541 @@
...
+        (model, value) of the current goal (key).
+        If there aren't any solution the function return None
```
mikand, line 114:
```suggestion
        If there is no solution the function returns None
```

pysmt/optimization/optimizer.py
```diff
@@ -0,0 +1,541 @@
...
+
+        The parameter 'goals' must be a list of 'Goals'(see file goal.py).
+
+        The boxed method consists of solving a list of single-objective optimization problems independent of
```
mikand, line 118:
```suggestion
        The boxed method consists of solving a list of single-objective optimization problems independently of
```

pysmt/optimization/optimizer.py
```diff
@@ -0,0 +1,541 @@
...
+        each other. The order of problems isn't important because the result of previews goals don't change
```
mikand, line 119: previews?

PR #640, opened 2020-04-25
OMT parsing

cespio:
Hi all, this is my first pull request for this project. I have talked in advance with @mikand  about the idea to include my master thesis work  (https://github.com/cespio/omt2mzn) in this project

This first pull request wants to extend the parsing capabilities of pysmt in order to understand the following SMTLIB-v2 extension http://optimathsat.disi.unitn.it/pages/smt2reference.html

The other pull requests will be more focused on the actual omt2mzn translation

PR #642, opened 2020-05-05
OMT parsing

mikand:
New version of #640 
This has been rebased on the optimization branch.

http://optimathsat.disi.unitn.it/pages/smt2reference.html


pysmt/smtlib/parser/parser.py
```diff
@@ -1128,6 +1150,42 @@ def _cmd_assert(self, current, tokens):
         self.consume_closing(tokens, current)
         return SmtLibCommand(current, [expr])
 
+    def _cmd_assert_soft(self, current, tokens):
```
marcogario, line 1153: I think this can be made more readable by not generalizing the handling of the annotations.

```python
        term_weight = None
        term_group_id = None
        curr = tokens.consume()
        while curr != ")":
            tokens.add_extra_token(curr)
            curr_parse = self.parse_atom(tokens, current)
            if curr_parse == ":weight" and term_weight is None:
                term_weight = self.get_expression(tokens)
            elif curr_parse == ":id" and term_group_id is None:
                term_group_id = self.parse_atom(tokens, "assert-soft")
            else:
                raise NotImplementedError("The current option is not currently implemented %s "
                                          "or some of them are repeated" % curr_parse)
            curr = tokens.consume()
        
        # Defaults
        if term_weight is None:
            term_weight = self.env.formula_manager.Int(1)
        if term_group_id is None:
            term_group_id =  "I"  # default identifier for soft-clause
        params = [ 
            (":weight", weight_value),
            (":id",  term_group_id),
        ]            
```

pysmt/smtlib/parser/parser.py
```diff
@@ -1128,6 +1150,42 @@ def _cmd_assert(self, current, tokens):
...
+                raise NotImplementedError("The current option is not currently implemented %s "
+                                          "or some of them are repeated" % curr_parse)
+            curr = tokens.consume()
+        if param_presence[":weight"]:
+            params.append((":weight", weight_value))
+        if param_presence[":id"]:
+            params.append((":id", group_clause_id))
+        return SmtLibCommand(current, [expr, params])
+
+    def _cmd_check_allsat(self, current, tokens):
```
marcogario, line 1183: This is not part of http://optimathsat.disi.unitn.it/pages/smt2reference.html . If possible, let's update the OMT reference page, so we have a single place where to "point" to explain the extension.

pysmt/smtlib/parser/parser.py
```diff
@@ -449,6 +449,8 @@ def fix_real(op, *args):
 
         # Command tokens
         self.commands = {smtcmd.ASSERT : self._cmd_assert,
```
marcogario, line 451: Having them alphabetically is nice, but I would have a slight preference to move all the OMT specific commands at the end following a comment like:

```
# OMT Extension (http://optimathsat.disi.unitn.it/pages/smt2reference.html)
```

pysmt/smtlib/parser/parser.py
```diff
@@ -1181,6 +1239,34 @@ def _cmd_get_value(self, current, tokens):
         self.consume_closing(tokens, current)
         return SmtLibCommand(current, params)
 
+    def _cmd_get_objectives(self, current, tokens):
+        """(get-objective)"""
+        self.parse_atoms(tokens, current, 0)
+        return SmtLibCommand(current, [])
+
+    def _cmd_objective(self, current, tokens):
+        """(maximize | minimize <term> [:id <string>] [:signed] [:lower <const_term>] [:upper <const_term>])"""
```
marcogario, line 1248: The `:upper`, `:lower` bits only apply to `minmax` and `maxmin`. I think this command is not considering those, no?
http://optimathsat.disi.unitn.it/pages/smt2reference.html

pysmt/smtlib/parser/parser.py
```diff
@@ -1181,6 +1239,34 @@ def _cmd_get_value(self, current, tokens):
...
+            tokens.add_extra_token(curr)
+            curr_parse = self.parse_atom(tokens, current)
+            if curr_parse == ":lower" or curr_parse == ":upper":
+                exp = self.get_expression(tokens)
+                params.append((curr_parse, exp))
+            elif curr_parse == ":id":
+                id = self.parse_atom(tokens, "maximization/minimization")
+                params.append((curr_parse, id))
+            elif curr_parse == ":signed":
+                params.append((curr_parse,))
```
marcogario, line 1262: Minor: Is using the unary tuple the best way of representing this? I did not look at the rest of the code, but `(:signed, True)` and `(:signed, False)` might be nicer from a client perspective.

pysmt/smtlib/parser/parser.py
```diff
@@ -1181,6 +1239,34 @@ def _cmd_get_value(self, current, tokens):
...
+        params = []
+        curr = tokens.consume()
+        while curr != ")":
+            tokens.add_extra_token(curr)
+            curr_parse = self.parse_atom(tokens, current)
+            if curr_parse == ":lower" or curr_parse == ":upper":
+                exp = self.get_expression(tokens)
+                params.append((curr_parse, exp))
+            elif curr_parse == ":id":
+                id = self.parse_atom(tokens, "maximization/minimization")
```
marcogario, line 1259: Minor: Could you add a docstring to `parse_atom` explaining that the second argument `command` is used only for logging? I had a but of a :confused: :question: moment....

pysmt/smtlib/script.py
```diff
@@ -77,16 +77,53 @@ def serialize(self, outstream=None, printer=None, daggify=True):
...
             outstream.write("(%s (" % self.name)
             for a in self.args:
                 printer.printer(a)
                 outstream.write(" ")
             outstream.write("))")
 
+        elif self.name in [smtcmd.MAXIMIZE, smtcmd.MINIMIZE]:
+            outstream.write("(%s " % self.name)
+            printer.printer(self.args[0])
+            for arg in self.args[1]:
```
marcogario, line 103: Can we use a different variable name? I would expect arg to be a member of `self.args`, not of `self.args[i]`. I think above we used `a`. In this case this would also be consistent as we are talking about annotations.

pysmt/smtlib/script.py
```diff
@@ -77,16 +77,53 @@ def serialize(self, outstream=None, printer=None, daggify=True):
...
+                if len(arg) == 1:
+                    outstream.write(" %s" % arg[0])
+                elif len(arg) == 2:
+                    outstream.write(" %s " % arg[0])
+                    printer.printer(arg[1])
+                elif len(arg) > 2:
```
marcogario, line 109: I would not try to enforce the shape of this command here. We should rely on the fact that commands are well-formed, and not restrict what the user might want to do here.

I do not know how the rest of the code was using this, so there might be a reason for this.

pysmt/smtlib/script.py
```diff
@@ -77,16 +77,53 @@ def serialize(self, outstream=None, printer=None, daggify=True):
...
 
+        elif self.name in [smtcmd.MAXIMIZE, smtcmd.MINIMIZE]:
+            outstream.write("(%s " % self.name)
+            printer.printer(self.args[0])
+            for arg in self.args[1]:
+                if len(arg) == 1:
+                    outstream.write(" %s" % arg[0])
+                elif len(arg) == 2:
+                    outstream.write(" %s " % arg[0])
+                    printer.printer(arg[1])
```
marcogario, line 108: Minor: We might want to generalize this.

pysmt/test/smtlib/test_parser_omt.py
```diff
@@ -0,0 +1,146 @@
...
+            script = self.parse(file_id)
+            # Check cmds
+            for i, cmd in enumerate(script):
+                self.assertEqual(cmd.name, TESTS[file_id][i],
+                                 "Test %d: %s != %s " %
+                                 (file_id, cmd.name, TESTS[file_id][i]))
+            # Serialize
+            buf = StringIO()
+            script.serialize(buf)
+            self.assertTrue(True)
```
marcogario, line 41: We can probably be more strict here. At least buf should not be empty. Ideally, it would be the same size as the input file. Even better, if we parse it again, it should be the same as script.

pysmt/test/smtlib/test_parser_omt.py
```diff
@@ -0,0 +1,146 @@
...
+from pysmt.test import TestCase
+from pysmt.smtlib.parser import SmtLibParser
+
+
+class TestSmtLibParserOMT(TestCase):
+
+    def test_parse_omt(self):
+        for file_id in range(1, 4):
+            # Parse
+            script = self.parse(file_id)
```
marcogario, line 32: Minor: Could we make an iterator that yields the relevant bits here?

```python
for file_id, script, expected in self.examples():
    for i, cmd in enumerate(script):
       self.assertEqual(cmd.name, expected[i],  ...)
```

pysmt/test/smtlib/test_parser_omt.py
```diff
@@ -0,0 +1,146 @@
...
+                                 (file_id, cmd.name, TESTS[file_id][i]))
+            # Serialize
+            buf = StringIO()
+            script.serialize(buf)
+            self.assertTrue(True)
+
+    def test_omt_parsing_exception(self):
+        parser = SmtLibParser()
+        with self.assertRaises(NotImplementedError):
+            _ = parser.get_script(StringIO("(assert-soft false :weight (+ 3 (- 4 2)) :id goal :id goal)"))
```
marcogario, line 46:
```suggestion
            parser.get_script(StringIO("(assert-soft false :weight (+ 3 (- 4 2)) :id goal :id goal)"))
```

pysmt/smtlib/parser/parser.py
```diff
@@ -1181,6 +1239,34 @@ def _cmd_get_value(self, current, tokens):
...
+                params.append((curr_parse, id))
+            elif curr_parse == ":signed":
+                params.append((curr_parse,))
+            else:
+                raise NotImplementedError("The current option is not currently implemented %s" % curr_parse)
```
marcogario, line 1264: Using `NotImplementedError` suggests that this should be implemented. However, from the reference it does not seem to be the case. I would raise either a SyntaxError or a ValueError (or the corresponding in pySMT that we use elsewhere).

pysmt/smtlib/parser/parser.py
```diff
@@ -1128,6 +1150,42 @@ def _cmd_assert(self, current, tokens):
...
+            if curr_parse == ":weight" and param_presence[curr_parse]:
+                weight_value = self.get_expression(tokens)
+                params.append((curr_parse, weight_value))
+                param_presence[curr_parse] = False
+            elif curr_parse == ":id" and param_presence[curr_parse]:
+                group_clause_id = self.parse_atom(tokens, "assert-soft")
+                params.append((curr_parse, group_clause_id))
+                param_presence[curr_parse] = False
+            else:
+                raise NotImplementedError("The current option is not currently implemented %s "
```
marcogario, line 1174: See comment below about using `NotImplementedError`.

pysmt/test/smtlib/test_parser_omt.py
```diff
@@ -0,0 +1,146 @@
...
+            _ = parser.get_script(StringIO("(assert-soft false :not-implemented 12)"))
+        with self.assertRaises(NotImplementedError):
+            _ = parser.get_script(StringIO("(maximize z :upper 50 :lower 50 :not-implemented :id objective-1)"))
+
+    def test_command_option_value_correctness(self):
+        parser = SmtLibParser()
+        script = parser.get_script(StringIO("(assert-soft false :weight 2 :id goal)"))
+        cmd = next(iter(script))
+        assert cmd.name == "assert-soft"
+        assert len(cmd.args) == 2
```
marcogario, line 61: This should use `self.assertEquals`. Even better would be to combine this data into the test array below (`TESTS`).

pysmt/smtlib/parser/parser.py
```diff
@@ -591,7 +597,25 @@ def _smtlib_underscore(self, stack, tokens, key):
...
+            fnv = self.get_expression(tokens)
+            if fnv.is_int_constant():
+                v = fnv.constant_value()
+            else:
+                raise PysmtSyntaxError("Expected number in '_ to_bv' expression: "
+                    "'%s'" % op, tokens.pos_info)
+            if v >= 0:
+                fun = mgr.BV(v, width)
+            else:
+                fun = mgr.SBV(v, width)
```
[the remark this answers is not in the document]
cespio: It is not bug, but the extended SMTLIB-v2  version supported by optimathasat it allows to express the the FixedSizeBitvectors also with the following syntax `((_ to_bv bvsize) value)`

http://optimathsat.disi.unitn.it/pages/smt2examples.html

pysmt/smtlib/parser/parser.py
cespio: That command is inherited from mathsat 
https://mathsat.fbk.eu/smt2examples.html

pysmt/smtlib/parser/parser.py
cespio: `:upper` and `:lower` are not used anymore also in `minmax` and `maxmin`. Currently I haven't implement the parsing for `minmax` and `maximin` yet.

pysmt/smtlib/parser/parser.py
marcogario: I would be careful with the extensions to the parser that are not part of the SMTLIB format. We need to keep track of where they are coming from, so that we know how to deal with potential conflicts in the future.

I can prepare a patch for the documentation to include both this change and the `to_bv` change above (in addition to the OMT stuff).

cespio:
Hi guys, I just want to gently ask if there are any update on this PR! Thanks :slightly_smiling_face:

marcogario:
:bowing_man: Sorry, this fell off my radar. From a quick look, it seems we are only missing some documentation. I would be ok to merge this now, as I won't be able to get around to update the docs in the next few weeks.

@mikand this targets he optimization branch that you were working on. If this is ok for you, can you merge?

PR #647, opened 2020-06-09
Fixing macOS X compatibility

vrjuliao:
To use "sed -I" command in macOS is necessary to add backslash to represent the regular expression parameter. It's also necessary put the regex between single quotes

### DIFF
--- a/pysmt/cmd/installers/cvc4.py
+++ b/pysmt/cmd/installers/cvc4.py
@@ -61,9 +61,10 @@ class CVC4Installer(SolverInstaller):
         PYTHON_LIBRARY = os.environ.get('PYSMT_PYTHON_LIBDIR')
         if not PYTHON_LIBRARY:
             PYTHON_LIBRARY = sysconfig.get_config_var('LIBDIR')
-        PYTHON_INCLUDE_DIR = sysconfig.get_python_inc()
-        SolverInstaller.run(['sed', '-i',
-                             's|cmake_opts=""|cmake_opts="-DPYTHON_LIBRARY=' + PYTHON_LIBRARY + ' -DPYTHON_INCLUDE_DIR=' + PYTHON_INCLUDE_DIR + '"|g',
+        PYTHON_LIBRARY = self.__add_backslash(PYTHON_LIBRARY)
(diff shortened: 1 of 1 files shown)


pysmt/cmd/installers/cvc4.py
```diff
@@ -84,3 +85,14 @@ def compile(self):
 
     def get_installed_version(self):
         return self.get_installed_version_script(self.bindings_dir, "cvc4")
+
+    def __add_backslash(self, str_path):
```
mikand, line 89: Isn't this simply
```python
str_path.replace('/', '\\')
```
?

pysmt/cmd/installers/cvc4.py
```diff
@@ -61,9 +61,10 @@ def compile(self):
         PYTHON_LIBRARY = os.environ.get('PYSMT_PYTHON_LIBDIR')
         if not PYTHON_LIBRARY:
             PYTHON_LIBRARY = sysconfig.get_config_var('LIBDIR')
-        PYTHON_INCLUDE_DIR = sysconfig.get_python_inc()
-        SolverInstaller.run(['sed', '-i',
-                             's|cmake_opts=""|cmake_opts="-DPYTHON_LIBRARY=' + PYTHON_LIBRARY + ' -DPYTHON_INCLUDE_DIR=' + PYTHON_INCLUDE_DIR + '"|g',
+        PYTHON_LIBRARY = self.__add_backslash(PYTHON_LIBRARY)
+        PYTHON_INCLUDE_DIR = self.__add_backslash(sysconfig.get_python_inc())
+        SolverInstaller.run(["perl", "-i''", "-pe",
```
mikand, line 66: why using `perl` instead of `sed`? The latter is much more common (at least in the linux world).
If on OSX this is not available, you can condition the use of `perl` by checking the platform:
```python
if self.os_name == 'darwin':
    ...
```

mikand (changes requested):
@vrjuliao Thanks for the contribution.
I added few comments in the code.

Moreover, you should change the `azure-pipelines.yml` file to enable the automatic testing of CVC4 on OSX, so that we can check that everything works. Let me know if you need help for this.

PR #655, opened 2020-08-23
omtlib-solver

arrighi79:


pysmt/smtlib/parser/parser.py
```diff
@@ -1243,8 +1242,15 @@ def _cmd_get_objectives(self, current, tokens):
         self.parse_atoms(tokens, current, 0)
         return SmtLibCommand(current, [])
 
+    def _cmd_minmax_maxmin_obj(self, current, tokens):
+        """(minmax | maxmin (<term>+) [:id <string>] [:signed])"""
+        params = self.parse_expr_list(tokens, current)
+        self.consume_closing(tokens, current)
```
mikand, line 1248: Add a todo or something because here you are not supporting the `:signed` nor the `:id` options...

pysmt/smtlib/script.py
```diff
@@ -392,3 +419,36 @@ def evaluate_command(cmd, solver):
...
+        if client_data is not None:
+            client_data[0].append(rt)
+        return rt
+    elif cmd.name == smtcmd.MINIMIZE:
+        rt = MinimizationGoal(cmd.args[0])
+        if client_data is not None:
+            client_data[0].append(rt)
+        return rt
+    elif cmd.name == smtcmd.CHECK_SAT:
+        if client_data != None:
```
mikand, line 436: Never use `!=` with `None`
```suggestion
        if client_data is not None:
```

pysmt/smtlib/script.py
```diff
@@ -176,6 +177,20 @@ def serialize_to_string(self, daggify=True):
         self.serialize(buf, daggify=daggify)
         return buf.getvalue()
 
+class GoalsManager(object):
```
mikand, line 180: Why here? I think this bit of logic should go into `shell.py` or better into a class representing a gnereic SmtLib-based OMT Solver...

Moreover, to me this class is useless, it does nothing more than a list...

pysmt/smtlib/script.py
```diff
@@ -176,6 +177,20 @@ def serialize_to_string(self, daggify=True):
...
+    def __init__(self):
+        self.goals = []
+
+    def add(self, g):
+        self.goals.append(g)
+
+    def p(self):
```
mikand, line 187: Cant' understand the meaning of this method:

- `rt` is useless
- it always returns None
- it is not documented

pysmt/smtlib/script.py
```diff
@@ -194,10 +209,23 @@ def add_command(self, command):
     def evaluate(self, solver):
         log = []
         for cmd in self.commands:
-            r = evaluate_command(cmd, solver)
+            r = omt_evaluate_command(cmd, solver)
             log.append((cmd.name, r))
+
         return log
 
+    def evaluate(self, solver, client_data):
```
mikand, line 217: In python there is no function overload! You cannot have two functions with the same name: only the second one is considered (it "overrides" the first one). I am sure your IDE gives you a warning about this, **please read the warnings**!

Moreover, do not change the old behavior, add new behavior as needed: in this case you can keep the old `evaluate` function and simply add a new one `omt_evaluate`...

pysmt/smtlib/script.py
```diff
@@ -392,3 +419,36 @@ def evaluate_command(cmd, solver):
                                   "Please open a bug-report." % cmd.name)
     else:
         raise UnknownSmtLibCommandError(cmd.name)
+
+def omt_evaluate_command(cmd, optimizer, client_data = None):
```
mikand, line 423: We discussed this: if you simply rename `client_data` into `optimization_goals` and make it non-optional you can simply avoid the ugly 
```
 if client_data is not None:
￼            client_data[0].append(rt)
```
And simply do `optimization_goals.append(rt)` when needed... Isn't this simpler?

pysmt/cmd/shell.py
```diff
@@ -107,18 +115,32 @@ def print_result(self, cmd, result):
...
     def smtlib_solver(self, stream):
         smt_parser = SmtLibParser()
-        name = self.args.solver
-        if name == "auto":
-            solver = Solver()
+        s_name = self.args.solver
+        opt_name = self.args.optimizer
+        client_data = None
+        if opt_name is not None:
+            client_data = ([], [])
```
mikand, line 132: Why on earth do you need a pair of lists?

pysmt/smtlib/script.py
```diff
@@ -392,3 +419,36 @@ def evaluate_command(cmd, solver):
...
+    elif cmd.name == smtcmd.MINIMIZE:
+        rt = MinimizationGoal(cmd.args[0])
+        if client_data is not None:
+            client_data[0].append(rt)
+        return rt
+    elif cmd.name == smtcmd.CHECK_SAT:
+        if client_data != None:
+            client_data[1].clear()
+            for g in client_data[0]:
+                client_data[1].append((g.term(), optimizer.optimize(g)[1]))
```
mikand, line 439: Why? Can't you simply return the optimal model or a dictionary with the objectives values?

mikand (changes requested):
Mostly this is OK, but I think this code is over-complicated and can be greatly simplified.

pysmt/cmd/shell.py
arrighi79: because in the first list i save the goals and in the second the results of it. Since the calculation is run in different time from the print i prefer keep both list, with a corrispondetion 1:1 goal => result

pysmt/smtlib/script.py
arrighi79: it's necessary to return the model?

pysmt/optimization/optimizer.py
```diff
@@ -320,7 +320,7 @@ def optimize(self, goal, strategy='linear',
         rt = None, None
         if goal.is_maximization_goal() or goal.is_minimization_goal():
             rt = self._optimize(goal, strategy)
-        else:
+        else:#elif maxsmt => gestione maxsmt()
```
mikand, line 323: Remove comment

pysmt/optimization/optimsat.py
```diff
@@ -106,15 +106,15 @@ def _assert_msat_goal(self, goal):
             cost_function = goal.term()
             obj_fun = self.converter.convert(cost_function)
 
-        else:
+        else: #elif maxsmt
```
mikand, line 109: remove commented code

pysmt/optimization/goal.py
```diff
@@ -108,12 +111,13 @@ class MinimizationGoal(Goal):
     Attention: some Optimizer may not support this goal
     """
 
-    def __init__(self, formula):
+    def __init__(self, formula, sign = True):
```
mikand, line 114:
```suggestion
    def __init__(self, formula, signed=False):
```
sign = segno
signed = con-segno

signed suona meglio.

Moreover, I think that by default we shall be unsigned  (just like z3 and optimathsat)

pysmt/optimization/goal.py
```diff
@@ -86,12 +88,13 @@ class MaximizationGoal(Goal):
...
```
mikand, line 91:
```suggestion
    def __init__(self, formula, signed=False):
```

pysmt/optimization/goal.py
```diff
@@ -75,7 +74,10 @@ def get_logic(self):
             return logic
 
     def signed(self):
-        return True
+        return self.bv_signed
```
mikand, line 77: Not a fan of getters and setters, but OK.

A better alternative would be to use the `@property` decorator: https://docs.python.org/3/library/functions.html#property

pysmt/optimization/goal.py
```diff
@@ -125,15 +129,31 @@ def term(self):
 
 class MinMaxGoal(MinimizationGoal):
 
-    def __init__(self, terms):
-        MinimizationGoal.__init__(self, get_env().formula_manager.Max(terms))
+    def __init__(self, terms, sign = True):
+        if len(terms) > 0:
+            if get_env().stc.get_type(terms[0]).is_bv_type():
```
mikand, line 134: It's a bit dirty to use `get_env()` in pysmt classes. The correct solution would be to have an `env` field in the goal classes that is initialized to `get_env()` if no explicit environment is passed. 

This is a bit technical into the pysmt structure. If you have problems, just leave a `#TODO` comment about this here and I'll fix it in the `optimization` branch

pysmt/optimization/z3.py
```diff
@@ -42,7 +42,12 @@ def __init__(self, environment, logic, **options):
         self.z3 = z3.Optimize()
 
     def _assert_z3_goal(self, goal):
-        obj = self.converter.convert(goal.term())
+        term = goal.term()
+        ty = self.environment.stc.get_type(term)
+        if goal.signed() and ty.is_bv_type():
```
mikand, line 47: Maybe add a comment with a link to the Z3 issue discussing this: https://github.com/Z3Prover/z3/issues/1339

pysmt/smtlib/parser/parser.py
```diff
@@ -1243,8 +1242,16 @@ def _cmd_get_objectives(self, current, tokens):
         self.parse_atoms(tokens, current, 0)
         return SmtLibCommand(current, [])
 
+    def _cmd_minmax_maxmin_obj(self, current, tokens):
+        """(minmax | maxmin (<term>+) )"""
+        """TODO: [:id <string>] [:signed]"""
```
mikand, line 1247: Please check and resolve all TODOs...

pysmt/smtlib/script.py
```diff
@@ -329,66 +333,129 @@ def smtlibscript_from_formula(formula, logic=None):
...
+                                      "but it is currently not supported. "\
+                                      "Please open a bug-report." % cmd.name)
+        else:
+            raise UnknownSmtLibCommandError(cmd.name)
+
+
+class InterpreterOMT(InterpreterSMT):
+
+    def __init__(self):
+        self.optimization_goals = ([],[])
```
mikand, line 408: Maybe add some comments about the class and how this field is used for future reference

pysmt/test/test_optimization.py
```diff
@@ -323,5 +323,104 @@ def test_infinitesimal(self):
...
+        for oname in get_env().factory.all_optimizers(logic=QF_LIA):
+            with Optimizer(name=oname) as opt:
+                opt.add_assertion(f1)
+                opt.add_assertion(f2)
+                model, values = opt.lexicographic_optimize([obj1, obj2, obj3, obj4, obj5])
+                self.assertEqual(values[0], Int(5))
+                self.assertEqual(values[1], Int(0))
+                self.assertEqual(values[2], Int(10))
+
+    def test_minimization_basic_BV(self):
```
mikand, line 349: Not a fan of capital letters in function names...

pysmt/formula.py
```diff
@@ -528,31 +528,53 @@ def Xor(self, left, right):
...
+
+    def MinBV(self, sign, *args):
+        """Returns the encoding of the minimum expression within args"""
+        le = self.BVULE
+        if sign:
+            le = self.BVSLE
+        return self.MinWrap( le, *args)
+
+    def MaxBV(self, sign, *args):
+        """Returns the encoding of the maximum expression within args"""
```
mikand, line 565: Fix this documentation comment

pysmt/formula.py
```diff
@@ -528,31 +528,53 @@ def Xor(self, left, right):
...
             a, b = exprs
-            return self.Ite(self.LE(a, b), b, a)
+            return self.Ite(le(a, b), b, a)
         else:
             h = len(exprs) // 2
-            return self.Max(self.Max(exprs[0:h]), self.Max(exprs[h:]))
+            return self.MaxWrap(le, self.MaxWrap(le,exprs[0:h]), self.MaxWrap(le,exprs[h:]))
+
+    def MinBV(self, sign, *args):
+        """Returns the encoding of the minimum expression within args"""
```
mikand, line 558: Fix this documentation comment

pysmt/formula.py
```diff
@@ -528,31 +528,53 @@ def Xor(self, left, right):
...
-            return self.Ite(self.LE(a, b), a, b)
+            return self.Ite(le(a, b), a, b)
         else:
             h = len(exprs) // 2
-            return self.Min(self.Min(exprs[0:h]), self.Min(exprs[h:]))
+            return self.MinWrap(le, self.MinWrap(le, exprs[0:h]), self.MinWrap(le, exprs[h:]))
 
-    def Max(self, *args):
+    def MaxWrap(self, le, *args):
         """Returns the encoding of the maximum expression within args"""
```
mikand, line 545: Fix this documentation comment

pysmt/formula.py
```diff
@@ -528,31 +528,53 @@ def Xor(self, left, right):
         """Returns the xor of left and right: left XOR right """
         return self.Not(self.Iff(left, right))
 
-    def Min(self, *args):
+    def MinWrap(self, le, *args):
         """Returns the encoding of the minimum expression within args"""
```
mikand, line 532: Fix this documentation comment

pysmt/optimization/goal.py
```diff
@@ -86,12 +88,13 @@ class MaximizationGoal(Goal):
     Attention: some Optimizer may not support this goal
```
mikand, line 88: "Attention" used in this way is not English :smile: 

```suggestion
    Warning: some Optimizer classes may not support this goal
```

pysmt/optimization/goal.py
```diff
@@ -125,15 +129,31 @@ def term(self):
 
 class MinMaxGoal(MinimizationGoal):
```
mikand, line 130: These classes lack documentation....

pysmt/optimization/optimsat.py
```diff
@@ -106,15 +106,15 @@ def _assert_msat_goal(self, goal):
...
             raise GoalNotSupportedError("optimathsat", goal)
 
-        msat_obj = make_fun(self.msat_env(), obj_fun, False)
+        msat_obj = make_fun(self.msat_env(), obj_fun, signed = goal.signed())
         self._msat_lib.msat_assert_objective(self.msat_env(), msat_obj)
         return msat_obj
 
     def optimize(self, goal, **kwargs):
-
+        #if goal esiste fai altrimenti ...
```
mikand, line 117: Please check for comments in Italian and remove/translate them.

Also, this comment is quite useless IMHO...

mikand (changes requested):
Technically this looks good, there is just some code-cleaning to do.

mikand:
@arrighi79 There is still an error on the tests: https://dev.azure.com/micheliandrea/PySMT/_build/results?buildId=533&view=logs&jobId=bc7e454d-2c8a-5fcd-9192-2a7565b6d905&j=bc7e454d-2c8a-5fcd-9192-2a7565b6d905&t=816b5524-d063-584a-d13a-0408c0e7884c

mikand (approved):

PR #658, opened 2020-10-13
Handling of algebraic constants in simplify

EnricoMagnago:
I found an issue in which the simplify procedure crashed because it could not handle an algebraic constant.
In my case the expression to be simplified was an addition.
The code `res = sum(x.constant_value() for x in args)` was returning a res of type z3.Numeral, which the later call to `self.manager.Real(res)` could not handle.

I copy pasted the structure used in walk_times for walk_plus, and tried to handle the algebraic constant also in the walk for the pow and minus operators.

I added the `walk_algebraic_constant` method to the identity walker.

This happened while using z3 as SMT solver.
The converter in z3.py was not able to handle node of type algebraic constant, I tried to implement it. I am not sure whether this is the correct or best way to implement it. In my use cases it seems to work. 

In addition, I have also added the possibility to specify some options for the UnsatCoreSolver.

### DIFF
--- a/pysmt/oracles.py
+++ b/pysmt/oracles.py
@@ -191,12 +191,13 @@ class TheoryOracle(walkers.DagWalker):
 
     @walkers.handles(op.REAL_CONSTANT, op.BOOL_CONSTANT)
     @walkers.handles(op.INT_CONSTANT, op.BV_CONSTANT)
+    @walkers.handles(op.ALGEBRAIC_CONSTANT)
     @walkers.handles(op.STR_CONSTANT)
     def walk_constant(self, formula, args, **kwargs):
         """Returns a new theory object with the type of the constant."""
         #pylint: disable=unused-argument
         theory_out = Theory()
-        if formula.is_real_constant():
+        if formula.is_real_constant() or formula.is_algebraic_constant():
             theory_out.real_arithmetic = True
             theory_out.real_difference = True
         elif formula.is_int_constant():
--- a/pysmt/simplifier.py
+++ b/pysmt/simplifier.py
@@ -280,7 +280,8 @@ class Simplifier(pysmt.walkers.DagWalker):
         return self.manager.Exists(varset, sf)
 
     def walk_plus(self, formula, args, **kwargs):
-        new_args = []
+        to_sum = []
+        to_sub = []
         constant_add = 0
         stack = list(args)
         ttype = self.env.stc.get_type(args[0])
@@ -293,8 +294,31 @@ class Simplifier(pysmt.walkers.DagWalker):
                 constant_add += x.constant_value()
             elif x.is_plus():
                 stack += x.args()
+            elif x.is_minus():
+                to_sum.append(x.arg(0))
+                to_sub.append(x.arg(1))
+            elif x.is_times() and x.args()[-1].is_constant():
+                const = x.args()[-1]
+                const_val = const.constant_value()
+                if const_val < 0:
+                    new_times = list(x.args()[:-1])
+                    if const_val != -1:
+                        const_val = -const_val
+                        if const.is_algebraic_constant():
+                            from pysmt.constants import Numeral
+                            const = self.manager._Algebraic(Numeral(const_val))
+                        elif ttype.is_real_type():
+                            const = self.manager.Real(const_val)
+                        else:
+                            assert ttype.is_int_type()
+                            const = self.manager.Int(const_val)
+                        new_times.append(const)
+                    new_times = self.manager.Times(new_times)
+                    to_sub.append(new_times)
+                else:
+                    to_sum.append(x)
             else:
-                new_args.append(x)
+                to_sum.append(x)
 
         const = None
         if is_algebraic:
@@ -306,11 +330,17 @@ class Simplifier(pysmt.walkers.DagWalker):
             assert ttype.is_int_type()
             const = self.manager.Int(constant_add)
 
-        if len(new_args) == 0:
+        if len(to_sum) == 0 and len(to_sub) == 0:
             return const
-        elif not const.is_zero():
-            new_args.append(const)
-        return self.manager.Plus(new_args)
+        if not const.is_zero():
+            to_sum.append(const)
+
+        res = self.manager.Plus(to_sum)
+
+        if to_sub:
+            sub = self.manager.Plus(to_sub)
+            res = self.manager.Minus(res, sub)
+        return res
 
     def walk_times(self, formula, args, **kwargs):
         new_args = []
--- a/pysmt/simplifier.py
+++ b/pysmt/simplifier.py
@@ -354,8 +354,12 @@ class Simplifier(pysmt.walkers.DagWalker):
         new_args = sorted(new_args, key=FNode.node_id)
         return self.manager.Times(new_args)
 
-
     def walk_pow(self, formula, args, **kwargs):
+        if args[0].is_algebraic_constant():
+            from pysmt.constants import Numeral
+            l = args[0].constant_value()
+            r = args[1].constant_value()
+            return self.manager._Algebraic(Numeral(l**r))
         if args[0].is_real_constant():
             l = args[0].constant_value()
             r = args[1].constant_value()
@@ -371,6 +375,13 @@ class Simplifier(pysmt.walkers.DagWalker):
 
         sl = args[0]
         sr = args[1]
+
+        if sl.is_algebraic_constant() and \
+           sr.is_algebraic_constant():
+            from pysmt.constants import Numeral
+            l = sl.constant_value()
+            r = sr.constant_value()
+            return self.mananger._Algebraic(Numeral(l - r))
 
         if sl.is_real_constant() and sr.is_real_constant():
             l = sl.constant_value()
--- a/pysmt/simplifier.py
+++ b/pysmt/simplifier.py
@@ -15,6 +15,8 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 #
+from collections import defaultdict
+
 from six.moves import xrange
 
 import pysmt.walkers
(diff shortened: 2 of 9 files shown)


pysmt/simplifier.py
```diff
@@ -280,47 +280,37 @@ def walk_exists(self, formula, args, **kwargs):
         return self.manager.Exists(varset, sf)
 
     def walk_plus(self, formula, args, **kwargs):
```
marcogario, line 282: This method was handling a few additional simplifications. 
Could you find or add a test that shows that old behaviors have been preserved?
Could you also add a test for showing that the handling of the algebraic numbers is correct?

pysmt/simplifier.py
```diff
@@ -280,47 +280,37 @@ def walk_exists(self, formula, args, **kwargs):
...
-        is_real = any(x.is_real_constant() for x in args)
-        is_int = any(x.is_int_constant() for x in args)
```
marcogario, line 284: To check: Do we allow mixing Integer and Real constants in a Plus? I thought we did to avoid the need for explicitly casting the constants, but it is worth checking. If we don't, it seems checking the first argument (as you do below) would be sufficient.

pysmt/simplifier.py
```diff
@@ -280,47 +280,37 @@ def walk_exists(self, formula, args, **kwargs):
...
+        constant_add = 0
+        stack = list(args)
+        ttype = self.env.stc.get_type(args[0])
+        is_algebraic = False
+        while len(stack) > 0:
+            x = stack.pop()
+            if x.is_constant():
+                if x.is_algebraic_constant():
+                    is_algebraic = True
+                constant_add += x.constant_value()
```
marcogario, line 293: My key question is whether Z3 is doing the right thing here, and overloading the + . A test would be the easiest way to show this.

pysmt/simplifier.py
```diff
@@ -280,47 +280,37 @@ def walk_exists(self, formula, args, **kwargs):
...
-        ns = [x for x in args if not x.is_constant()]
-        if len(ns) < len(args):
-            num = sum(x.constant_value() for x in args if x.is_constant())
-            if num != 0:
-                if is_real:
-                    ns.append(self.manager.Real(num))
-                else:
-                    ns.append(self.manager.Int(num))
-
-        if len(ns) == 2:
```
marcogario, line 303: To check: I think these cases would not be handled, because the Times subexpression would just endup in `newargs` but never simplified. However, since we simplify bottom up, it might be that our `walk_times` got smarter since we wrote this and this is actually dead code.

pysmt/simplifier.py
```diff
@@ -364,8 +354,12 @@ def walk_times(self, formula, args, **kwargs):
         new_args = sorted(new_args, key=FNode.node_id)
         return self.manager.Times(new_args)
 
-
     def walk_pow(self, formula, args, **kwargs):
+        if args[0].is_algebraic_constant():
+            from pysmt.constants import Numeral
+            l = args[0].constant_value()
+            r = args[1].constant_value()
+            return self.manager._Algebraic(Numeral(l**r))
```
marcogario, line 362: Again, a simple test showing that this has the semantics you expect would be good.
Could you move this as the last case of this if? I expect this to be false more often than not.

pysmt/simplifier.py
```diff
@@ -382,6 +376,13 @@ def walk_minus(self, formula, args, **kwargs):
         sl = args[0]
         sr = args[1]
 
+        if sl.is_algebraic_constant() and \
```
marcogario, line 379: Same here, can you move this after the other (more likely) cases?

marcogario (commented):
Thanks for the patch. The changes here look good.

I see that you also extended the UnsatSolver functions to take options. That is good! Please add a comment about this in the PR description.

As you saw the Algebraic constants are not extensively supported, but hopefully this will be sufficient for your work.

Before we can merge this, we need some tests for these simplifications You can take a look at `test_simplifier.py` for the general approach. Typically, write some expression, simplify it, and ask the solver whether it is equivalent to the original.

pysmt/simplifier.py
EnricoMagnago: I don't think mixed Int - Real constants are supported in math operators +, -, * and /.

If I understand correctly the type checking of math operators is implemented in `type_checker.py` by the method `walk_realint_to_realint`.
The latter calls `walk_type_to_type` with either REAL or INT as arguments.

Also the following simple script throws a PysmtTypeError:
```
from pysmt.shortcuts import *
one = Real(1)
two = Int(2)
three = Plus(one, two)
```

pysmt/simplifier.py
EnricoMagnago: Yes, you are right. I will re-add such rewriting.

In addition, using a simple dictionary it could be possible to perform some other simplifications like adding the constant coefficients of the terms of the addition having the same symbolic factors (`2*x*y + 3*x*y` into `5xy`).

marcogario: The general approach looks reasonable. However, the simplifier is called in several places, so we need to make sure that this is not significantly slower than it was before. We have some scripts in the root directory that we used to benchmark the parser. The same could be extended to benchmark the simplifier and show that these changes are not significantly slower.
https://github.com/pysmt/pysmt/blob/master/SMT-LIB/parse_all.py

It would be best to move those changes to a different PR in the interest of merging this PR sooner.

marcogario:
Please let me know if you need help with writing the tests, and just mention me or re-request my review once this is ready. Thanks!

pysmt/simplifier.py
EnricoMagnago: I have added a few very simple test cases, for both the simplification of Plus and the handling of algebraic numbers.   
The test cases for algebraic numbers highlighted a few other places in which they where not handled: TheoryOracle and MsatConverter.    
I tried to implement those, please have a look at them.    
* In TheoryOracle I simply consider them as real constants, also in the type checker the are always associated with REAL type. However, in principle a z3num.Numeral could also contain an integer value, I guess algebraic constants should never appear in pysmt-expressions of type INT, but I am not sure if, where and how this is enforced.
* In MsatConverter I handled z3num.Numeral that represent rational values. In the case of an irrational one a convert exception is raised. It could be possible to handle also this case by traversing the AST of the z3 numeral and building a corresponding msat expression. However, I haven't looked at how such ASTs are built and this might be part of a possibly larger "issue" about the representation of irrational numbers in pysmt.

pysmt/simplifier.py
EnricoMagnago: I have added a simple test case for both addition and multiplication in test/test_simplify.py" `test_plus_algebraic` and `test_times_algebraic`.    
I have annotated them with `@skipIfSolverNotAvailable("z3")`, since if I am not mistaken algebraic constants are instantiated only when z3 is available.

pysmt/simplifier.py
EnricoMagnago: Ok, I reverted the changes that implemented the simplification of linear combination of symbols. They have been stored on an other branch.

The current implementation implements a slight generalization of the previous simplification. All terms with negative constant are grouped in the rhs of a Minus expression.   
This implementation never goes down the AST more than a couple of levels.

There is a bit of code duplication to handle the constant factor of Times expressions.

I haven't looked at the benchmarking part yet.

pysmt/simplifier.py
EnricoMagnago: I have added a simple test `test_times_algebraic` that is supposed to test this.

EnricoMagnago:
As a side note, while I was experimenting a bit, I quite often ended up with a segfault.   
For example, the following script deterministically ends with a segmentation fault.

```
from pysmt.shortcuts import *
from pysmt.constants import Numeral

env = get_env()
mgr = env.formula_manager
p_2 = mgr._Algebraic(Numeral(2))
print(p_2)
```

I have not investigated this any further.

pysmt/oracles.py
```diff
@@ -191,12 +191,13 @@ def walk_combine(self, formula, args, **kwargs):
...
     @walkers.handles(op.REAL_CONSTANT, op.BOOL_CONSTANT)
     @walkers.handles(op.INT_CONSTANT, op.BV_CONSTANT)
+    @walkers.handles(op.ALGEBRAIC_CONSTANT)
     @walkers.handles(op.STR_CONSTANT)
     def walk_constant(self, formula, args, **kwargs):
         """Returns a new theory object with the type of the constant."""
         #pylint: disable=unused-argument
         theory_out = Theory()
-        if formula.is_real_constant():
+        if formula.is_real_constant() or formula.is_algebraic_constant():
```
marcogario, line 200: I wonder if we want to distinguish this case. For example, for Arrays, we have the default constant value of arrays that is supported by MathSAT and Z3 (but not SMT-LIB), so we distinguish it explicitly in the logic https://github.com/pysmt/pysmt/blob/d1fadc09e745c7286abad93ba51e72cf6e285f58/pysmt/logics.py#L43 .

This would then allow you to cleanly state that only z3 supports this extension, and prevent pysmt from trying to use any of the other solvers if an algebraic constant is present.

pysmt/simplifier.py
marcogario: Yes, that is correct. Algebraic constants are currently defined by using the internal z3 representation. We have no other way to represent them, so z3 is a requirement.

marcogario (commented):
The changes look overall good, let me have a closer look at the simplification as it is more complex than I expected.

In general, Algebraic types are not supported in pysmt. Meaning that, as you point out, there are several places that ignore them. You can also see that within the `FormulaManager` the method is prefixed with a `_` that is meant to indicate that this is not a public method.

AFAIR they are truly only supported by z3. I would be curious to learn if this has changed, and more solvers support them, otherwise it is hard to justify spending time supporting them more widely. Nevertheless, it seems they are useful for you, so I value your opinion on the topic.

>  In TheoryOracle I simply consider them as real constants, also in the type checker the are always associated with REAL type. However, in principle a z3num.Numeral could also contain an integer value, I guess algebraic constants should never appear in pysmt-expressions of type INT, but I am not sure if, where and how this is enforced.

So far, we should only receive them from Z3 while reading back an expression (back conversion). We would need to check whether Z3 returns the simpler integer type in this case.
   
>  In MsatConverter I handled z3num.Numeral that represent rational values. In the case of an irrational one a convert exception is raised. It could be possible to handle also this case by traversing the AST of the z3 numeral and building a corresponding msat expression. However, I haven't looked at how such ASTs are built and this might be part of a possibly larger "issue" about the representation of irrational numbers in pysmt.

Indeed, I would just not support them outside z3.

pysmt/oracles.py
EnricoMagnago: I would suggest to distinguish explicitly in the logic only the case in which z3 numerals are used to represent irrational values.

One of the things I like about pysmt is the ability to use different smt solvers in the same application with very little effort. You can use one SMT solver for a certain part of the computation, take its output and feed it to another SMT solver to continue the computation. 
In my case I would like, for instance, to take whatever output or model I get from z3 and use it to compute some SMT-query I submit to mathsat. If z3 in the model returns some algebraic constant I would like pysmt to take care of that when converting the expression for mathsat as much as possible.   

That said, I agree that carrying around a structure specific to z3 is not the cleanest way to achieve this.
A possibly cleaner way could be to immediately convert any z3 Numeral representing an integer or rational value in the output of z3 to a pysmt fraction (can we achieve this in te backward converter?). While irrational values could be carried around as algebraic values, associating such expressions with the "specific" logic.     
However, irrational constants like `sqrt(2)` that might be represented as z3 Numeral could also be handled by any other solver supporting `sqrt` as operator. The solution above will not handle these cases, but I think it could be acceptable anyway.

marcogario (commented):
I'll be looking at the performance impact of this change today, and hopefully we will get it in soon. Thanks for the updates.

marcogario (approved):
The benchmarking looks ok. We are paying some penalty, but this does not seems < 5%. 
The main observation (for future reference) is that a statement like:

```python
        if all(x.is_constant() for x in args):
            res = sum(x.constant_value() for x in args)
```

can be evaluated mostly in CPython, while unfolding it in a stack like:

```python
        while len(stack) > 0:
            x = stack.pop()
```
requires jumping from Python to C.

:warning: My laptop is not currently setup for getting very reproducible numbers, but I ran these a few times and the results seem consistent. The number next to the logic is the number of instances/files. The threading was 5. See script in https://github.com/pysmt/pysmt/pull/669

```
QF_LIA (400):

> Branch
The total execution time was 1782.95 seconds
The mean execution time was 4.46 seconds
The max execution time was 381.92 seconds
The statistics file has been generated in 'stats.out'

> Master

The total execution time was 1759.35 seconds
The mean execution time was 4.40 seconds
The max execution time was 369.35 seconds
The statistics file has been generated in 'stats.out'

QF_NRA (200):

> Branch
The total execution time was 54.68 seconds
The mean execution time was 0.27 seconds
The max execution time was 8.17 seconds

> Master
The total execution time was 53.43 seconds
The mean execution time was 0.27 seconds
The max execution time was 7.34 seconds
```

marcogario:
I reverted the last 2 commits, as you need at least a solver to run the test (although you do not really care if it is z3 or not). Sorry for the confusion.

PR #659, opened 2020-10-16
BUG: define UFNIA as logic with integer arithmetic

johnyf:
The keyword argument `integer_arithmetic=True` was missing
from the constructor of `pysmt.logics.Logic` for the logic
UFNIA [1, 2] ("Non-linear integer arithmetic with uninterpreted
sort and function symbols." [1]). See for example how the
logic QF_UFNIA is defined in the same module.

A consequence of the previous definition of UFNIA was
type checking errors due to numerals without `'.'` being
parsed as reals instead of integers in the module
`pysmt.smtlib.parser.parser` [3].

[1] http://smtlib.cs.uiowa.edu/logics.shtml
[2] http://smtlib.cs.uiowa.edu/logics-all.shtml#UFNIA
[3] https://github.com/pysmt/pysmt/blob/35ac3fd88b01706ba01d2a7cf291194fda0c8f92/pysmt/smtlib/parser/parser.py#L665-L672

### DIFF
--- a/pysmt/logics.py
+++ b/pysmt/logics.py
@@ -590,6 +590,7 @@ UFNIA = Logic(name="UFNIA",
               description=\
 """Non-linear integer arithmetic with uninterpreted sort and function
 symbols.""",
+              integer_arithmetic=True,
               integer_difference=True,
               linear=False,
               uninterpreted=True)


pysmt/logics.py
```diff
@@ -590,6 +590,7 @@ def __hash__(self):
               description=\
 """Non-linear integer arithmetic with uninterpreted sort and function
 symbols.""",
+              integer_arithmetic=True,
               integer_difference=True,
```
marcogario, line 594: To check: I think we expect only one of these to be true. `integer_arithmetic` is a strict superset of `integer difference`, so I think we can remove the second. 
```suggestion
              integer_arithmetic=True,
```

marcogario (approved):
Thanks for the patch. This looks good. I will take a look on why the tests are failing, but I agree with your comment.

pysmt/logics.py
johnyf: Thank you for reviewing. The following definitions of logics may be relevant to defining only one of the attributes `integer_arithmetic` and `integer_difference` as `True`:

https://github.com/pysmt/pysmt/blob/b248cdc66e8ada4a602292869cbe24dab4ea8b25/pysmt/logics.py#L457-L464

https://github.com/pysmt/pysmt/blob/b248cdc66e8ada4a602292869cbe24dab4ea8b25/pysmt/logics.py#L533-L540

I could update the pull request with a new commit with the suggested change (and the commits can be squashed upon merging), or force-push an updated commit, whichever is preferred.

I could also add a commit that defines only `integer_arithmetic=True` (and leaves `integer_difference` to the default value of 0) for the logics `QF_IDL` and `QF_UFIDL`, if it is decided relevant for this pull request to change their definitions.

marcogario: Thanks for checking. This is correct like this then. The assumption is then that we want the more specific `integer_difference`.

johnyf:
Thank you for merging this pull request.

PR #662, opened 2020-10-20
Switched from nosetests to pytest

### LINKED ISSUE #661: Nosetests end-of-life
Nosetests does not work on Python 3.9 (and I was unable to get it working also on 3.8.6 on Windows). (See: https://tirkarthi.github.io/programming/2020/01/27/python-39-changes.html)

I think we shall switch to pytest like https://github.com/elastic/beats/pull/16883...

mikand:
Fixes #661

### DIFF
--- a/health_check.sh
+++ b/health_check.sh
@@ -1,6 +1,4 @@
 #!/bin/bash
 
 export PYTHONDONTWRITEBYTECODE=True
-nosetests -A "not slow" \
-          --with-xunit \
-          --with-coverage --cover-html --cover-package=pysmt
+python3 -mpytest -m "not slow"
--- a/health_check.sh
+++ b/health_check.sh
@@ -1,4 +1,0 @@
-#!/bin/bash
-
-export PYTHONDONTWRITEBYTECODE=True
-python3 -mpytest -m "not slow"
--- a/pysmt/solvers/z3.py
+++ b/pysmt/solvers/z3.py
@@ -147,14 +147,14 @@ class Z3Solver(IncrementalTrackingSolver, UnsatCoreSolver,
                                            environment=environment,
                                            logic=logic,
                                            **options)
-        # try:
-        #     self.z3 = z3.SolverFor(str(logic))
-        # except z3.Z3Exception:
-        #     self.z3 = z3.Solver()
-        # except z3.z3types.Z3Exception:
-        #     self.z3 = z3.Solver()
-        # except OSError:
-        self.z3 = z3.Solver()
+        try:
+            self.z3 = z3.SolverFor(str(logic))
+        except z3.Z3Exception:
+            self.z3 = z3.Solver()
+        except z3.z3types.Z3Exception:
+            self.z3 = z3.Solver()
+        except OSError:
+            self.z3 = z3.Solver()
         self.options(self)
         self.declarations = set()
         self.converter = Z3Converter(environment, z3_ctx=self.z3.ctx)
--- a/pysmt/solvers/z3.py
+++ b/pysmt/solvers/z3.py
@@ -147,14 +147,14 @@ class Z3Solver(IncrementalTrackingSolver, UnsatCoreSolver,
                                            environment=environment,
                                            logic=logic,
                                            **options)
-        try:
-            self.z3 = z3.SolverFor(str(logic))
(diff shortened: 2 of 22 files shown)


health_check.sh
```diff
@@ -1,6 +1,4 @@
 #!/bin/bash
```
marcogario, line 1: Let's drop this file. We were fetching more stats (e.g., coverage), but it seems we do not run it as part of CI. Easier to just drop it for now.

pysmt/solvers/z3.py
```diff
@@ -147,14 +147,14 @@ def __init__(self, environment, logic, **options):
...
-        except OSError:
-            self.z3 = z3.Solver()
+        # try:
+        #     self.z3 = z3.SolverFor(str(logic))
+        # except z3.Z3Exception:
+        #     self.z3 = z3.Solver()
+        # except z3.z3types.Z3Exception:
+        #     self.z3 = z3.Solver()
+        # except OSError:
+        self.z3 = z3.Solver()
```
marcogario, line 157: This does not seem related to the pytest transition. Was this related to how we resolve the import of z3?

pysmt/test/smtlib/test_parser_qf_arrays.py
```diff
@@ -16,10 +16,16 @@
...
+
 from pysmt.test.smtlib.parser_utils import execute_script_fname, SMTLIB_TEST_FILES, SMTLIB_DIR
 
-def test_generator():
+def _get_tests():
     for (logic, f, expected_result) in SMTLIB_TEST_FILES:
         smtfile = os.path.join(SMTLIB_DIR, f)
         if logic.theory.arrays:
-            yield execute_script_fname, smtfile, logic, expected_result
+            yield smtfile, logic, expected_result
```
marcogario, line 27: It seems we do not need a generator anymore. What about moving this into `pysmt.test` as :

```python
def smtlib_tests(logic_pred):
    tests = []
    for (logic, f, expected_result) in SMTLIB_TEST_FILES:
        smtfile = os.path.join(SMTLIB_DIR, f)
        if logic_pre(logic):
            tests.append(smtfile, logic, expected_result)
```

pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
+markers =
+    slow: marks tests as slow (deselect with '-m "not slow"')
```
marcogario, line 3: Maybe add "(typically > 10 seconds)", to explain what we consider slow?

run_tests.sh
```diff
@@ -17,13 +17,11 @@
...
-#export NOSE_PROCESSES=-1
-#export NOSE_PROCESS_TIMEOUT=240
 export PYTHONDONTWRITEBYTECODE=True
 
-# Skip slow tests (-A "not slow")
+# Skip slow tests (-m "not slow")
 # Exit on error (-x)
 # Rule of thumb: if a test takes more than 10 seconds it
 #                should be marked as slow using:
 #                    @attr("slow")
```
marcogario, line 26:
```suggestion
#                    @pytest.mark.slow
```

marcogario (approved):
LGTM.

There are some opportunities for refactoring the generators for the tests, that would be good to tackle now.

pysmt/solvers/z3.py
mikand: I think this was the source of the damn segfault on the CI... But you are right, it is unrelated. Shall we move to a separate PR or keep it?

marcogario: Let's move it out and make sure this is indeed the reason of the CI segfault. I also do not remember why that code might have been there in the first place, so a more thorough review would be prefereable.

health_check.sh
mikand: OK

pysmt/solvers/z3.py
mikand: OK

pysmt/test/smtlib/test_parser_qf_arrays.py
mikand: Done.

run_tests.sh
mikand: OK

PR #663, opened 2020-10-21
Black code-style enforcement

mikand:
This PR reformatted all the sources using black (https://github.com/psf/black) and imposes a strict coding rule: all the code must be formatted using black before committing it (A CI job ensures that the code is formatted according to black).

- [x] Reformat all pysmt code
- [x] Create a CI job checking that code is adherent to black rules
- [x] Write documentation on this new coding rule and on how to use `pre-commit` to check the code before committing


mikand:
@marcogario Documentation is still missing, but how do you like this?

marcogario (commented):
I am not really impressed by the result. It seems to me that being strict about the line length will have the side effect of promoting shorter named variables, thus reducing readability.

Breaking function calls with many arguments to multi lines makes it harder for me to read. I think this is a matter of getting used to it, but the typical way we align things means that your eye can scan the left side of the page and see the flow of the code. Now, instead, I am presented with a bunch of details I do not really care about when skimming the code. Again, this might encourage people to not use named arguments to keep the function call shorter.

I marked a few other points where the readability is impaired.

Overall, I am not writing in this codebase as much as I would like, so if you think this improves things, I am ok with the change.

Would it make sense to disable it for the docs/ folder, or you think this would make it harder to run it?

mikand:
Closing as wontfix

PR #645, opened 2020-05-23
 CLA: Add utility to check for CLA

marcogario:
We require all contributors to submit a change to CONTRIBUTORS as a way to confirming acceptance of the CLA.

To check this, we compare the git log with the entries in the CONTRIBUTORS file.

Since users might have sent commits with different names/emails, we introduce a `.mailmap` to provide canonical names.

The CONTRIBUTORS data has been updated with the content of CLAHub.

We seem to be missing a few:
image

Require CLA:
* Caleb Donovick b0f031d9f595dddbd4febd3dd3bad487f3e112d9
* en.magnago 2b9a879ac9fbe9253d05655b3212dd0a13a37273

Typos in documentation (might consider whitelisting these changes)
* Guillem Francès 6562e2662496291a5578facecb140d7f41f8ab4d
* Matthew Fernandez 39e6d252239d63472b2a28a901bd0bcce42237dd

PR #665, opened 2020-10-26
Update CONTRIBUTORS

cdonovick:
as requested here https://github.com/pysmt/pysmt/pull/645#issuecomment-633516108f


mikand:
@cdonovick Thanks!

PR #667, opened 2020-11-23
Integer div

EnricoMagnago:
This should improve the support for integer division.

In formula.py:
* I have removed an unnecessary exception that was raised if the rhs of a division was an integer constant.
* The warning "division by 0" is raised also in the case of integer division.

In simplifier.py:
* I think the simplification was buggy in python3 since the `/` operator performs division between rationals.
* In addition the standard C-like integer division is not consistent with the smtlib2 semantics.
As far as I understand the semantics of division between integers of smtlib2 is as follows:
`n / d` is `floor(float(n) / d)` if `d > 0` and `ceil(float(n) / d)` if `d < 0`.
http://smtlib.cs.uiowa.edu/theories-Ints.shtml

In msat.py:
* an integer division is rewritten in terms of `floor` and `divide` operations, where `divide` is the division between reals.

In test_nia.py:
* added some simple test cases for the integer division. I don't know if there was already a test file performing similar tests on NIA.

### DIFF
--- a/pysmt/formula.py
+++ b/pysmt/formula.py
@@ -262,7 +262,9 @@ class FormulaManager(object):
 
     def Div(self, left, right):
         """ Creates an expression of the form: left / right """
-        if right.is_constant(types.REAL, 0) and self.env.enable_div_by_0:
+        if (right.is_constant(types.REAL, 0) or
+            right.is_constant(types.INT, 0)) \
+           and self.env.enable_div_by_0:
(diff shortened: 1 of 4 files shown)


EnricoMagnago:
@mikand could you please have a look and let me know if you see any problem?    
Thanks!

mikand:
@EnricoMagnago LGTM, thanks!

PR #670, opened 2020-12-02
Added missing mailmap

mikand:


mikand:
@EnricoMagnago Can you please comment here stating that the mailmap I inserted is yours? (This is needed for confirming that you agree with the CLA)

EnricoMagnago:
Yes it is mine.

PR #671, opened 2020-12-18
Bug in times distributor

EnricoMagnago:
I found a simple bug in times distributor.   
These commits should fix the issue and add a simple test case. 

The current implementation of TimesDistributor rewrites the inequality `(x - (x + y)) < 0` into `(x + (-1 * x)) < 0`.


EnricoMagnago:
@mikand can you have a look please?

mikand:
@EnricoMagnago, LGTM. Thanks!

PR #675, opened 2021-01-14
Fix "get_model" when called from a generic solver (Fix #674)

### LINKED ISSUE #674: Generic Solver's "get_model" crashes
The generic solver's "get_model" crashes when some variables doesn't appear in the formula leading to a behavior different from the one given when using solvers' native API. Here an example

```python
env = get_env()

# Add the solver to the environment
env.factory.add_generic_solver("mathsat", path, logics)

r, s = Symbol("r", REAL), Symbol("s", REAL)
p, q = Symbol("p", INT), Symbol("q", INT)

f_lra = GT(r, s)
f_idl = GT(p, q)

print(get_model(f_lra & f_idl , solver_name="mathsat")) # works fine as all environment variables are in the formula
print(get_model(f_idl, solver_name="mathsat")) # this crushes as variables "r" and "s" doesn't appear in the formula
```

In the last "get_model", a `pysmt.exceptions.PysmtSyntaxError` is raised from `get_assignment_list`.

btwael:
This fixes the problem raised in this issue #674. "get_model" of generic solver fails because it call "get_value" for all variables in the environment, this may cause an error as some variables do not appear in the formula.

An error is raised when the parser fails to extract the value of a variable (doesn't appear in the formula) from the response of the generic solver.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -30,5 +30,6 @@ Suresh Gl <sureshgl009@gmail.com>
 Taylor Hodge <j.taylor.hodge@gmail.com>
 Varun Patro <varun.kumar.patro@gmail.com>
 Vijay Raghavan <vijaysgr8@gmail.com>
+Wael-Amine Boutglay <btwael@gmail.com>
 Yoni Zohar <yoni206@gmail.com>
 Yuri Malheiros <yurimalheiros@gmail.com>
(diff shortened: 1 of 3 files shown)


marcogario:
:wave: @btwael !

Thanks for the fix! It looks good, but would you mind adding the example from the issue (https://github.com/pysmt/pysmt/issues/674) as regression test?

marcogario (approved):
LGTM. Thanks!

PR #676, opened 2021-01-20
Generic interpolator (based on interpolation proposal for smtlib)

btwael:
This allow to define a generic interpolator based on the the following [proposal](https://ultimate.informatik.uni-freiburg.de/smtinterpol/proposal.pdf).

It was tested with SMTInterpol. Many other solvers don't follow this proposal. Or only follow some of it (opensmt by example: usi-verification-and-security/opensmt#213).

Remarque: For `QF_Lia`, SMTInterpol may produce interpolant with `div` or `mod`, this can be added to the smtlib parser but this is not included in this pull request.

### DIFF
--- a/pysmt/smtlib/commands.py
+++ b/pysmt/smtlib/commands.py
@@ -32,6 +32,7 @@ EXIT='exit'
 GET_ASSERTIONS='get-assertions'
 GET_ASSIGNMENT='get-assignment'
 GET_INFO='get-info'
+GET_INTERPOLANTS = 'get-interpolants'
 GET_MODEL='get-model'
 GET_OPTION='get-option'
 GET_PROOF='get-proof'
@@ -82,4 +83,8 @@ SMT_LIB_2_5 = SMT_LIB_2_0 + [
     RESET_ASSERTIONS,
 ]
 
-ALL_COMMANDS = SMT_LIB_2_5
+EXTENSIONS = [
+    GET_INTERPOLANTS
+]
+
+ALL_COMMANDS = SMT_LIB_2_5 + EXTENSIONS
--- a/pysmt/smtlib/interpolator.py
+++ b/pysmt/smtlib/interpolator.py
@@ -1,0 +1,87 @@
+#
+# This file is part of pySMT.
+#
+#   Copyright 2014 Andrea Micheli and Marco Gario
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+#
+
+import pysmt.smtlib.commands as smtcmd
+from pysmt.exceptions import SolverReturnedUnknownResultError, UnknownSolverAnswerError
+from pysmt.smtlib.script import SmtLibCommand
+from pysmt.smtlib.solver import SmtLibSolver
+from pysmt.solvers.interpolation import Interpolator
+
+
+class SmtLibInterpolator(Interpolator):
+
+    def __init__(self, args, environment, logic, LOGICS=None, **options):
+        Interpolator.__init__(self)
+        options["produce_interpolants"] = True
+        self.options = (args, environment, logic, LOGICS, options)
+
+    def binary_interpolant(self, a, b):
+        """Returns a binary interpolant for the pair (a, b), if And(a, b) is
+        unsatisfaiable, or None if And(a, b) is satisfiable.
+
+        """
+        res = self.sequence_interpolant([a, b])
+        if res is None:
+            return res
+        return res[0]
+
+    def sequence_interpolant(self, formulas):
+        """Returns a sequence interpolant for the conjunction of formulas, or
+        None if the problem is satisfiable.
+
+        """
+        solver = SmtLibSolver(self.options[0], self.options[1], self.options[2], self.options[3], **self.options[4])
+        for f in formulas:
+            sorts = solver.to.get_types(f, custom_only=True)
+            for s in sorts:
+                if all(s not in ds for ds in solver.declared_sorts):
+                    solver._declare_sort(s)
+            deps = f.get_free_variables()
+            for d in deps:
+                if all(d not in dv for dv in solver.declared_vars):
+                    solver._declare_variable(d)
+
+        names = []
+        i = 0
+        for f in formulas:
+            self._send_named_assert_command(f, "IP_%d" % i, solver)
+            names.append("IP_%d" % i)
+            i += 1
+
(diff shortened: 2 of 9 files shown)


pysmt/smtlib/solver.py
```diff
@@ -96,6 +96,9 @@ def __init__(self, args, environment, logic, LOGICS=None, **options):
             self.solver_stdout = TextIOWrapper(self.solver.stdout)
 
         # Initialize solver
+        if "produce_interpolants" in options and options["produce_interpolants"]:
+            self.set_option(":print-success", "true")
```
marcogario, line 100: is the `:print-success` needed for the interpolation work or was this for debugging?

pysmt/smtlib/parser/parser.py
```diff
@@ -1094,6 +1094,23 @@ def get_assignment_list(self, script):
         self.cache.unbind_all(symbols)
         return res
 
+    def get_interpolant_list(self, script):
```
marcogario, line 1097: could you add a docstring comment simliar to the ones for the other methods?

pysmt/smtlib/interpolator.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class SmtLibInterpolator(Interpolator):
+
+    def __init__(self, args, environment, logic, LOGICS=None, **options):
+        Interpolator.__init__(self)
+        options["produce_interpolants"] = True
+        self._solver = SmtLibSolver(args, environment, logic, LOGICS, **options)
+
+    def binary_interpolant(self, a, b):
```
marcogario, line 33: This is a new interface. Could you add a comment on both `binary_interpolant` and `sequence_interpolant` on how to use them?

pysmt/smtlib/commands.py
```diff
@@ -66,6 +67,7 @@
     GET_PROOF,
     GET_UNSAT_CORE,
     GET_INFO,
+    GET_INTERPOLANTS,
```
marcogario, line 70: This is not really part of SMTLIB, as the extension is not part of the standard. What if we move this to a new list called `EXTENSIONS` and then define `ALL_COMMANDS` below as `SMT_LIB_2_5 + EXTENSIONS` ?

pysmt/smtlib/interpolator.py
```diff
@@ -0,0 +1,76 @@
...
+            raise SolverReturnedUnknownResultError
+        else:
+            raise UnknownSolverAnswerError("Solver returned: " + ans)
+        self._solver._send_command(SmtLibCommand(smtcmd.GET_INTERPOLANTS, names))
+        return self._solver.parser.get_interpolant_list(self._solver.solver_stdout)
+
+    def _send_named_assert_command(self, f, assert_name):
+        cmd = SmtLibCommand(smtcmd.ASSERT, [f])
+        cmd.assert_name = assert_name
+        self._solver._send_silent_command(cmd)
```
marcogario, line 73: If named assertions are not part of SMTLIB, let's add `NAMED_ASSERT` as a proper command, and avoid the `hasattr` check in the `SmtLibCommand.serialize` method.

marcogario (commented):
Thanks for the feature improvement!

I think this looks good. We need some tests for this feature. Ideally, a script to setup SMTInterpol with one or more tests for each of the interpolation methods you added.

You can take a look at `pysmt/test/smtlib/test_generic_wrapper.py` to see how we test the generic wrapper. Something along those lines would call SMTInterpol and run some simple tests. See `pysmt/test/test_interpolation.py` for the interpolation tests we have for MathSAT. The new tests could just re-run the same against the generic interpolator.

You can either use MathSAT or assume that SMTInterpol is installed in CI, no need to extend `pysmt-install` to install SMTInterpol (although that would also be cool). We can then extend the CI do download SMTInterpol (see how we do for MathSAT here https://github.com/pysmt/pysmt/blob/80cb7f958e53fc3e71682c7b6460da1e4d8dbaa0/ci/install_unix.sh#L121).

I can help with the CI, if you need.

btwael:
Hello @marcogario, thank you for the reviews, I will consider all the changes you suggested.

My changes still has a problem when interpolation is called multiple times, this is due the fact that `_get_answer` returns an empty result while `success` is expected (this is probably related to how interpolants are parsed, for the moment I guess that some new line `\n` at the end of the output are not consumed).

btwael:
The failing CI checks are not related to my changes; they are probably related the configuration with Pip and Python2.

pysmt/smtlib/interpolator.py
btwael: I remember in the standard they have named expressions, but I will check the standard SMTLIB 2 to find what is the best way to have names (assertion or expression) with minimal changes.

btwael: To name `<expr>`, We must use (<symbol> is the name) `(! <expr> :named <symbol>)`.
A more generic way is to define attributed expressions with the following syntax `(! <expr> <attribute>+)`

pysmt/smtlib/parser/parser.py
```diff
@@ -1094,6 +1094,27 @@ def get_assignment_list(self, script):
...
+        """
+        Parse an list of interpolants produced by get-interpolants
```
marcogario, line 1099:
```suggestion
        Parse a list of interpolants produced by get-interpolants
```

pysmt/smtlib/solver.py
```diff
@@ -96,6 +96,8 @@ def __init__(self, args, environment, logic, LOGICS=None, **options):
             self.solver_stdout = TextIOWrapper(self.solver.stdout)
 
         # Initialize solver
+        if "produce_interpolants" in options and options["produce_interpolants"]:
```
marcogario, line 99:
```suggestion
        if options.get("produce_interpolants"):
```

marcogario (commented):
If you add a test in `test/smtlib/test_generic_wrapper.py` I can help you flush out the last bug and get this past the finish line.

marcogario (approved):

marcogario (commented):
I think the test using smtinterpol is still missing. I can try to add it in the next few days, but otherwise the changes look good.

btwael:
Hello, sorry for not adding that, I was a little bit busy. Both SMTInterpol and OpenSMT 2 (master branch) can be used.

I din't also figure out how to better use named expressions (for the moment named assertions is sufficient), this must be addressed later in a different pull request as it requires a lot of changes to the FNode structure.

marcogario:
>  Both SMTInterpol and OpenSMT 2 (master branch) can be used.

Great!

> I didn't also figure out how to better use named expressions (for the moment, named assertions are sufficient). I will address that in a future pull request as it requires a lot of changes to the FNode structure.

Changes to the FNode structure need a very strong reason, as that is the key structure we use for representing an expression, so the impact of adding a bit there is huge.

I am not sure how named expressions should behave. I assume these are annotated expressions in SMT-LIB with a `name` (or similar) annotation. What we can do is to add an object that keeps track of the `FNode->Name` mapping. If this is very useful, we could fit that object within the `Environment`. This should probably work for most use-cases, with the caveat that in pySMT two occurrences of the same expression are represented by the same object. This could be a problem if you wanted to have one occurrence of  `A & B` mapped to `Name1` and a difference occurrence of `A & B` mapped to `Name2`.

We can move this discussion to a dedicated issue if you find you really need named expressions.

PR #679, opened 2021-01-23
implement add_assertions method for solver.

EnricoMagnago:
I often end up writing something like:
```
for constr in iterable:
     solver.add_assertion(constr)
```
so I thought having a method in solver to add all constraints from an iterable could be useful.    
This does not do anything clever like trying to optimise the addition of multiple constraints.

Another possibility could be to replicate what is done in the FormulaManager for methods such as `And` and `Or`, hence having a single method accepting both a single instance and an iterable.

### DIFF
--- a/pysmt/solvers/solver.py
+++ b/pysmt/solvers/solver.py
@@ -349,9 +349,9 @@ class IncrementalTrackingSolver(Solver):
         self._assertion_stack.append(tracked)
         self._last_command = "assert"
 
-    def add_assertions(self, formulae, names=None):
-        for idx, formula in enumerate(formulae):
-            self.add_assertion(formula, names[idx] if names else None)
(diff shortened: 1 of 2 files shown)


marcogario:
:wave: Doing `add_assertion(And(iterable))` is semantically equivalent. Is this not the case for what you want to achieve?

> Another possibility could be to replicate what is done in the FormulaManager for methods such as And and Or, hence having a single method accepting both a single instance and an iterable.

I am concern this could lead to non-intuitive behavior. I'd rather stick with a dedicated method.

Thanks for the PR, I am generally ok with this change, although I would remove the option to specify names, as it seems providing a dictionary would be a better interface to have a map of formulae-names. We should also add a test.

marcogario (approved):

PR #680, opened 2021-01-26
Added assumptions into unsat cores when solving under assumptions is used

mikand:
This is a bug fix.

The assumptions passed to `solve()` were silently dropped from the unsat
core. Asserting `x | y` and calling `solve([!x, !y])` returned a core of
`{x | y}`, which is satisfiable, so the core did not explain the
unsatisfiability it was supposed to explain.

### The expectation this PR settles

`get_unsat_core()` reports the assumptions of the failing `solve()` call
alongside the assertions, in both the `"all"` and the `"named"` unsat
cores mode.

`get_named_unsat_core()` does not. Assumptions are not added via
`add_assertion(..., named=...)` and so carry no name; the earlier
revision of this PR invented `_a_<n>` keys for them, but those keys are
assigned while iterating a `set`, so they are neither stable across runs
nor traceable back to anything by the caller. A caller who wants the
assumptions asks `get_unsat_core()` for them.

This is the same split SMT-LIB draws between `get-unsat-core` and
`get-unsat-assumptions`. The contract is now written down on
`UnsatCoreSolver`, `Solver.solve` and `Solver.add_assertion`, which is
where it was missing.

### Also fixed

Two Boolector bugs that the above uncovered:

* Assumptions given as an iterator rather than a list were consumed by
  the conversion loop and stored as an empty list, so they never reached
  the core.
* Once that was fixed, every `get_unsat_core()` call under assumptions
  raised `BoolectorException: 'exp' must be an assumption`. `_solve()`
  re-`Assume()`s the assertion stack right after `Sat()`, and Boolector
  discards failed-assumption information on the next `Assume()`. The
  core is now extracted inside `_solve()`, before that loop.
* `get_named_unsat_core()` zipped the `Failed()` results of the named
  assertions against `self._assertion_stack`, pairing names with the
  wrong assertions whenever the two lists differed.

### Tested

`test_assumptions_in_unsat_core` and `test_assumptions_as_iterator` run
against every available unsat-core solver in both modes. Verified
locally with z3, Boolector and OptiMathSAT; the three now agree, which
they did not before.


pysmt/solvers/btor.py
```diff
@@ -230,9 +231,11 @@ def get_model(self):
 
     @clear_pending_pop
     def _solve(self, assumptions=None):
+        self._last_assumptions = []
         if assumptions is not None:
             btor_assumptions = [self.converter.convert(a) for a in assumptions]
             self.btor.Assume(*btor_assumptions)
+            self._last_assumptions = list(assumptions)
```
marcogario, line 238: If assumption is an iterator, we will consume it in L234, and then L238 will return an empty list.
Shall we do `list(assumptions)` earlier?

pysmt/solvers/btor.py
```diff
@@ -191,6 +191,7 @@ def __init__(self, environment, logic, **options):
         self.mgr = environment.formula_manager
         self.declarations = {}
         self._named_assertions = {}
+        self._last_assumptions = []
```
marcogario, line 194: Maybe just `assumptions`? We cannot have multiple assumptions at the same time, right? That's why we are satisfied with storing only one set of assumptions.

pysmt/solvers/btor.py
```diff
@@ -258,9 +261,10 @@ def get_unsat_core(self):
             unsat_core = set()
             # relies on this assertion stack being ordered
             assert isinstance(self._assertion_stack, list)
-            btor_assertions = [self.converter.convert(a) for a in self._assertion_stack]
+            lst = self._assertion_stack + self._last_assumptions
```
marcogario, line 264: Maybe:

```suggestion
            assertions = self._assertion_stack + self._last_assumptions
```

pysmt/solvers/btor.py
```diff
@@ -275,13 +279,16 @@ def get_named_unsat_core(self):
         if self.options.unsat_cores_mode == "named":
             unsat_core = {}
             # relies on this assertion stack being ordered
+            lst = list(self._named_assertions.keys()) + self._last_assumptions
```
marcogario, line 282: As above, let's use a more meaningful name here.

pysmt/solvers/btor.py
```diff
@@ -275,13 +279,16 @@ def get_named_unsat_core(self):
...
             assert isinstance(self._assertion_stack, list)
-            btor_named_assertions = [self.converter.convert(a) for a in self._named_assertions.keys()]
+            btor_named_assertions = [self.converter.convert(a) for a in lst]
             in_unsat_core = self.btor.Failed(*btor_named_assertions)
-            for a, in_core in zip(self._assertion_stack, in_unsat_core):
+            cnt = 0
+            for a, in_core in zip(lst, in_unsat_core):
                 if in_core:
-                    name = self._named_assertions[a]
+                    name = self._named_assertions.get(a, '_a_%d' % cnt)
```
marcogario, line 289: We are making up the name for the assumption? If so, how is the caller supposed to link this information back? I wonder if in the named case returning the assumptions makes sense.

marcogario (approved):
LGTM.

could you update the PR description to clarify whether this is a bug or not?
Is the expectation that we always return assumptions in unsat cores or not?

I have some doubts about returning assumptions in the named unsat core case, but I do not see any harm in keeping this.

pysmt/solvers/btor.py
mikand: Good catch, and it was worse than it looked. Fixed by materializing first:

```python
self._assumptions = [] if assumptions is None else list(assumptions)
```

The empty list was also masking a second bug. Once the assumptions actually reached `get_unsat_core()`, every Boolector call under assumptions started raising `BoolectorException: 'exp' must be an assumption`: `_solve()` re-`Assume()`s the assertion stack right after `Sat()`, and Boolector discards failed-assumption information on the next `Assume()`. The core is now extracted inside `_solve()`, before that loop, which also removes the `Failed()` plumbing from both getters.

There is a regression test for the iterator case (`test_assumptions_as_iterator`).

pysmt/solvers/btor.py
mikand: Done, renamed to `self._assumptions`. And yes, only one set at a time: it is overwritten at the top of every `_solve()`.

pysmt/solvers/btor.py
mikand: The local is gone entirely. Since the core now has to be extracted inside `_solve()` (see the iterator thread above), `get_unsat_core()` reduces to reading the two sets computed there:

```python
if self.options.unsat_cores_mode == 'all':
    return set(self._core) | set(self._core_assumptions)
else:
    return set(self.get_named_unsat_core().values()) | \
        set(self._core_assumptions)
```

pysmt/solvers/btor.py
mikand: Same here, the local is gone. This branch also had a latent bug: it zipped the `Failed()` results of `self._named_assertions` against `self._assertion_stack`, so names were paired with the wrong assertions whenever the two lists differed. It is now a single comprehension over the extracted core.

pysmt/solvers/btor.py
mikand: You were right to push on this, the made-up names do not work. They are assigned while iterating a `set`, so they are not even stable between two runs of the same query, and the caller has nothing to map them back to.

So assumptions no longer appear in `get_named_unsat_core()` at all. Only assertions added via `add_assertion(..., named=...)` can be reported there. Assumptions are reported by `get_unsat_core()`, in both `"all"` and `"named"` mode. That is the same split SMT-LIB draws between `get-unsat-core` and `get-unsat-assumptions`, and it is applied consistently to z3, MathSAT and Boolector, which did not agree with each other before.

On your question in the review body: this is a bug fix, and the PR description now states the expectation. The old behavior returned a core that was itself satisfiable, so it did not explain the unsatisfiability it was supposed to explain.

The contract was written down nowhere, which is what made the question unanswerable from the docs, so the last commit documents it on `UnsatCoreSolver`, `Solver.solve`, `Solver.add_assertion` and `unsat_cores_mode`. While doing that I found the docs disagreed with reality twice: `named=` is honored in `"all"` mode by z3 but replaced with a generated name by Boolector and MathSAT, and the README listed Boolector as not supporting unsat cores at all. Both noted/fixed.

If you want the link-back to be explicit rather than implied, a public `get_unsat_assumptions()` mirroring SMT-LIB would now be about three lines per solver, since the private helpers exist. I left it out as nobody has asked for it. Happy to add it if you prefer.

PR #681, opened 2021-02-08
SmtLib model validation support

mikand:
This PR introduces the needed machinery to parse the result of a `(get-model)` command in SMTLib.

It also adds support for empty symbol names.


pysmt/fnode.py
```diff
@@ -121,13 +121,14 @@ def simplify(self):
         """Return a simplified version of the formula."""
         return _env().simplifier.simplify(self)
 
-    def substitute(self, subs):
+    def substitute(self, subs, interpretations=None):
         """Return a formula in which subformula have been substituted.
 
         subs is a dictionary mapping terms to be subtituted with their
         substitution.
```
marcogario, line 128: Can you extend the comment on interpretations? Something along the lines of:
```suggestion
        subs is a dictionary mapping terms to be subtituted with their
        substitution.
        interpretations is a dictionary mapping function symbols to an FunctionInterpretation objects describing the semantics of the function.
```

pysmt/smtlib/parser/parser.py
```diff
@@ -865,6 +868,69 @@ def get_command_generator(self, script):
             yield cmd
         return
 
+    def parse_model(self, script):
+        """This function pasres the result of a `(get-model)` command and
```
marcogario, line 872: This docstring is very clear! :sparkles:

pysmt/smtlib/parser/parser.py
```diff
@@ -865,6 +868,69 @@ def get_command_generator(self, script):
...
+
+        cmd_gen = self.get_command(tokens)
+        model, interpretation = {}, {}
+        while current != ")":
+            if (current != "("):
+                raise PysmtSyntaxError("'(' expected", tokens.pos_info)
+            tokens.add_extra_token(current)
+            cmd = next(cmd_gen)
+            if cmd.name != 'define-fun':
+                raise PysmtSyntaxError("Unsupported model command type: %s" % cmd.name)
```
marcogario, line 916: Maybe:

```suggestion
                raise PysmtSyntaxError("Unsupported model command: %s" % cmd.name)
```

pysmt/smtlib/parser/parser.py
```diff
@@ -865,6 +868,69 @@ def get_command_generator(self, script):
...
+            vname, formal, rtype, ebody = cmd.args
+            if len(formal) == 0: # Constant assignment
+                model[mgr.Symbol(vname, rtype)] = ebody
```
marcogario, line 919: `ebody` is an FNode, right? How do we distinguish between `Int` and `Real`?

pysmt/smtlib/utils.py
```diff
@@ -0,0 +1,57 @@
...
+            l = sl.constant_value()
+            r = sr.constant_value()
+            return self.manager.Bool(l == r)
+        elif sl == sr:
+            return self.manager.TRUE()
+        elif sl.is_symbol() and sr.is_symbol() and \
+             sl.symbol_name().startswith('@') and \
+             sr.symbol_name().startswith('@'):
+            return self.manager.Bool(sl == sr)
+
```
marcogario, line 55:
```suggestion
```

pysmt/smtlib/utils.py
```diff
@@ -0,0 +1,57 @@
...
+        if sl.is_constant() and sr.is_constant():
+            l = sl.constant_value()
+            r = sr.constant_value()
+            return self.manager.Bool(l == r)
+        elif sl == sr:
+            return self.manager.TRUE()
+        elif sl.is_symbol() and sr.is_symbol() and \
+             sl.symbol_name().startswith('@') and \
+             sr.symbol_name().startswith('@'):
+            return self.manager.Bool(sl == sr)
```
marcogario, line 54: The case at line 49 covers the positive case of `sl` and `sr` both being the same object.
I wonder if changing the condition here to the negative case is better. Indeed, one could argue that the "magic" of `@` is adding an extra condition for things to be non-equal.

 ```suggestion
        elif sl == sr:
            return self.manager.TRUE()
        elif sl.is_symbol() and sr.is_symbol() and \
             sl.symbol_name().startswith('@') and \
             sr.symbol_name().startswith('@'):
            return self.manager.FALSE()
```

File: pysmt/solvers/btor.py [hunk 7]
```diff
@@ -389,10 +389,11 @@ def walk_symbol(self, formula, **kwargs):
             res = self._btor.Array(self._btor.ArraySort(self._btor.BitVecSort(index_type.width),
                                                         self._btor.BitVecSort(elem_type.width)),
                                    formula.symbol_name())
-        else:
-            assert symbol_type.is_bv_type()
+        elif symbol_type.is_bv_type():
             res = self._btor.Var(self._btor.BitVecSort(formula.bv_width()),
                                  formula.symbol_name())
+        else:
+            raise SolverReturnedUnknownResultError("Unknown type fro BTOR")
```
marcogario, line 396:
```suggestion
            raise SolverReturnedUnknownResultError("Unknown type for BTOR")
```

pysmt/substituter.py
```diff
@@ -20,7 +20,73 @@
...
+        special symbols (e.g. @val1) in SmtLib.
+        """
+        if any(not x.is_symbol() or not x.is_term() for x in formal_params):
+            raise PysmtValueError('Formal parameters of a function '
+                                  'interpretation must be non-function symbols')
+        if not allow_free_vars and \
+           not function_body.get_free_variables().issubset(set(formal_params)):
+            raise PysmtValueError('the body of a function interpretation cannot'
+                                  ' contain free varibales other than formal '
+                                  'paramteters')
```
marcogario, line 73:
```suggestion
                                  'parameters')
```

pysmt/substituter.py
```diff
@@ -20,7 +20,73 @@
...
+        because functions with uninterpreted return value return
+        special symbols (e.g. @val1) in SmtLib.
+        """
+        if any(not x.is_symbol() or not x.is_term() for x in formal_params):
+            raise PysmtValueError('Formal parameters of a function '
+                                  'interpretation must be non-function symbols')
+        if not allow_free_vars and \
+           not function_body.get_free_variables().issubset(set(formal_params)):
+            raise PysmtValueError('the body of a function interpretation cannot'
+                                  ' contain free varibales other than formal '
```
marcogario, line 72:
```suggestion
                                  ' contain free variables other than formal '
```

pysmt/substituter.py
```diff
@@ -20,7 +20,73 @@
...
+
+    def __init__(self, formal_params, function_body, allow_free_vars=False):
+        """Constructor, taking in input the list of formal parameters and the
+        function body.
+
+        The parameter `allow_free_vars` is used to skip the check that
+        the function body has no free variables other than formal
+        parameter and is used in the SmtLib model-validation utility
+        because functions with uninterpreted return value return
+        special symbols (e.g. @val1) in SmtLib.
```
marcogario, line 64: We are not enforcing that these symbols must start with `@`, are we?

marcogario (approved):
LGTM. Left a few comments mostly for my understanding.

Docs look great!

pysmt/smtlib/parser/parser.py
mikand: :1st_place_medal:

pysmt/smtlib/utils.py
mikand: Not sure to follow the rationale: the underlying assumption is that two symbols **with different names** must correspond to different objects, but the same name must mean the same thing.

Consider the following case: the interpretation of `f1(x)` is `ITE(x > 0, @s1, @s2)` and the interpretation of `f2(y)` is `ITE(y > 0, @s1, @s3)`.
Clearly, a term like `Equals(f1(10), f2(10))` should simplify to true.

Does this make sense?

pysmt/substituter.py
mikand: We are not. In the model-validation pipeline this is enforced by the fact that non-@-symbols are not simplified and thus the model is recognized as invalid.

Do you think we shall enforce this here?

pysmt/smtlib/parser/parser.py
mikand: Excellent point! Solved in 0469170

mikand:
@marcogario I fixed most of the comments. Can you have a look at the two points left?

pysmt/smtlib/utils.py
marcogario: Isn't `Equals(@s1, @s1)` covered by line 49? They are the same symbol, independently of the name. There is no special logic due to the `@`. Indeed L54 returns `self.manager.Bool(sl == sr)`, rather than checking the name itself `sr.symbol_name() == sl.symbol_name()`. The equality of the symbol should be guaranteed by the memoization of the FNodes.

I am fine with keeping the logic as you wrote it if you find it more clear to read/explain.

pysmt/substituter.py
marcogario: I think this is ok as is.

PR #683, opened 2021-02-09
Fix to correctly pass logic to solvers started by Portfolio

ekilmer:
Thank you for this project!

I encountered an issue while using `Portfolio` where the solvers weren't respecting the passed `logic`. I found that they were using the default logic ([`QF_UFLIRA`](https://github.com/pysmt/pysmt/blob/22520e2a012c980285fae10fe8aa388e263ea317/pysmt/factory.py#L47)), and these changes were required in order to start each solver with the correct logic, as passed to `Portfolio`.


marcogario (approved):
:wave: 

Thanks for preparing this patch. Looks good!

PR #684, opened 2021-02-09
Remove six and python 2 compatibility code

marcogario:
We officially [dropped support for python 2 in 0.9.0 (2020-04-26)](https://github.com/pysmt/pysmt/blob/master/docs/CHANGES.rst#090-2020-04-26----pysmt-as-module).

We now remove the dependency from six and all its uses in the code. This means that pySMT now does not work anymore with python 2!

This change is motivated by python 2 EOL and aimed at reducing maintenance overhead.

Thanks to the six.py team for the awesome work!


mikand:
LGTM, but you also need to fix `examples`, in particular `examples/smtlib.py`

mikand (approved):
LGTM

PR #688, opened 2021-03-22
Added Psuedo-Boolean for Z3

junichiro-kishi:
Thanks for contributing to pySMT!

To whom it may concern,

I'm Junichiro Kishi, a master's student who belongs Nagoya University, Japan.
Our research group has developed [CombSQL+](https://www.trs.css.i.nagoya-u.ac.jp/projects/CombSQLplus/), which is a combinatorial optimization solver based on SQL, and we use pySMT on CombSQL+ to use multiple SMT solvers on the same interface.

I added Psuedo-Boolean interface for Z3 to pySMT to speed up CombSQL+'s execution speed.

Because this is my first time contribution to another project, I may have made various mistakes. (but I intend to check it properly...)
We appreciate your feedbacks.

Sincerely, Junichiro Kishi.

Here are good things to check:

- [-] For bugs or new features, make sure there is a test showing the bug or demonstrating the use of the new feature
- [-] Are all tests green?
- [-] (First contribution only) You accepted the licensing terms by adding name and email to the CONTRIBUTORS file (see https://github.com/pysmt/pysmt/blob/master/docs/development.rst#licensing).

### DIFF
--- a/pysmt/fnode.py
+++ b/pysmt/fnode.py
@@ -44,9 +44,10 @@ from pysmt.operators import (FORALL, EXISTS, AND, OR, NOT, IMPLIES, IFF,
                              STR_TO_INT, INT_TO_STR,
                              STR_CHARAT,
                              ARRAY_SELECT, ARRAY_STORE, ARRAY_VALUE,
-                             ALGEBRAIC_CONSTANT)
+                             ALGEBRAIC_CONSTANT,
+                             PBLE, PBGE, PBEQ)
 
-from pysmt.operators import  (BOOL_OPERATORS, THEORY_OPERATORS,
+from pysmt.operators import  (BOOL_OPERATORS, PSEUDO_BOOLEAN, THEORY_OPERATORS,
                               BV_OPERATORS, IRA_OPERATORS, ARRAY_OPERATORS,
                               STR_OPERATORS,
                               RELATIONS, CONSTANTS)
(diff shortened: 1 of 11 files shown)


mikand:
Dear @junichiro-kishi thanks for using pysmt and for the contribution!

If I understand correctly, you added three new operators to represent the atoms `a + b + c op k` where `a`, `b` and `c` are boolean variables, `op` is `>=`. '<=' or `=` and k is a constant because Z3 is able to natively deal with such constraints efficiently.

Since these operators are specific to Z3, an easier route would have been to recognize a pseudo-boolean atom in the `Z3Converter` and convert it into a PB constraint on-the-fly. The issue with the current approach is that one needs to implement these constraints for all the other solvers (+ all the rewritings and simplifications). This is why all the tests are currently failing.

I think that recognizing a sum of `Ite(x, 1, 0)` where x is a boolean variable in the `equals`, `le` and `ge` cases of the `Z3Converter` would be much simpler and cleaner. Then, we can easily provide the `PBXX` constructors that work on all the solvers.

@marcogario what do you think?

marcogario:
:wave: @junichiro-kishi ! 

I agree with @mikand that this route might not be the best. The key problem I see is that psuedo-boolean is not part of the SMT-LIB. We have some cases in which we implemented extensions of the SMT-LIB (e.g., interpolation or optimization), but I am not aware of any SMT-LIB extension describing pseudo-boolean.

This is evident in [examples.py](https://github.com/pysmt/pysmt/pull/688/files#diff-fcaf30e35e253808475cf98cc5374e132350fb6766561e31df2e23511da443b4R143), where the associated logic for the example is defined as `QF_BOOL`, but this is not strictly true as that logic does not include the pseudo-boolean operators.

I think it is worth exploring the option of detecting this construct in the Z3Converter, and if this turns out to be difficult, or fragile, we can consider having an extension of `QF_BOOL` to support this, but then we should check which solvers have native support for this operators and which ones would require rewriting.

mikand:
Closing due to inactivity. Feel free to open a new PR if this is still relevant.

PR #691, opened 2021-04-08
Fixed SMTLib Model Validation

mikand:
See https://github.com/SMT-COMP/postprocessors/issues/10 for a description

### DIFF
--- a/pysmt/smtlib/parser/parser.py
+++ b/pysmt/smtlib/parser/parser.py
@@ -513,6 +513,14 @@ class SmtLibParser(object):
                 return res
             stack[-1].append(handler)
         else:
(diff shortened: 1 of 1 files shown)


jhoenicke:
From our side, this pull request does what we need.

However, I wonder if this is acceptable for pySMT.  At least for our parser we want that parsing a script and printing it again should more or less yield the same result.  Wouldn't this patch change the name of the model variable also when printing it?

There's also the theoretical problem of someone using `@x_y` and `(@as @x y)` as two different identifiers, but I guess since you generally don't want to work around name-clashes, and since `@` is reserved for model-specific things, it is not a problem.

PR #695, opened 2021-06-29
Fix iss694

### LINKED ISSUE #694: msat named unsatcore might ignore non-boolean assumption
When using mathsat for named unsat core computation, it might happen that some assumption given to `solve` is ignored.
The attached file should show the issue: the query should be unsat but a model is returned.

[test.py.txt](https://github.com/pysmt/pysmt/files/6733560/test.py.txt)
 
As far as I understood the source of the issue is the following.
* in the `_solve` method in msat.py, let `assumptions` contain a non-boolean, user specified, assumption.
* such assumption will end up in the `other_ass` list, and added to the solver via `add_assertion`.
* `_add_assertion` of msat.py for named unsat cores introduces a new boolean symbol `key` and adds the assertion `key -> formula`.
* the control returns to the `_solve` method, which does not add the newly introduces key as an assumption.

Then the solver can assign false to the newly introduced key, obtaining a model that does not satisfy the user specified assumption.

EnricoMagnago:
This PR should fix #694

### DIFF
--- a/pysmt/solvers/msat.py
+++ b/pysmt/solvers/msat.py
@@ -238,33 +238,30 @@ class MathSAT5Solver(IncrementalTrackingSolver, UnsatCoreSolver,
     @clear_pending_pop
     def _solve(self, assumptions=None):
         res = None
+        bool_ass = []
+        other_ass = []
+        for x in assumptions if assumptions is not None else []:
+            if x.is_literal():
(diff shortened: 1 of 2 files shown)


mikand:
@EnricoMagnago Thanks for this, it looks perfect (tests are failing due to another issue with CVC4)

PR #699, opened 2021-07-26
Remove call to `FNode.substitute` in SmtLibExecutionCache

EnricoMagnago:
Do not use `FNode.substitute` in SmtLibExecutionCache: the method relies on the substituter in the environment obtained via `get_env`, this might not match the environment given to the SmtLibParser and cause errors.

I also applied the following minor changes.
Remove `\` within parentheses.
Remove duplicated call to `Dict.get`.


mikand:
LGTM! Thanks @EnricoMagnago!

PR #700, opened 2021-07-27
Fix env usage

enmag:
I tried to make the naming of `environment` more uniform: now it should always appear as `env`.    
This breaks backward compatibility since a call like `pippo(environment=my_env)` should be modified to `pippo(env=my_env)`.

In addition, I tried to reduce the number of places where the code ends up calling `get_env` to obtain an environment.
I am using multiple environments and this behavior was causing some issues.
In some classes and methods I have added an optional `env` parameter so that they should work within the given environment if any.
I should have "fixed" al the calls to `get_free_variables`, `simplify` and `substitute`, and most of the calls to `get_type`.


marcogario:
:wave: Is this something that we still want to do? 

If so, I would probably separate the "cosmetic" breaking change of renaming `environment` to `env` from the "correctness" change of avoiding the global environment when there is one that is more appropriate.

We can also look into making the rename non-breaking by introducing a secondary keyword argument, but I am not sure this is worth the effort.

mikand:
I second @marcogario: the aesthetic fix is secondary compared to add the customizability of the environment.

I think very few users are using custom environments, so the breaking change should not be too disruptive

EnricoMagnago:
At the moment I don't exactly remember the status of this PR.
I believe I might have already changed "all" `environment` to `env`.

In the next few days I will try to check its status and update it.

EnricoMagnago:
It appears there is at least 1 test case failing: yices on macos.

I am unable to reproduce the error on my linux machine.

Any suggestions?
@mikand

mikand:
Thanks @enmag, and sorry this sat for so long.

The branch is now 409 commits behind master and no longer applies, so rather than rebase it I have opened #839, which ports the changes to the current master. Following the review here from @marcogario and myself, #839 carries **only the correctness part** — using the given environment instead of the global one — and leaves out the `environment` → `env` rename, which is the breaking, cosmetic half and can be done on its own later.

Nothing from this PR had landed in the meantime, and master has since grown new instances of the same problem (`cvcfive.py`, `optimization/`, more of `rewritings.py`), so #839 covers those as well. It also adds regression tests that fail on master, which should settle the kind of question the yices-on-macOS failure raised here.

Closing this in favour of #839. Your work is the basis for it — please do review it if you have the time.

PR #703, opened 2021-11-26
Added printing of annotations to smt lib printers

agirardi-fbk:
Added support of SMT-LIB annotations printing

### DIFF
--- a/pysmt/smtlib/printers.py
+++ b/pysmt/smtlib/printers.py
@@ -190,8 +190,8 @@ class SmtPrinter(TreeWalker):
         yield formula.arg(0)
         self.write(")")
 
-    @write_annotations
     @handles(op.BV_ROR, op.BV_ROL)
+    @write_annotations
     def walk_bv_rotate(self, formula):
         if formula.is_bv_ror():
             rotate_type = "rotate_right"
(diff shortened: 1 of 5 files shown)


pysmt/smtlib/printers.py
```diff
@@ -23,14 +23,45 @@
...
+        if self.annotations is not None and formula in self.annotations:
+            res = ["(! "]
+            res.append(f(self, formula, *args, **kwargs))
+            for key, values in self.annotations[formula].items():
+                res.append(f" :{key}")
+                for value in values:
+                    res.append(f" {value}")
+            res.append(")")
+            return ''.join(res)
+        return f(self, formula, *args, **kwargs)
```
johnyf, line 54: A comment on nesting and indentation: by complementing the predicate of the `if` statement, eight lines can be moved to a smaller indentation level, and the maximal nesting depth of syntactic constructs reduced from 5 to 4. Together with other changes:

```python
def write_annotations_dag(f):
    def resf(self, formula, *arg, **kwargs):
        annots = self.annotations
        app = f(self, formula, *args, **kwargs)
        if annots is None or formula not in annots:
            return app
        items = list()
        for key, values in annots[formula].items():
            items.append(f' :{key}')
            items.extend(f' {v}' for v in values)
        kv = ''.join(items)
        return f'(! {app}{kv})'
```

mikand: Thanks for the contribution @johnyf!
@agirardi-fbk can you have a look and update the code / comment here?

agirardi-fbk: Thank you for your suggestion @johnyf, I have updated the pull request

johnyf: Thank you all for reviewing these changes, and updating the pull request.

mikand (approved):

PR #705, opened 2022-01-18
Integer div

EnricoMagnago:
In #667 I wrongly defined the integer division as based on the sign of the numerator.    
However, in https://smtlib.cs.uiowa.edu/theories-Ints.shtml it is defined depending on the sign of the denominator.
This should fix this issue.


mikand:
:+1: Thanks!

PR #714, opened 2022-04-13
Add possibility to use several BV operators as left associative

agirardi-fbk:
This pull request adds the possibility to use several BV operators as left-associative, affected operators are BVAnd, BVOr, BVAdd, and BVMul.
The nodes themselves still only have two children, the formula manager was updated to allow passing an arbitrary number of operators after the left and right operator. The generated formula is created in a left-associative manner as indicated in https://smtlib.cs.uiowa.edu/theories-FixedSizeBitVectors.shtml.

This also allows pySMT to read SMT-LIB scripts using these BV operators with more than two children.

A test was added in pysmt/test/test_formula.py demonstrating the new behavior.


mikand (approved):
Excellent work, thanks @agirardi-fbk

PR #723, opened 2022-06-11
Upgrade Z3 to 4.8.17

marcogario:
Upgrades z3 to the latest release. This does not include the logic necessary for picking the arm release.

I see that z3 has been publishing the python packages to pypi, so I wonder if we should try to simplify the solver to just use the python package (pip install).

### DIFF
--- a/pysmt/cmd/install.py
+++ b/pysmt/cmd/install.py
@@ -35,7 +35,7 @@ INSTALLERS = [
     Installer(MSatInstaller,    "5.6.7", {}),
     Installer(CVC4Installer,    "1.7-prerelease",
               {"git_version" : "391ab9df6c3fd9a3771864900c1718534c1e4666"}),
-    Installer(Z3Installer,      "4.8.7", {"osx": "10.14.6"}),
+    Installer(Z3Installer,      "4.8.17", {"osx": "10.16"}),
     Installer(YicesInstaller,   "2.6.2",
               {"yicespy_version": "f0768ffeec15ea310f830d10878971c9998454ac"}),
     Installer(BtorInstaller,    "3.2.1", {}),


marcogario:
I pushed a commit with some debug print on the z3 back conversion. I focused on the smtlib path, in which we ask Z3 to print the SMT-LIB of the formula and then use our parser (I renamed the parameter in the test to make it clearer).

The problem is that the expression we get back adds a `(to_real 0)` to an INT expression.

```
Z3: SMTLIB
(set-info :status unknown)
(declare-fun p () Int)
(assert
 (let ((?x12 (^ p 2)))
 (= ?x12 (to_real 0))))
(check-sat)
```

I think this is a bug in Z3, but it might also be that they are more relaxed with the types here and treat everything like a real for simplicity. The same SMT-LIB file is parsed by Z3 CLI without problems.

marcogario:
The following:

```python
from z3 import *

p = Int('p')
f = p ** 2
print(f.is_real(), f.is_int())
g = (f == 0)
print(g.sexpr())
```

Outputs

```
True False
(= (^ p 2) 0.0)
```

It seems `Power` has signature `INT x INT -> REAL`. I cannot find the definition in the SMT-LIB. Can it be that this is an extension of Z3?

It does make sense if we do not consider Natural numbers, but only Integers: as p ** (-1) is not necessarily INT.

I think we should revisit our implementation of `Pow`. @mikand WDYT?

mikand:
In our implementation, we made some assumptions (see: https://github.com/pysmt/pysmt/blob/master/pysmt/type_checker.py#L324) so we have a general `REAL x REAL -> REAL` and a `INT x INT -> INT` that ssumes that the exponent is a positive constant.

Probably the signature of Z3 is more sensible, as `2^(-1)` is not an integer. I do not think this operator has been used a lot and I also cannot find it in SMTLib, so Ok in changing the typing for me.

pysmt/solvers/z3.py
```diff
@@ -595,6 +595,8 @@ def back_via_smtlib(self, expr):
                                              None, None,
                                              0, None,
                                              expr.ast)
+        print("Z3: SMTLIB")
+        print(s)
```
mikand, line 599: Leftover print?

pysmt/type_checker.py
```diff
@@ -33,6 +33,8 @@ class SimpleTypeChecker(walkers.DagWalker):
 
     def __init__(self, env=None):
         walkers.DagWalker.__init__(self, env=env)
+        # Return None if the type cannot be computed rather than
+        # raising an exception.
```
mikand, line 37:
```suggestion
        # If `be_nice` is true, the `get_type` method will return None if 
        # the type cannot be computed instead of than raising an exception.
```

mikand (approved):
@marcogario LGTM, I left two minor comments

pysmt/solvers/z3.py
marcogario, line 599:
```suggestion
```

pysmt/test/test_nlira.py
```diff
@@ -55,8 +55,15 @@ def test_irrational(self):
             model = s.get_model()
             xval = model[x]
             self.assertTrue(xval.is_algebraic_constant())
-            approx = Fraction(-3109888511975, 2199023255552)
-            self.assertEqual(xval.algebraic_approx_value(), approx)
+            # There are two solutions, thad only differ by sign
```
marcogario, line 58:
```suggestion
            # There are two solutions, that only differ by sign
```

PR #726, opened 2022-07-08
Parse logic str in Portfolio

### LINKED ISSUE #708: TypeError: '<=' not supported between instances of 'str' and 'Logic' (PySMT==0.9.1.dev139)
On PySMT==0.9.1.dev139, running one of the examples from the README (below), I get the following error: 

```
  File "/home/user/del.py", line 6, in <module>
    with Portfolio(["cvc4",
  File "/home/user/.local/lib/python3.9/site-packages/pysmt/shortcuts.py", line 977, in Portfolio
    return pf.Portfolio(solvers_set=solvers_set,
  File "/home/user/.local/lib/python3.9/site-packages/pysmt/solvers/portfolio.py", line 80, in __init__
    all_solvers = set(self.environment.factory.all_solvers(logic=logic))
  File "/home/user/.local/lib/python3.9/site-packages/pysmt/factory.py", line 381, in all_solvers
    return self._filter_solvers(self._all_solvers, logic=logic)
  File "/home/user/.local/lib/python3.9/site-packages/pysmt/factory.py", line 361, in _filter_solvers
    if logic <= l:
TypeError: '<=' not supported between instances of 'str' and 'Logic'
```

Here is file I was running: 

```python
from pysmt.shortcuts import Portfolio, Symbol, Not

x, y = Symbol("x"), Symbol("y")
f = x.Implies(y)

with Portfolio(["cvc4",
                "yices"],
               logic="QF_UFLIRA",
               incremental=False,
               generate_models=False) as s:
  s.add_assertion(f)
  s.push()
  s.add_assertion(x)
  res = s.solve()
  v_y = s.get_value(y)
  print(v_y) # TRUE

  s.pop()
  s.add_assertion(Not(y))
  res = s.solve()
  v_x = s.get_value(x)
  print(v_x) # FALSE
```

I am on Ubuntu 21.10, Python 3.9.7, and here is the output from `pysmt-install --check`

```
Installed Solvers:
  msat      True (5.6.1)             Not in Python's path!
  cvc4      True (1.7-prerelease)    
  z3        True (4.8.7)             
  yices     True (2.6.2)             
  btor      True (3.2.1)             
  picosat   True (965)               
  bdd       False (None)              Not in Python's path!

Solvers: z3, cvc4, yices, picosat, btor
Quantifier Eliminators: z3, shannon, selfsub
UNSAT-Cores: z3, btor
Interpolators: 
```

marcogario:
Portfolio does not rely on the factory infrastructure, but instead we
instantiate the class directly. In the factory module we take care of
converting the logic from str to a pysmt.logic object.
This step was missing in the Portfolio constructor.

We could do the conversion later, but I think it makes sense to ensure
that logic is always a pysmt object.

Fixes #708.

### DIFF
--- a/pysmt/solvers/portfolio.py
+++ b/pysmt/solvers/portfolio.py
@@ -20,6 +20,7 @@ from multiprocessing import Process, Queue, Pipe
 
 from pysmt.solvers.solver import IncrementalTrackingSolver, SolverOptions
 from pysmt.decorators import clear_pending_pop
+from pysmt.logics import convert_logic_from_string
 
 
 LOGGER = logging.getLogger(__name__)
@@ -70,6 +71,7 @@ class Portfolio(IncrementalTrackingSolver):
 
         One process will be used for each of the solvers.
         """
+        logic = convert_logic_from_string(logic)
         IncrementalTrackingSolver.__init__(self,
                                            environment=environment,
                                            logic=logic,


pysmt/solvers/portfolio.py
```diff
@@ -70,6 +71,7 @@ def __init__(self, solvers_set, environment, logic, **options):
 
         One process will be used for each of the solvers.
         """
+        logic = convert_logic_from_string(logic)
```
mikand, line 74: Would it make sense to accept both a `str` or a `Logic` as `logic`?

Something like:
```python
if isinstance(logic, str):
   logic = convert_logic_from_string(logic)
```

marcogario: That is what the helper does internally. Maybe the name should make it more explicit:
```python
def convert_logic_from_string(name):
    """Helper function to parse function arguments.

    This takes a logic or a string or None, and returns a logic or None.
    """
    if name is not None and isinstance(name, str):
        name = get_logic_by_name(name)
    return name
```

PR #728, opened 2022-07-08
Walker functions removal

### LINKED ISSUE #410: Remove Walker.functions
Complete the deprecation of instance-based walkers by removing the Walker.functions dictionary.

marcogario:
The `functions` dictionary in the `Walker` was used to map each nodetype to the corresponding `walk` method. We called this "instance based walkers" because each instance could (in principle) associate different functions.

We changed this behavior long time ago (0.6), in favor of "class based walkers". The idea being that the walk method will follow the name of the node type: e.g., `walk_add` will be used for the `add` operator.

This PR removes the last uses of the `functions` dictionary.

TODO:
- [ ] The way I changed the DWF implementation does not allow the use of multiple environments (while the previous version did). It probably makes sense to explore if we want to keep this behavior.
- [ ] There are some tests for the walkers to update as they were testing the `functions` dictionary. It might make sense to update those tests to show the same behavior using the new approach.

Fixes https://github.com/pysmt/pysmt/issues/410

### DIFF
--- a/pysmt/walkers/generic.py
+++ b/pysmt/walkers/generic.py
@@ -77,15 +77,6 @@ class Walker(object, metaclass=MetaNodeTypeHandler):
             env = pysmt.environment.get_env()
         self.env = env
 
-        self.functions = {}
-        for o in op.all_types():
-            try:
-                # getattr will raise an AttributeError exception if a
-                # method does not exist
-                self.functions[o] = getattr(self, nt_to_fun(o))
-            except AttributeError:
-                self.functions[o] = self.walk_error
-
     def set_function(self, _, *__):
         raise NotImplementedError(
             "Instance-based walkers (<=0.6.0) walkers are deprecated."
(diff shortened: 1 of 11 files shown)


pysmt/walkers/generic.py
```diff
@@ -77,24 +77,10 @@ def __init__(self, env=None):
...
-        node_types with the given function
-        """
-        from warnings import warn
-        warn("Instance-based walkers (<=0.6.0) walkers are deprecated. "
-             "You should use new-style/class based walkers", stacklevel=2)
-        for nt in node_types:
-            self.functions[nt] = function
+    def set_function(self, _, *__):
+        raise NotImplementedError(
+            "Instance-based walkers (<=0.6.0) walkers are deprecated."
```
marcogario, line 82:
```suggestion
            "Instance-based walkers (<=0.6.0) are deprecated."
```

mikand:
Superseded by #849

Issue #601, opened 2019-08-23
Normalize expressions over commutative operators


marcogario:
+, *, =, and & are all examples of commutative operations. In pySMT we treat the expression `a = b` as a different expression than `b = a`. This is because we do not try to normalize the order of the operands. Since we have a total order between expressions (due to the node id), it would be simple to normalize such operators.

The main advantages would be:
1. Reduced memory use: A leaf expression that is not correctly memoized, can lead to an entire different subtree that is however equivalent
2. More predictable ways to perform substitution. We currently need to accont for both `a = b` and `b = a`.

EnricoMagnago:
I have a piece of code that tries to do that. In attachment the class implementation.
[canonize.txt](https://github.com/pysmt/pysmt/files/5860703/canonize.txt)

I am using the tuple `(x.is_constant(), x.node_type(), x.node_id())` as key.   
Across different runs this gives me a little more consistency than using just the node id (in case the nodes are generated in a non-deterministic order).

I also required some normalisation over expressions, so I have a class `Ineq` that performs a kind of linear simplification of expressions to write equivalent inequalities in the same way.
However, I think this might be out of the scope of this issue.

PR #730, opened 2022-07-16
reorder operands of commutative operators based on node-id.

enmag:
#601


EnricoMagnago:
@marcogario,
please have a look and check if the changes meet the requirements you had in mind.

pysmt/formula.py
```diff
@@ -86,6 +86,11 @@ def _do_type_check(self, formula):
         self._do_type_check = self._do_type_check_real
         return self._do_type_check(formula)
 
+    def _sort_nodes(self, nodes):
+        if len(nodes) == 1 and isinstance(next(iter(nodes)), Iterable):
+            nodes = nodes[0]
```
marcogario, line 91: In which cases do you need this?

pysmt/formula.py
```diff
@@ -86,6 +86,11 @@ def _do_type_check(self, formula):
...
+        return sorted(nodes, key=lambda p: p.node_id())
```
marcogario, line 92: Could this method return directly a tuple?
It seems that the main way we use it is `tuple(_sort_nodes(...))`.

pysmt/formula.py
```diff
@@ -233,7 +238,7 @@ def Times(self, *args):
          - Arguments must be all of the same type
          - Arguments must be INT or REAL
         """
-        tuple_args = self._polymorph_args_to_tuple(args)
+        tuple_args = self._polymorph_args_to_tuple(self._sort_nodes(args))
```
marcogario, line 241: can you push the call to `_sort_nodes` within `polymorph_args_to_tuple`?

marcogario (commented):
Thanks for putting this together. It looks good to me.

I left some questions for understanding whether we can push further the ergonomics of this change. These are not blocking.

Do we have any data on the performance impact of doing this? I think a simple evaluation of how long it takes to parse some SMT-LIB files would be enough. `SMT-LIB/parse_all.py` is a script to do this type of thing.

Thanks!

EnricoMagnago:
I have run the modified `parse_all` script on the 93 smt2.bz2 files in `pysmt/test/smtlib` and also on the 1753 QF-LRA  benchmarks from https://clc-gitlab.cs.uiowa.edu:2443/SMT-LIB-benchmarks/QF_LRA.

I run them on my personal laptop so there could be some variability between different runs.

Each of the pysmt/test/smtlib benchmarks was executed 10 times (via timeit) and I considered the total amount of time required.
Each benchmark from QF-LRA was run only once.

[data.zip](https://github.com/pysmt/pysmt/files/9223368/data.zip) archive with the data I obtained and script used to generate the plots and compute ratios.

The average ratio between the time required without re-ordering (main branch) and the one required by this branch is:
**0.9733** in the case of the pysmt/test/smtlib benchmarks and
**0.9825** for the QF-LRA benchmarks. 
Therefore, a slowdown between 2% and 3%.

pysmt/test/smtlib scatter plot:
pysmt_test_smtlib
QF-LRA scatter plot:
qf_lra


Other details:
in the `parse_all` script I catch and mask all exceptions, in both groups of benchmarks there are inputs that pysmt fails to parse.
In particular these are the 29 inputs that failed in the QF-LRA benchmarks (I have not investigated the reason for the failures).
'QF_LRA/2019-ezsmt/Labyrinth/SCC_Strong/0011-labyrinth-12-0.smt2'
'QF_LRA/2019-ezsmt/Labyrinth/SCC_Strong/0067-labyrinth-18-0.smt2'
'QF_LRA/2019-ezsmt/blending/19.smt2'
'QF_LRA/2019-ezsmt/Labyrinth/SCC_Strong/0110-labyrinth-16-0.smt2'
'QF_LRA/2019-ezsmt/Labyrinth/SCC_Strong/0198-labyrinth-19-0.smt2'
'QF_LRA/2019-ezsmt/Labyrinth/SCC_Strong/0084-labyrinth-17-0.smt2'
'QF_LRA/2019-ezsmt/Labyrinth/SCC_Strong/0051-labyrinth-12-0.smt2'
'QF_LRA/2019-ezsmt/Labyrinth/SCC_Strong/0059-labyrinth-13-0.smt2'
'QF_LRA/2019-ezsmt/Labyrinth/SCC_Strong/0175-labyrinth-15-0.smt2'
'QF_LRA/2019-ezsmt/Labyrinth/SCC_Strong/0036-labyrinth-17-0.smt2'
'QF_LRA/2019-ezsmt/blending/25.smt2'
'QF_LRA/2019-ezsmt/Labyrinth/SCC_Strong/0092-labyrinth-17-0.smt2'
'QF_LRA/2019-ezsmt/Labyrinth/SCC_Strong/0100-labyrinth-17-0.smt2'
'QF_LRA/2019-ezsmt/blending/20.smt2'
'QF_LRA/2019-ezsmt/Labyrinth/SCC_Strong/0143-labyrinth-14-0.smt2'
'QF_LRA/2019-ezsmt/blending/21.smt2'
'QF_LRA/2019-ezsmt/blending/10.smt2'
'QF_LRA/2019-ezsmt/Labyrinth/SCC_Strong/0150-labyrinth-16-0.smt2'
'QF_LRA/2019-ezsmt/Labyrinth/SCC_Strong/0206-labyrinth-20-0.smt2'
'QF_LRA/2019-ezsmt/Labyrinth/SCC_Strong/0075-labyrinth-13-0.smt2'
'QF_LRA/2019-ezsmt/Labyrinth/SCC_Strong/0028-labyrinth-12-0.smt2'
'QF_LRA/2019-ezsmt/Labyrinth/SCC_Strong/0132-labyrinth-12-0.smt2'
'QF_LRA/2019-ezsmt/Labyrinth/SCC_Strong/0159-labyrinth-20-0.smt2'
'QF_LRA/2019-ezsmt/Labyrinth/SCC_Strong/0018-labyrinth-11-0.smt2'
'QF_LRA/2019-ezsmt/Labyrinth/SCC_Strong/0165-labyrinth-13-0.smt2'
'QF_LRA/2019-ezsmt/Labyrinth/SCC_Strong/0003-labyrinth-11-0.smt2'
'QF_LRA/2019-ezsmt/Labyrinth/SCC_Strong/0043-labyrinth-13-0.smt2'
'QF_LRA/2019-ezsmt/routingMax/SCC_strong/p50.56.smt2'
'QF_LRA/2019-ezsmt/blending/23.smt2'

EnricoMagnago:
I have investigated a bit the failing test case.
It happens non-deterministically on my machine on the input at example.py:672.
It seems the issue is related to the detection of the logic.

In particular when the test case passes I observe the following:
```
Formula:
  (((q = (p - 3)) -> ((p < ack0) | (ack1 <= p))) & (((r = r) & (p = (q + 3))) -> (ack0 = ack1)))

Theory:
  Arrays: False, ArraysConst: False, BV: False, FP: False, IA: True, RA: True, ID: False, RD: True, Linear: True, EUF: False, Type: False, String: False

Filtered logics:
  [QF_AUFBVLIRA*, QF_NIRA, UFLIRA, QF_UFLIRAt, QF_AUFBVLIRAt, QF_UFLIRA, QF_AUFBVLIRA, QF_NIRAt, UFLIRAt]
Min logic: QF_UFLIRA
```

Instead when it fails:
```
Formula:
  (((q = (p - 3)) -> ((p < ack0) | (ack1 <= p))) & (((r = r) & (p = (q + 3))) -> (ack0 = ack1)))
Theory:
  Arrays: False, ArraysConst: False, BV: False, FP: False, IA: True, RA: True, ID: False, RD: True, Linear: True, EUF: False, Type: False, String: False
  
Filtered logics:
  [QF_NIRA, QF_AUFBVLIRA*, QF_UFLIRAt, UFLIRAt, QF_UFLIRA, UFLIRA, QF_NIRAt, QF_AUFBVLIRA, QF_AUFBVLIRAt]
Min logic: QF_NIRA
```

The difference seems to be in the order of the `filtered_logics`, which corresponds to the list of pysmt-supported logics greater than the requested one computed at logic.py:798.
The different orderings are possible since `PYSMT_LOGICS` is a `fronzenset`.
Then the min logic is computed by applying `min` on the list.
However both `QF_NIRA < QF_UFLIRA` and `QF_UFLIRA < QF_NIRA` are false.

Creating `QF_LIRA` and adding it to the `PYSMT_LOGICS` set appears to solve this issue.    
However, it does not solve the problem of having a partial definition of `<`.

@marcogario @mikand could you please make sure I am not missing something?
If my analysis is correct the problem is unrelated to the current changes. Have you observed anything similar on the main branch?

SMT-LIB/parse_all.py
```diff
@@ -53,27 +54,39 @@ def main():
...
         files_cnt = args.count
-    print("Submitting %d jobs, %d at the time" % (files_cnt, chunks))
-    timings = p.map(execute_script_fname, islice(file_list, files_cnt), chunks)
+
+    timings = []
+    errors = []
+
+    with multiprocessing.Pool() as p:
+        print(f"Submitting {files_cnt} jobs, {chunks} at the time")
+        for el in p.starmap(execute_script_fname,
```
marcogario, line 76: Minor: Wouldn't `repeat(zip(islice(...)), args.reps)` achieve the same and avoid requiring passing `reps` to `execute_script_fname`.
It seems this is the only reason you are extending `execute_script_fname` with reps.

SMT-LIB/parse_all.py
```diff
@@ -1,44 +1,45 @@
...
 def dump_stats(timings, fname):
     if fname is None:
         fname = "stats.out"
-    with open(fname, "w") as f:
+    with open(fname, mode='w', encoding="utf-8") as f:
         f.write('filename, time\n')
-        for k in timings:
-            f.write('%f, "%s"\n' % k)
+        for val, bench_name in timings:
+            f.write(f'"{bench_name}", {val}\n')
```
marcogario, line 42: FYI, fstrings are python >=3.6. If we decide to ignore older versions of python (and considering 3.5 is EoL I am generally positive) we should probably document this more clearly.

I think no change is required here as this is an util and not part of library itself.

SMT-LIB/parse_all.py
```diff
@@ -53,27 +54,39 @@ def main():
...
+                                repeat(args.reps)),
+                            chunks):
+            (timings if el[0] > 0 else errors).append(el)
 
     mean = sum(x[0] for x in timings) / len(timings)
-    print("The mean execution time was %0.2f seconds" % mean)
-    print("The max execution time was %0.2f seconds" % max(x[0] for x in timings))
+    print(f"The mean execution time was {mean:.2f} seconds")
+    print(f"The max execution time was {max(x[0] for x in timings):.2f} seconds")
+    print(f"Found {len(errors)} errors: {[el[1] for el in errors]}")
```
marcogario, line 85:
```suggestion
    print(f"Found {len(errors)} errors in : {[x[1] for x in errors]}")
```

We are using `el` above as the return of a single call. In the line above we use `x` as iterator. I'd use `x` here too for consistency.

pysmt/formula.py
```diff
@@ -1099,17 +1095,22 @@ def normalize(self, formula):
             self._normalizer = FormulaContextualizer(self.env)
         return self._normalizer.walk(formula)
 
-    def _polymorph_args_to_tuple(self, args):
+    def _sorted_tuple(self, nodes):
+        return tuple(sorted(nodes, key=lambda p: p.node_id()))
+
+    def _polymorph_args_to_tuple(self, args, sort=False):
```
marcogario, line 1101: In what cases do we want this NOT to be sorted? I'd expect this method is always called with `sort=True`, but I see a few instances that do not. Those instances are then not guaranteed to create the same formula.

marcogario (commented):
Thanks for updating the PR. It looks good.
I left a few minor comments. The only thing that I'd like to understand before approving is the need for the option `sort` argument to `polymorph_args_to_tuple`.

Thank you for running the eval. I think the performance difference is acceptable, as we gain more predictability and (hopefully) memory re-use in some cases.

I'll now take a closer look at the failing test that you are describing and post another comment if I get any insight.

marcogario:
> Then the min logic is computed by applying min on the list.
> However both QF_NIRA < QF_UFLIRA and QF_UFLIRA < QF_NIRA are false.

Yes, and this is correct as they are not comparable.

> However, it does not solve the problem of having a "partial" definition of <.

The partial definition of `<` is correct. Indeed both QF_NIRA and QF_UFLIRA are valid answers to the question "what is the closest logic that can solve QF_LIRA problems?".
The tests fail because QF_NIRA support is limited. 

> Have you observed anything similar on the main branch?
> If my analysis is correct the problem is unrelated to the current changes.

This started failing recently, but it is indeed unrelated. See https://github.com/pysmt/pysmt/issues/731

> Creating QF_LIRA and adding it to the PYSMT_LOGICS set appears to solve this issue.

I think that is a reasonable fix, but I wonder if we can fix the test instead.

pysmt/formula.py
enmag: I believe in cases such as `StrConcat`, `BVConcat` we cannot sort the arguments.

In addition, at the moment I have not applied the sorting to all possible operators. e.g. I have omitted BV operators.
It should be possible to sort those as well. I can them.

I decided to add the extra optional argument to avoid duplicating the code to get the tuple from either a n-ary args or a single list argument.

An approach to avoid the extra arguments would be to define a function to get the list of arguments `_polymorph_args_to_lst(self, args)` and then define two variants `_args_to_tuple(self, args)` and `_args_to_sorted_tuple(self, args)`.

marcogario: > I believe in cases such as StrConcat, BVConcat we cannot sort the arguments.

You are right, makes sense, as we use `_polymorph_args_to_tuple` for all n-ary expressions.

> In addition, at the moment I have not applied the sorting to all possible operators. e.g. I have omitted BV operators. It should be possible to sort those as well. I can them.

It would be good to have all operators use the same approach. I am good if you want to do this as a follow-up though.

> I decided to add the extra optional argument

The extra argument is totally fine. As this is not semantic preserving in some cases (:point_up:), I also think the default is the correct one.

marcogario (approved):
LGTM. Up to you whether to include BVs in this PR or not.

enmag:
I believe it make sense to do it in this PR.

However, it seems that expressions on BV keep the recursive structure of the formula.
They build the left-associative binary tree representing the operations (e.g. `BVAnd`, `BVOr`, `BVXor` ...).
Test case `test/test_formula.py:test_left_associative_bv` explicitly tests that the sub-trees are generated using left-associativity on the argument list.

Was there a specific reason for this or can I go ahead and change it?

marcogario:
Those changes were introduced in https://github.com/pysmt/pysmt/pull/714. I would need to dig a bit deeper to understand how to proceed. @mikand any insight?

mikand:
@enmag Tahnk you so much for all this work. Concerning #714, we needed a quick way to support n-ary BV operators and we did not change the internal PySMT representation, we simply generalized the factory method. 

Making these BV operators really n-ary is totally possible, but touches virtually any walker (and solver) working with BVs.

enmag:
I believe I could do that in a reasonable amount of time.
I can add those changes to this PR, or, if you prefer, I can make another PR.

Do we want to sort the operands of BV operators as well or just flatten the representation?

mikand:
Since we are here, I vote for sorting also the BV operators... @marcogario?

enmag:
There seems to be a non-deterministic failure when running the `pysmt/test/smtlib/test_parser_qf_lra.py` tests using cvc5.
The following is the error I get on my laptop:
```bash
pysmt/test/smtlib/test_parser_qf_lra.py .............Fatal failure within bool cvc5::internal::smt::ProcessAssertions::simplifyAssertions(cvc5::internal::preprocessing::AssertionPipeline&) at /home/pycvc5/src/smt/process_assertions.cpp:442
Internal error detected A bad expression was produced.  Original exception follows:
Error during type checking: Branches of the ITE must have comparable type.
then branch: 8
its type   : Int
else branch: 0.0
its type   : Real

(ite (>= x_92 1) 8 0.0)
The ill-typed expression: (ite (>= x_92 1) 8 0.0)
Fatal Python error: Aborted

Current thread 0x00007691e94acb80 (most recent call first):
  File "pysmt/pysmt/solvers/cvcfive.py", line 149 in solve
  File "pysmt/pysmt/solvers/smtlib.py", line 300 in check_sat
  File "pysmt/pysmt/smtlib/script.py", line 311 in evaluate_command
  File "pysmt/pysmt/smtlib/script.py", line 162 in evaluate
  File "pysmt/pysmt/test/smtlib/parser_utils.py", line 65 in execute_script_fname
```

If I add a print statement in the `test_parser_qf_lra.py` file, the failure does not occur.
Also when I comment out some of the inputs in `small_set/QF_LRA/` from `parser_utils.py` the failure does not occur.

enmag:
Superseded by #840 .

PR #732, opened 2022-07-31
Enforce consistent logic in ackermannization test

### LINKED ISSUE #731: test_ackermannization_for_examples failure
`pysmt/test/test_rewritings.py:test_ackermannization_for_examples` fails non-deterministically.

Kudos @enmag for the [analysis](https://github.com/pysmt/pysmt/pull/730#issuecomment-1200018521):

> The difference seems to be in the order of the filtered_logics, which corresponds to the list of pysmt-supported logics greater than the requested one computed at logic.py:798.
The different orderings are possible since PYSMT_LOGICS is a fronzenset.
Then the min logic is computed by applying min on the list.
However both QF_NIRA < QF_UFLIRA and QF_UFLIRA < QF_NIRA are false.

The test then fails because the available solvers do not support QF_NIRA.

The behavrio of computing supported logic is as expected, but the test is a bit naive.

**Next Steps**:

I wonder if we can improve the test. see whether we can further limit the conditions under which we run the test. Below a copy of the test:
```
def test_ackermannization_for_examples(self):
        for (f, _, _, logic) in get_example_formulae():
            if not logic.is_quantified() and logic.theory.uninterpreted:
                if self.env.factory.has_solvers(logic=logic):
                    self._verify_ackermannization(f)
```
Notice how we are checking that the logic declared in the example is supported. I guess the failure occurs because `logic == QF_UFLIRA`, in this check but then when we call `is_sat` (within `_verify_ackermannization` we resolve the logic to `QF_NIRA`.

marcogario:
Pass logic to _verify_ackermannization so we can check for preconditions there and ensure we use the same logic throughout the test.

Fixes https://github.com/pysmt/pysmt/issues/731

### DIFF
--- a/pysmt/test/test_rewritings.py
+++ b/pysmt/test/test_rewritings.py
@@ -166,7 +166,7 @@ class TestRewritings(TestCase):
         for (f, _, _, logic) in get_example_formulae():
             if not logic.is_quantified() and logic.theory.uninterpreted:
                 if self.env.factory.has_solvers(logic=logic):
-                    self._verify_ackermannization(f)
+                    self._verify_ackermannization(f, logic)
 
     def test_ackermannization_dictionaries(self):
         self.env.enable_infix_notation = True
@@ -192,7 +192,7 @@ class TestRewritings(TestCase):
             if atom.is_function_application():
                 self.assertIsNotNone(terms_to_consts[atom])
 
-    def _verify_ackermannization(self, formula):
+    def _verify_ackermannization(self, formula, logic=None):
         ackermannization = Ackermannizer()
         ack = ackermannization.do_ackermannization(formula)
         #verify that there are no functions in ack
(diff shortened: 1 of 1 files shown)


mikand (approved):
LGTM! :+1:

mikand:
Rebased onto master and simplified the change.

**On #731:** the nondeterminism is already gone. `get_closer_logic` used to pick with `min()` over a frozenset, so `QF_UFLIRA` resolved to either `QF_UFLIRA` or `QF_NIRA` depending on iteration order. Commit 7231657a replaced that with `sorted(res, key=str)[0]`, so the result is now stable.

**What this PR still fixes:** the test checks `has_solvers(logic=logic)` with the example's declared logic, then calls `is_sat(f)` with no logic, so the oracle re-infers one that may differ and may not be supported. Passing `logic` to both `is_sat` calls makes the check and the solve use the same value.

**Difference from the previous version:** I dropped the move of the two skip conditions into `_verify_ackermannization` and kept them in the loop. Reasons:

1. The test above uses `@skipIfNoSolverForLogic`, so an early `return` inside the helper adds a second skip mechanism.
2. `logic=None` would mean "skip the filters" rather than "no logic", and that branch is unreachable for one of the two callers.
3. `_verify_ackermannization` could return without asserting anything and still pass.

Behaviour is unchanged. Spying on `is_sat` gives 52 calls over 26 formulae on both branches, with the same logics:

```
QF_AUFBVLIRA, QF_AUFBVLIRA*, QF_SLIA, QF_UF, QF_UFLIRA, QF_UFLRA
```

`test_rewritings.py` passes (22 tests) with z3, cvc4, cvc5, yices, btor, msat, bdd. Diff is +4/-4, down from +13/-6.

PR #733, opened 2022-07-31
Deterministic get_closer_logic

marcogario:
In https://github.com/pysmt/pysmt/issues/731#issuecomment-1200380943, we discussed how having a non-deterministic answer in `get_logic` can lead to problems.

We do not expect the logic detection in `get_logic` to pick "the best" logic, but we want at least to fail consistently so that users are aware that a logic needs to be specified.

This PR does two things:
1. Adds QF_LIRA to the set of `PYSMT_LOGICS`, this was the trigger for the work, but only adding this logic does not solve the problem.
2. Adds a test for showing that `get_closer_logic` is deterministic independently of the the order of the input argument `supported_logics`.

**Alternatives Considered**
There was another way to fix the same that would have been to make both `PYSMT_LOGICS` and `SMTLIB_LOGICS` an ordered set. I considered and discarded this option because in general we do not need the ordering, and the solution seemed fragile.
Overall, I believe the root cause is `get_closer_logic` not properly accounting for multiple logics being "close" (as we do instead for `most_generic_logic`).

/cc @enmag

### DIFF
--- a/pysmt/logics.py
+++ b/pysmt/logics.py
@@ -806,7 +806,14 @@ def get_closer_logic(supported_logics, logic):
     res = [l for l in supported_logics if logic <= l]
     if len(res) == 0:
         raise NoLogicAvailableError("Logic %s is not supported" % logic)
-    return min(res)
+    # There might be multiple logics that are close.
+    # Instead of throwing an exception (as we do in most_generic_logic),
+    # we enforce a deterministic answer, as we expect this case to be
+    # fairly common.
+    # Note: we could refine this with more heuristics to identify
+    # which logics are preferable, but making this deterministic makes
+    # it easier to spot cases where specifying the logic might be needed.
+    return sorted(res, key=lambda x:str(x))[0]
 
 
 def get_closer_pysmt_logic(target_logic):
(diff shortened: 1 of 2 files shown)


mikand (approved):
LGTM :+1:

pysmt/logics.py
```diff
@@ -798,7 +806,14 @@ def get_closer_logic(supported_logics, logic):
...
         raise NoLogicAvailableError("Logic %s is not supported" % logic)
-    return min(res)
+    # There might be multiple logics that are close.
+    # Instead of throwing an exception (as we do in most_generic_logic),
+    # we enforce a deterministic answer, as we expect this case to be
+    # fairly common.
+    # Note: we could refine this with more heuristics to identify
+    # which logics are preferable, but making this deterministic makes
+    # it easier to spot cases where specifying the logic might be needed.
+    return sorted(res, key=lambda x:str(x))[0]
```
mikand, line 816: I think this is the culprit why the tests are failing: we need to order the logics according to their expressivity and *then* make a deterministic choice among those logics which are incomparable bottoms of the lattice. The problem we have is that the ordering relation among logics is not a total order...

PR #738, opened 2022-10-03
Consistency between the toy problem and code

ashfaq92:
If we look at the code:
```
problem = And(Equals(sum_hello, sum_world),
              Equals(sum_hello, Int(36)))
```
It seems that the code is attempting to solve the problem:  "Is there a value for each letter so that H+E+L+L+O = W+O+R+L+D = 36?"

So, either we can change that code line or the problem line.

Thanks for contributing to pySMT!

Please feel free to delete this text and provide a description for your PR. Here are good things to check:

- [x] For bugs or new features, make sure there is a test showing the bug or demonstrating the use of the new feature
- [x] Are all tests green?
- [x] (First contribution only) You accepted the licensing terms by adding name and email to the CONTRIBUTORS file (see https://github.com/pysmt/pysmt/blob/master/docs/development.rst#licensing).

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -95,7 +95,7 @@ The problem that we are going to solve is the following:
   Lets consider the letters composing the words *HELLO* and *WORLD*,
   with a possible integer value between 1 and 10 to each of them.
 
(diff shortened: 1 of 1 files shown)


mikand:
This has been already fixed in the meantime

PR #754, opened 2023-04-13
Simplified Substituters code

mikand:
Removed redundant code from Substiturer implementations. The removed code was completely redundant with the basic handling functions, and thus removing it should not alter the behavior in any way
 
(Thanks to @Framba-Luca for noticing this)


mikand:
Ops, we were wrong about this: the code actually serves the purpose of avoiding the substitution of variables inside quantifiers.

PR #755, opened 2023-04-29
example/parallel.py: typo fix

tias:
minimal doc change

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -32,6 +32,7 @@ Samuele Gallerani <samuele@gallerani.com>
 Sebastiano Mariani <mariani.sebastiano@gmail.com>
 Suresh Gl <sureshgl009@gmail.com>
 Taylor Hodge <j.taylor.hodge@gmail.com>
+Tias Guns <tias@ulyssis.org>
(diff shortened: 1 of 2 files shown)


mikand:
Thanks @tias !

Can you add your name and emails in the CONTRIBUTORS file to mark the acceptance of the pysmt CLA?

tias:
OK done.

Note that the CLAhub link on 
https://pysmt.readthedocs.io/en/latest/development.html

does not work, and also that I think you moved from nosetest to pytest (lower on that page).

Kind regards,
Tias

mikand:
@tias Thanks for the report! 

It seems that the readthedocs is out-of-sync, I'll look into this ASAP. See https://github.com/pysmt/pysmt/blob/master/docs/development.rst for the updated doc.

Instead, from the CI it seems that you used a different email for your commits (Tias Guns <tias@ulyssis.org>), we are checking that all the contributors emails are in the `CONTRIBUTORS` file. Can you please add the correct email (or an alias in the `.mailmap` file).

mikand:
Thanks @tias .
I know this procedure is a burden, but alt least is a one-shot thing. From now on you can contribute without problems :)

PR #756, opened 2023-06-15
Fix term conversion of important SMT-atoms in AllSMT

masinag:
The conversion of SMT atoms when used as important atoms in AllSMT was broken, because the conversion code only worked for (Boolean) variables, whereas SMT atoms as important symbols [are supported from MathSAT version 5.6.7](https://mathsat.fbk.eu/releasenotes.html). 

I fixed it by using similar code as [the allsat example](https://github.com/pysmt/pysmt/blob/8c79de2635936f980595f4a43ee20a7da7554844/examples/allsat.py#L35).
The private method `MathSAT5Solver._var2term` was not used anymore, so I deleted it.

Also, I refactored `examples/allsat.py` and `examples/allsmt.py` to use the `MathSAT5Solver.all_sat` method.

### DIFF
--- a/examples/allsat.py
+++ b/examples/allsat.py
@@ -1,15 +1,10 @@
 # This example requires MathSAT to be installed.
-#
-# This example shows how to use a solver converter to access
-# functionalities of the solver that are not wrapped by pySMT.
-#
-# Our goal is to call the method msat_all_sat from the MathSAT API.
-#
-import mathsat
-from pysmt.shortcuts import Or, Symbol, Solver, And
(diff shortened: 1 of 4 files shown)


mikand:
Hi @masinag thanks for this PR, looks great!

Can you please add your name and email to CONTRIBUTORS to signal your acceptance of the DCO? (See https://pysmt.readthedocs.io/en/latest/development.html#licensing)

masinag:
Done! Thank you :)

PR #761, opened 2023-07-27
Update of all solvers 

mikand:
Objectives:
- [x] Bumping the versions of all solvers
- [x] Substitute CVC4 with CVC5
- [x] Use the new python api provided by yices


mikand:
@marcogario this should bring pysmt up to date with the current SMT solvers version. I also added (experimentally) support for non-linearity in cvc5, but we have to fix our support of algebraic numbers to make it work properly...

pysmt/cmd/check_version.py
```diff
@@ -41,20 +41,18 @@ def check_version(module):
...
+            import cvc5
+            solver = cvc5.Solver()
+            version = solver.getVersion().decode('ascii')
 
         elif module == "picosat":
             import picosat
             version = picosat.picosat_version()
 
         elif module == "yices":
-            import yicespy
```
marcogario, line 53: Shell we archive this project, now that we do not rely on it anymore?

pysmt/cmd/install.py
```diff
@@ -32,13 +32,11 @@
...
 INSTALLERS = [
-    Installer(MSatInstaller,    "5.6.7", {}),
-    Installer(CVC4Installer,    "1.7-prerelease",
-              {"git_version" : "391ab9df6c3fd9a3771864900c1718534c1e4666"}),
-    Installer(Z3Installer,      "4.8.17", {"osx": "10.16"}),
-    Installer(YicesInstaller,   "2.6.2",
-              {"yicespy_version": "f0768ffeec15ea310f830d10878971c9998454ac"}),
-    Installer(BtorInstaller,    "3.2.1", {}),
+    Installer(MSatInstaller,    "5.6.10", {}),
+    Installer(CVC5Installer,    "1.0.8", {"git_tag" : "cvc5-1.0.8"}),
```
marcogario, line 36: What is the reason for using the `tag` ?

pysmt/cmd/installers/cvc4.py
```diff
@@ -1,92 +0,0 @@
-# Copyright 2014 Andrea Micheli and Marco Gario
```
marcogario, line 1: Do we have to remove this? 
This is a breaking change (that I think is ok) but then I'd probably combine this with a version bump.

pysmt/cmd/installers/msat.py
```diff
@@ -103,10 +103,10 @@ def move(self):
             if f.endswith(".so") or f.endswith(".dll") or f.endswith(".dylib"):
                 SolverInstaller.mv(os.path.join(libdir, f), self.bindings_dir)
 
-        # Fix issue in MathSAT 5.5.1 linking to incorrect directory on OSX
+        # Fix issue in MathSAT 5.6.10 linking to incorrect directory on OSX
```
marcogario, line 106: Minor. If this is still happening, I wonder if we can try to help the team fix their building process.

pysmt/solvers/cvc.py
```diff
@@ -0,0 +1,534 @@
+#
```
marcogario, line 1: The filename is inconsistent, we are calling the solver `cvc5` but the new files do not have the number.
I wonder if just calling the solver CVC would be better at this point.

pysmt/solvers/yices.py
```diff
@@ -261,27 +263,41 @@ def get_value(self, item):
...
             return self.mgr.Bool(bool(res))
         elif ty.is_int_type():
-            res = yicespy.yices_get_integer_value(self.model, titem)
-            return self.mgr.Int(res)
+            mpz = yices_api.mpz_t()
+            errcode = yices_api.yices_get_mpz_value(self.model, titem, mpz)
+            self._check_error(errcode)
+            mpz_term = yices_api.yices_mpz(mpz)
+            s = yices_api.yices_term_to_string(mpz_term, PRINTING_WIDTH, 1, 0)
+            return self.mgr.Int(int(s))
```
marcogario, line 277: (Not introduced here). These are a lot of steps going through the API! Do you have any idea of the performance of `yices_api`?

pysmt/solvers/yices.py
```diff
@@ -306,97 +322,99 @@ def convert(self, formula):
...
     def walk_and(self, formula, args, **kwargs):
-        res = yicespy.yices_and(len(args), args)
+        values = (yices_api.term_t * len(args))(*args)
+        res = yices_api.yices_and(len(args), values)
         self._check_term_result(res)
         return res
 
     def walk_or(self, formula, args, **kwargs):
-        res = yicespy.yices_or(len(args), args)
+        values = (yices_api.term_t * len(args))(*args)
```
marcogario, line 335: I am not sure what is going on here, can you help me understand? Is this allocating a list with the values of `args`?

marcogario (commented):
Thanks for doing this. Quite a few things going on here :sweat_smile: !

Overall looks good. My main concern is that we are breaking the API by "renaming" cvc4 to cvc5.

A few options:
1. We just keep cvc4 around and mark it as deprecated. Main advantage is that people relying on cvc4 keep the existing behavior.
   * It seems you had to introduce a new installer and solver definition for cvc5, so both could co-exist. Given we probably should drop the tests to reduce maintenance.
2. Shall we put an alias that maps cvc4 to cvc5? This avoids breaking integrations "syntactically", but semantically cvc5 and cvc4 are not the same thing. Maybe they are close enough that this is ok?
3. Bump the major version of the library to indicate the break. We just accept that CVC5 is the thing people are most likely interested in and upgrading from cvc4 is easy enough.

I am fine with any of those options.

pysmt/solvers/yices.py
mikand: Exactly, they use ctypes and in this way we create a C-array containing all the args. Maybe we can wrap this in a utility function?

pysmt/solvers/yices.py
mikand: It is quite fast, but I do not have benchmark numbers... Here we basically print an `mpz_t` to avoid finite-precision arithmetic errors (overflows in particular). I could be possible to directly access `mpz` without creating the `yices_mpz` term, but to do so we would need to import and directly use GMP API via ctypes...

pysmt/cmd/installers/msat.py
mikand: I think so, but I have not checked

pysmt/cmd/installers/cvc4.py
mikand: We could keep it if you prefer.

mikand:
@marcogario Thanks a lot for the review!
You are right, we can probably go for option 1 (keep both CVC4 and CVC5). I dislike having the solver named just CVC because the current version of CVC5 is 1.0.8, so in pysmt this would look like cvc version 1.0.8...

I'll restore CVC4 then, but maybe we can exclude it from CI?

pysmt/solvers/yices.py
marcogario: Thanks. I don't think we need to do anything more here, it was just a bit surprising as the syntax is odd.

mikand:
@marcogario I finished backporting CVC4 and I am now fixing the tests. Can I merge when the tests run OK?

marcogario (approved):
Thanks for the update! That's a big change, but is CI is :green_circle: I think we can go ahead! :rocket:

Issue #764, opened 2023-10-12
distutils is removed from Python 3.12 standard library


hauff:
Hello,

distutils is removed from Python 3.12 standard library as you can read here:
https://peps.python.org/pep-0632/

files in pysmt that use distutils:
https://github.com/search?q=repo%3Apysmt%2Fpysmt%20distutils&type=code

mikand:
Thanks for the heads-up @hauff !

@marcogario Maybe before the next release we have to deal with this as it will break pysmt-install in Python 3.12

A.

mikand:
Fixed on master by c3c031d2, merged in #765: there are no `distutils` imports
left under `pysmt/`. `pysmt/cmd/installers/base.py` uses
`setuptools.Distribution` instead, and `pysmt-install` works on Python 3.12 and
3.13 (3.13 is what CI runs).

Closing. See also #822, which reported the resulting `pysmt-install` failure,
and #770, which removed the other Python 3.12 casualty, the `imp` module.

PR #765, opened 2023-10-24
Remove deprecated distutils

marcogario:
* Fix for https://github.com/pysmt/pysmt/issues/764

Mostly followed the migration advice from https://peps.python.org/pep-0632/#migration-advice


mikand:
@marcogario LGTM!
Shall we also add some tests with Python 3.12 (either here or in another PR)?

PR #772, opened 2024-03-14
Fix Z3 back conversion for bitvector operations with more than two arguments

nbailluet:
Prior to this PR, the back conversion did not handle bitvector operations with more than two arguments (but it did for integer operations).
This PR fixes the issue by allowing BVAdd, BVMul, BVAnd, BVOr and BVConcat to be back-converted even when they have more than two arguments.


mikand:
Closing in favor of #833

Issue #778, opened 2024-07-16
floating points are not printed as floating points


MikolasJanota:
From the following code, I'm getting `(* 1.0 (/ 1 4))`.  This makes it difficult for external tools. For instance, vampire refuses to parse ints in place of reals.

```
  from pysmt.shortcuts import Div, Real
  print(Div(Real(1), Real(4)).to_smtlib(daggify=False))
```

mikand:
Hi @MikolasJanota,

thanks for the report. We never noticed this because most SMT parsers work without issues. Nonetheless, I prepared a simple fix in #779. Can you try if this works for you?

PR #779, opened 2024-07-17
Fixed issue #778

mikand:
When printing a real constant in SMTLib, we used a notation `(/ n d)` where `n` is the integer numerator  and `d` the integer denominator. 

This causes issues on more restrictive parsers, so we now print `(/ n.0 d.0)` to force a rational interpretation.

### DIFF
--- a/pysmt/smtlib/printers.py
+++ b/pysmt/smtlib/printers.py
@@ -143,7 +143,7 @@ class SmtPrinter(TreeWalker):
         (n,d) = abs(formula.constant_value().numerator), \
                     formula.constant_value().denominator
         if d != 1:
-            res = template % ( "(/ " + str(n) + " " + str(d) + ")" )
+            res = template % ( "(/ " + str(n) + ".0 " + str(d) + ".0)" )
         else:
             res = template % (str(n) + ".0")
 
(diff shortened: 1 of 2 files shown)


marcogario (approved):
lgtm

mikand:
@MikolasJanota can you confirm this fixes your issue?

MikolasJanota:
> @MikolasJanota can you confirm this fixes your issue?

Hi, this looks correct.  Embarrassingly, I do not know how to easily import packages from github, but  your modification coincides  with what I did in the meantime:

```
  class SmtPrinterFix(pysmt.smtlib.printers.SmtPrinter):
      def walk_real_constant(self, formula: pysmt.formula.FNode):
          c = formula.constant_value()
          template = "(- %s)" if c < 0 else "%s"

          n: int
          d: int
          n, d = abs(c.numerator), c.denominator
          if d != 1:
              res = template % ("(/ " + str(n) + ".0 " + str(d) + ".0)")
          else:
              res = template % (str(n) + ".0")

          self.write(res)
```

mikand:
Good enough :)

FYK you can do `pip install git+https://github.com/pysmt/pysmt.git@issue_778`

PR #781, opened 2024-08-11
Extend pysmt-z3 converter to handle stirngs

enmag:
Z3 supports the theory of [Strings](https://smt-lib.org/theories-UnicodeStrings.shtml) via Sequences. In the PR I extend the pysmt-z3 converter to handle string values, constants and operations.
The changes have been tests via the already existing test_string pytests.


enmag:
The failure on CVC5 looks very much like the one I observed in [PR-730](https://github.com/pysmt/pysmt/actions/runs/9999621404/job/27640945137?pr=730). I am starting to suspect there might be an issue somewhere else or within cvc5 itself.

PR #782, opened 2024-08-11
Type hints

enmag:
In this PR I am adding type-hints to the FNode and FormulaManager.
* I'm adding a basic mypy configuration to check only a few files. As we expand the hints we can include more files and eventually switch to a "opt-out" mechanism and with mypy in strict mode,
* Run `python3 -m mypy` to perform checks, currently raises no errors on the opted-in files.
* in order to avoid circular imports I guard the imports required for type-hints via  `if typing.TYPE_CHECKING`. Then the `from __future__ import annotations` avoids running into errors due to the type-hints referring to undefined symbols.
    see: https://stackoverflow.com/questions/39740632/python-type-hinting-without-cyclic-imports


enmag:
Ouch, I didn't realized that I based this branch on the one I used for #781 . Will fix.

mikand:
Fixed in #813

Issue #784, opened 2024-10-07
Create node does not raise TypeError


annabeks:
When a node with an invalid type is created multiple times, PysmtTypeError is raised only on the first time.

This is because function `FormulaManger::create_node()` type checks the node only when it is created.
If the node is already present in the `FormulaManager.formulae` map, the formula is simply returned without invoking the typechecker.

## Minimal Example
```python
>>> from pysmt.exceptions import PysmtTypeError
>>> from pysmt.shortcuts import Times, Int, Real

>>> def try_building_node():
...     try:
...             a = Times(Int(2), Real(4))
...             print(f"Successfully built {a}")
...     except PysmtTypeError:
...             print("Node is not well-formed, as expected.")
...

>>> try_building_node()
Node is not well-formed, as expected.

>>> try_building_node()
Successfully built (2 * 4.0)
```

## Consequences on smtlib parsing
When parsing a smtlib script, if a TypeError is raised while building an arithmetic operator, function `SmtLibParser::__init__::fix_real` tries interpreting as reals the constant operands that were initially interpreted as integers.
There are two connected problems:

### Un-coerced expression
If the same arithmetic expression is built twice in the smtlib script, the second occurrence does not raise the TypeError and integers are not fixed.
In the following example, while the first occurrence of `(* 1 x)` is fixed into `(* 1.0 x)`, the second one is not fixed and results in a type error when trying to build the LT node.
```python
TEST_SMTLIB=\
"""
(declare-fun x () Real)
(assert (< (* 1 x) 0))
(assert (< 2 (* 1 x)))
"""
p = SmtLibParser()
script = p.get_script(StringIO(TEST_SMTLIB))
```

### Un-raised TypeError
If no integer constant is found, and therefore the arguments are not changed, `SmtLibParser::__init__::fix_real` should avoid retrying to apply again the same operator, and should raise directly a TypeError.
This is implemented by checking https://github.com/pysmt/pysmt/blob/71c35b0b82f7ea63240218777940bed06a18c60c/pysmt/smtlib/parser/parser.py#L353
which is always False (`args` is a tuple and `new_args` is a list).

While this bug should only result in a missed optimization opportunity, in combination with the `create_node` problem, it results in not finding an actual type violation. 
If `op` is actually not well-formed, when invoked again on the same arguments no TypeError is raised.

This already happens (although uncatched) in the current test suite. For example, this test should result in a type error (LT is not a BV operator), which instead is missed (no TypeError is raised).
[...]

PR #785, opened 2024-10-07
Fix Create node does not raise type error

annabeks:
Fixes: https://github.com/pysmt/pysmt/issues/784

This fixes the issue by type-checking every newly created node even when it is found in the `formulae` map.
No overhead should have been added as the type is cached.

This also fixes the tests where a type error in smtlib was not noticed, and adds new ones.

Minor: This also fixes the decision on retrying building a type-invalid node in the smtlib parser. Namely, if there are no integer constants within the arguments, there is nothing to change and an exception is directly raised.


mikand:
@annabeks Thanks a lot! This looks perfect!

PR #790, opened 2025-03-21
Bump CVC5 to latest version

mikand:
- [x] Fixed type-checking of POW in NIRA
- [x] Bumped CVC5 version to latest


mikand:
Already fixed in master

PR #793, opened 2025-03-26
Add optimization tests

Framba-Luca:
This PR aims to improve the coverage of the optimization branch. Still WIP

Please feel free to delete this text and provide a description for your PR. Here are good things to check:

- [x] For bugs or new features, make sure there is a test showing the bug or demonstrating the use of the new feature
- [x] Are all tests green?
- [X] (First contribution only) You accepted the licensing terms by adding name and email to the CONTRIBUTORS file (see https://github.com/pysmt/pysmt/blob/master/docs/development.rst#licensing).


pysmt/optimization/optimsat.py
```diff
@@ -127,54 +128,52 @@ def optimize(self, goal, **kwargs):
...
+        msat_objs = {}
+
         for g in goals:
-            self._assert_msat_goal(g)
+            msat_objs[g] = self._assert_msat_goal(g)
 
         while self.solve():
-            model = self.get_model()
-            yield model, [model.get_value(goal.term()) for goal in goals]
+            if all(self._check_unsat_unbound_infinitesimal(msat_objs[g]) for g in goals):
```
mikand, line 158:
```suggestion
            if all(self._check_unsat_unbound_infinitesimal(v) for v in goals.values()):
```

pysmt/optimization/optimsat.py
```diff
@@ -127,54 +128,52 @@ def optimize(self, goal, **kwargs):
...
+        msat_objs = {}
+
         for g in goals:
-            self._assert_msat_goal(g)
+            msat_objs[g] = self._assert_msat_goal(g)
 
         rt = self.solve()
-        if rt:
+
+        if rt and all(self._check_unsat_unbound_infinitesimal(msat_objs[g]) for g in goals):
```
mikand, line 176: same as above

pysmt/optimization/optimsat.py
```diff
@@ -207,6 +211,36 @@ def get_model(self):
...
         return False
 
+    def _check_unsat_unbound_infinitesimal(self, msat_obj):
+        optres = self._msat_lib.msat_objective_result(self.msat_env(), msat_obj)
+        if optres == self._msat_lib.MSAT_OPT_UNKNOWN:
+            raise SolverReturnedUnknownResultError()
+        elif optres == self._msat_lib.MSAT_OPT_UNSAT:
+            return False
+        else:
+
```
mikand, line 221:
```suggestion
```

pysmt/smtlib/script.py
```diff
@@ -239,14 +239,18 @@ def get_define_fun_parameter_symbols(self):
                 res.add(s)
         return res
 
-    def get_last_formula(self, mgr=None):
+    def get_last_formula(self, mgr=None, return_optimizations=False):
         """Returns the last formula of the execution of the Script.
 
         This coincides with the conjunction of the assertions that are
         left on the assertion stack at the end of the SMTLibScript.
```
mikand, line 246: add docs about the new parameter

pysmt/smtlib/script.py
```diff
@@ -255,14 +259,53 @@ def get_last_formula(self, mgr=None):
...
+                        del max_smt_goals[k]
+                        del max_smt_goals_backtrack[k]
+            elif cmd.name in (
+                smtcmd.MAXIMIZE,
+                smtcmd.MINIMIZE,
+                smtcmd.MAXMIN,
+                smtcmd.MINMAX
+            ):
+                goals.append(_parse_goal(cmd))
+            elif cmd.name == smtcmd.ASSERT_SOFT:
```
mikand, line 298: why not merging this case with the previous one? Not handling the `None` case is not significant from the performances point of view and makes some assumptions on the behavior of the `_parse_goal` function...

pysmt/smtlib/script.py
```diff
@@ -290,6 +333,43 @@ def __str__(self):
         return "\n".join((str(cmd) for cmd in self.commands))
 
 
+def _parse_goal(command, max_smt_goals=None, goal_position=None):
```
mikand, line 336: This is not a real "parsing": maybe `_command_to_goal`?

pysmt/smtlib/script.py
```diff
@@ -290,6 +333,43 @@ def __str__(self):
...
+    assert len(command.args) == 2, f"Command {command.name} must have 2 arguments"
+    singed = False
+    if len(command.args) >= 2:
+        options = command.args[1]
+        if options is not None:
+            for arg in options:
+                if arg[0] == ":signed":
+                    singed = arg[1]
```
mikand, line 344: Just to be sure, raise an error if `signed` is not a bool

pysmt/smtlib/script.py
mikand: Also, even if you pass the extra arguments should not be a problem IMHO

pysmt/smtlib/script.py
```diff
@@ -290,6 +333,43 @@ def __str__(self):
...
+    elif command.name == smtcmd.MAXMIN:
+        return MaxMinGoal(command.args[0], singed)
+    elif command.name == smtcmd.ASSERT_SOFT:
+        assert max_smt_goals is not None
+        assert goal_position is not None
+        formula = command.args[0]
+        cmd_annotations = dict(command.args[1])
+        max_smt_id = cmd_annotations.get(":id", "")
+        max_smt_weight = cmd_annotations.get(":weight", 1)
+        if not isinstance(max_smt_weight, int):
```
mikand, line 361: ??
According to the parser, `max_smt_weight` should be an `FNode`, never a raw number. We have to make sure this is globally true.

pysmt/smtlib/script.py
mikand: Also, looking at the function below, maybe it is simpler to inline it here since you split on all the cases anyway...

pysmt/smtlib/script.py
```diff
@@ -290,6 +333,43 @@ def __str__(self):
...
+            max_smt_weight = max_smt_weight.constant_value()
+        _, goal = max_smt_goals.setdefault(max_smt_id, (goal_position, MaxSMTGoal()))
+        goal.add_soft_clause(formula, max_smt_weight)
+        if len(goal.soft) == 1:
+            return goal
+        # Return None because the max_smt_goal is already in the goal list
```
mikand, line 367: This is quite a strange use of `None`. consider inlining this function, since you separately consider all the cases anyway...

pysmt/solvers/z3.py
```diff
@@ -143,15 +143,15 @@ def __call__(self, solver):
 class Z3Solver(IncrementalTrackingSolver, UnsatCoreSolver,
                SmtLibBasicSolver, SmtLibIgnoreMixin):
 
-    LOGICS = PYSMT_LOGICS - set(x for x in PYSMT_LOGICS if x.theory.strings)
+    LOGICS = PYSMT_LOGICS - set(x for x in PYSMT_LOGICS if x.theory.strings or not x.theory.linear)
```
mikand, line 146: WAIT! in this way, you are disabling non-linear logics even for pure SMT, not only for OMT. You shoudl instead override the `LOGICS` in the optimizers

pysmt/test/omt_examples.py
```diff
@@ -0,0 +1,551 @@
...
+    """
+    def __init__(self, name, assertions, logic, solvable, goals):
+        self._name = name
+        self._assertions = assertions
+        self._logic = logic
+        self._solvable = solvable
+        self._goals = goals
+
+        # consistency checks
+        assert bool(goals) == solvable, "Goals must be specified if and only if the test case is solvable"
```
mikand, line 64: These (including all the ones below this line) should not be assertions, but soething like `PySMTValueError`s

pysmt/test/omt_examples.py
```diff
@@ -0,0 +1,551 @@
...
+
+def get_full_example_omt_formuale(environment=None, fast=True, slow=True):
+    """Return a list of OMTTestCases using the given environment."""
+
+    if environment is None:
+        environment = get_env()
+    assert fast or slow, "If neither fast or slow are True this method returns no test cases"
+
+    with environment:
+        omt_examples = []
```
mikand, line 125: why not construct an iterator instead of building the list?

pysmt/test/omt_examples.py
```diff
@@ -0,0 +1,551 @@
...
+def generate_examples_with_solvers(optimization_examples):
+    """
+    This method takes a list of OMTTestCases and yields all the possible
+    combinations of an OMTTestCase and the name of a solver that support
+    the logic of said test.
+    """
+    for optimization_example in optimization_examples:
+        for solver_name in get_env().factory.all_optimizers(logic=optimization_example.logic):
+            # is_sua_or_incr is passed to the methods to skip the unbound or infinitesimal cases
+            is_sua_or_incr = "_sua" in solver_name or "_incr" in solver_name
```
mikand, line 143: Why not checking the `can_diverge_for_infinitesimal` method (that should be static)?

pysmt/test/omt_examples.py
```diff
@@ -0,0 +1,551 @@
...
+
+
+def solve_given_example(pytest_test_case, optimization_example, solver_name, test_to_skip=None):
+    """
+    Method to solve a single OMTTestCase using the given solver.
+    """
+    with warnings.catch_warnings():
+
+        # ignore the z3 defining a new symbol warning, since it is raised many times
+        warnings.filterwarnings("ignore", r"Defining new symbol: z3name!\d+")
```
mikand, line 157: avoid these warning silencing, we leave the in all other tests

pysmt/exceptions.py
```diff
@@ -163,12 +163,23 @@ class PysmtUnboundedOptimizationError(PysmtException):
...
-        return self.goal
\ No newline at end of file
+        return self._goal
+
+    @property
+    def mode(self):
+        return self._mode
+
+    def __str__(self):
+        mode_str = f" in mode '{self.mode}'" if self.mode else ""
```
mikand, line 184: Pysmt is still compatible with Python 3.5, we disallow f-strings. Please remove them from all your code (check this with `grep -RiI 'f"' pysmt/`)

pysmt/test/omt_examples.py
```diff
@@ -0,0 +1,551 @@
...
+        # ignore the z3 warning on boxed optimization being suboptimal
+        warnings.filterwarnings("ignore", r"Boxed optimization is not working in Z3 .*")
+
+        # ignore the MathSAT warning on UF with bool arguments
+        warnings.filterwarnings("ignore", r"MathSAT convert\(\): UF with bool arguments have been .*")
+
+        # test basic in boxed only if there is no basic optimization explicitly specified
+        test_basic_in_boxed = all(ot != OptimizationTypes.BASIC for _, ot in optimization_example.goals.keys())
+
+        is_sua_or_incr = "_sua" in solver_name or "_incr" in solver_name
```
mikand, line 168: same as above, in addition never rely on naming conventions on the solver names

pysmt/optimization/optimsat.py
```diff
@@ -207,6 +211,36 @@ def get_model(self):
...
+            if unbounded > 0:
+                raise PysmtUnboundedOptimizationError("The optimal value is unbounded")
+
+            is_strict = self._msat_lib.msat_objective_value_is_strict(
+                self.msat_env(),
+                msat_obj,
+                self._msat_lib.MSAT_OPTIMUM
+            )
+            if is_strict:
+                raise PysmtUnboundedOptimizationError("The optimal value is infinitesimal")
```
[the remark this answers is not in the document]
mikand: And why this is not found by any test?

pysmt/test/omt_examples.py
```diff
@@ -0,0 +1,551 @@
...
+            assert len(goals) == len(costs) == len(goals_values), test_id_str
+
+            for goal, goal_value, cost in zip(goals, goals_values, costs):
+                _check_oracle_goal(pytest_test_case, goal, goal_value, cost, test_id_str)
+    elif not is_sua_or_incr:
+        with pytest_test_case.assertRaises(raised_class, msg=test_id_str):
+            optimizer.pareto_optimize(goals)
+
+
+def _test_boxed(pytest_test_case, optimizer, goals, goals_values, test_id_str, test_basic, is_sua_or_incr):
```
mikand, line 261: In general, I do not like that so much business logic of the tests is in this file (same for the other kinds of optimization). Why not put this in the `test_xxx` file(s)?

pysmt/test/smtlib/omt/smtlib2_allsat.smt2
```diff
@@ -0,0 +1,86 @@
+; -*- SMT2 -*-
```
mikand, line 1: Fix this intro: there is no allsat anymore...

pysmt/test/smtlib/omt/smtlib2_boxed_int.smt2
```diff
@@ -0,0 +1,66 @@
+; -*- SMT2 -*-
+;
+; Author: Patrick Trentin <patrick.trentin@unitn.it>
+;
+; This file is part of OptiMathSAT.
```
mikand, line 5: Same for all the other files: say that we took these files from the optimathsat distribution with the permission of the authors.

mikand: And maybe remove the rest of the intro comments

pysmt/test/smtlib/parser_utils.py
```diff
@@ -194,3 +202,119 @@ def formulas_from_smtlib_test_set(logics=None):
...
+    If logics is None, all the test cases will be returned.
+
+    Note: This resets the Environment at each call.
+    """
+    for (logic, fname, is_sat, expected_result) in OMTLIB_TEST_FILES:
+        if logics is not None and logic not in logics:
+            continue
+        reset_env()
+        smtfile = os.path.join(OMTLIB_DIR, fname)
+        parser = SmtLib20Parser()
```
mikand, line 221: Why insisting on the 2.0 version of the parser? The basic one is more general...

pysmt/test/smtlib/test_omt_lib_solver.py
```diff
@@ -329,5 +334,29 @@ def test_omt_lex_unsat(self):
         )
 
 
+# define a dummy to have access to the assertEqual etc. methods
+# inside the parametrized tests
+dummy = TestCase()
```
mikand, line 339: this is a bit weird. Is there no other solution?

pysmt/test/test_optimization.py
```diff
@@ -15,13 +15,15 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 #
```
mikand, line 17: Why adding imports if the file is unchanged? In general, try to see with VSCode if we can remove unused imports...

mikand (changes requested):
Overall, it's a good piece of work. I left some comments for improving the code.

pysmt/optimization/optimsat.py
Framba-Luca: `goals` is not a `dict`, did you mean `msat_objs.values()`?

mikand: Yes, sorry :smile:

pysmt/optimization/optimizer.py
```diff
@@ -288,12 +302,14 @@ def __init__(self, goal, environment):
...
     def get_costraint_strict(self):
         if self.val is not None:
+            assert self.val.is_constant()
             return self.op_strict(self.goal.term(), self.val)
         else:
             return None
 
     def get_costraint_ns(self):
         if self.val is not None:
+            assert self.val.is_constant(), self.val
```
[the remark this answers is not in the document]
Framba-Luca: I actually noticed that the `get_constraint_strict` and `get_constraint_ns` could be refactored in a single method `get_constraint(strict: bool)`, which is cleaner and easier to understand. 

Let me know if this is OK

pysmt/test/omt_examples.py
Framba-Luca: Ok, I put those as assert because I was not expecting the user to use this class, but only test-developers.

Setting them as exceptions anyway

pysmt/test/omt_examples.py
Framba-Luca: Yes, good idea!

Just 2 notes:

1.  the `can_diverge_for_unbounded_cases` method is not static
2. here we have just the string `solver_name`, so the Optimizer class has to be instantiated.

The cleanest solution I see is:
- make the method static
- make a factory method that returns the class of an optimizer from the name (should I also do it for solvers, just for completeness?)
- use the static method here

Let me know what you think about this

pysmt/test/smtlib/test_omt_lib_solver.py
Framba-Luca: I did a bit more research and didn't find a better option.

This is to be able to use the `assertEqual(x1, x2)` etc. methods, which are better than `assert x1 == x2` and provide a clearer error message.

The dummy could also be created inside the single parametrized test that uses it.

Let me know which of the 3 options you like more:
1) leave as is
2) move the dummy creation inside the parametrized test
3) use the simple python `assert` instead of the `unittest` methods

pysmt/optimization/optimizer.py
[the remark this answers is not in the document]
mikand: OK!

pysmt/test/smtlib/test_omt_lib_solver.py
mikand: It seems that solution 3 is the one suggested on the official documentation and is the one we also use for the other smtlib tests, so let's go with that

pysmt/test/omt_examples.py
mikand: I like the solution, but there is no need to touch the factory: the `all_optimizers` method already returns a dict frm solver name to optimizer class :smile:

pysmt/test/omt_examples.py
mikand: Yes, but these are not part of the examples, they are more testing utils. Maybe we can put them in a separate file that is imported by the testing files? Something like `optimization_utils.py`?

mikand (changes requested):
Excellent, I left just a couple of comments to consider before merging this!

PR #797, opened 2025-04-08
Added Abs shortcut as shorthand for Ite(x > 0, x, -x)

aemartinez:
I added `Abs` as a shortcut for Ite(x > 0, x, -x). 

I haven't yet added support for parsing the `abs` SMT-LIB function and simplifying it into Ite(x > 0, x, -x). Let me know if you think this would be a useful addition.

- [ x ] For bugs or new features, make sure there is a test showing the bug or demonstrating the use of the new feature
- [ x ] Are all tests green?
- [ x ] (First contribution only) You accepted the licensing terms by adding name and email to the CONTRIBUTORS file (see https://github.com/pysmt/pysmt/blob/master/docs/development.rst#licensing).


mikand:
Thanks @aemartinez!

This looks great!

PR #799, opened 2025-05-08
Fix compatibility with Z3 4.14.1

McSinyx:
Z3 4.14.1 introduces support for SMT-LIB 2.7's operators `ubv_to_int`, `sbv_to_int` and `int_to_bv`.  pySMT does not seem to support the latter to operation out of the box, so we alias `ubv_to_int` to `bv2int`/`bv2nat` here to at least get `pysmt/test/test_back.py::TestBasic::test_z3_back_formulae` to pass.

References: https://github.com/Z3Prover/z3/issues/7572
References: https://issues.guix.gnu.org/78007

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -29,6 +29,7 @@ M. Fareed Arif <fareed.arif@yahoo.com>
 Marco Gario <marco.gario@gmail.com>
 Mathias Preiner <mathias.preiner@gmail.com>
 Matthew Venn  <matt@mattvenn.net>
+Nguyễn Gia Phong <cnx@loang.net>
 Nicolas Bailluet <nicolas.bailluet@inria.fr>
 Nitish  <nkd.2195@gmail.com>
 Quantik <40385569+quantik-git@users.noreply.github.com>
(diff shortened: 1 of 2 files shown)


mikand:
Thanks for the contribution @McSinyx.

Before merging, can you please add an example smt2 file in the `test/smtlib` directory and test it analogously to `test_parser_qf_ufbv.py`?

McSinyx:
It seems that `ubv_to_int` is specified under [`QF_BV`](https://smt-lib.org/logics-all.shtml#QF_BV).  Why aren't `smtlib_tests(lambda x: x == QF_BV)` defined/run?  I was thinking about putting a `conversion.smt2.bz2` into `test/smtlib/small_set/QF_BV`.

mikand:
@McSinyx You are right, we are missing something there. Nevermind, we re-discovered the issue in #808 while updating Z3 and copied your fix there. @Framba-Luca will add a test there and, if it is OK for you, we will add your name to the contributors there.

McSinyx:
> tests are already failing if we upgrade Z3

Yes, that was the purpose of this PR.

> if it is OK for you, we will add your name to the contributors there.

That would be lovely!

Just curious, should GH-808 cover `sbv_to_int` and `int_to_bv`?

PR #803, opened 2025-05-13
Fix issue 801

### LINKED ISSUE #801: `CNFizer` raise an exception because of an undefined function (`function_symbol`)
Precisely here in `walk_function`, it tries to call `function_symbol` instead of `function_name`:
https://github.com/pysmt/pysmt/blob/b8ecdc8034919e3971b96691add14c0b6f2b6596/pysmt/rewritings.py#L165
I guess it was the old name of `function_name` and was not changed.

Framba-Luca:
This PR fixes the `CNFizer` `walk_function` method, renaming `function_symbol` (deprecated name) with `function_name`

fixes #801


mikand:
@Framba-Luca Can you add a no-regressio test in `test_regression.py`?

PR #813, opened 2025-05-27
Add mypy checks in CI

Framba-Luca:
This PR aims to add mypy checks in CI and bring the typing to a working state


mikand (changes requested):
I reviewed some of the files, in general please check the occurences of `Any`

PR #840, opened 2026-09-01
reorder operands of commutative operators based on node-id

enmag:
* sort arguments of commutative operations based on FNode id in FormulaManager.
* update walks in solvers: btor, cvc5, cvc4, mathsat, yices, z3.
* remove unused imports
* fix formatting

Supersedes [#730](https://github.com/pysmt/pysmt/pull/730).

### DIFF
--- a/pysmt/formula.py
+++ b/pysmt/formula.py
@@ -25,12 +25,13 @@ defined (operators.py). This is because many operators are rewritten,
 and therefore are only virtual. Common examples are GE, GT that are
 rewritten as LE and LT. Similarly, the operator Xor is rewritten using
 its definition.
+In addition,  operators of n-ary commutative operators are sorted.
 """
 
 import sys
 import fractions
 import warnings
-from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Union, cast, Iterable
+from typing import Any, Callable, Dict, Optional, Sequence, Tuple, Union, cast, Iterable
 
 if sys.version_info >= (3, 3):
     from collections.abc import Iterable as CollectionsIterable
@@ -42,11 +43,10 @@ import pysmt
 
 import pysmt.typing as types
 import pysmt.operators as op
-from pysmt.fnode import FNode
 from pysmt.typing import PySMTType
 from pysmt.fnode import FNode, FNodeContent
 from pysmt.exceptions import (UndefinedSymbolError, PysmtValueError,
-                             PysmtTypeError, PysmtEmptySymbolNameError)
+                              PysmtTypeError, PysmtEmptySymbolNameError)
 from pysmt.walkers.identitydag import IdentityDagWalker
 from pysmt.constants import Fraction, Numeral
 from pysmt.constants import (is_pysmt_fraction,
@@ -165,7 +165,7 @@ class FormulaManager(object):
          - Formula must be of boolean type
          - Variables must be BOOL, REAL or INT
         """
-        variables_tuple = tuple(variables)
+        variables_tuple = self._sorted_tuple(variables)
         if len(variables_tuple) == 0:
             return formula
         return self.create_node(node_type=op.FORALL,
@@ -180,7 +180,7 @@ class FormulaManager(object):
          - Formula must be of boolean type
          - Variables must be BOOL, REAL or INT
         """
-        variables_tuple = tuple(variables)
+        variables_tuple = self._sorted_tupletuple(variables)
         if len(variables_tuple) == 0:
             return formula
         return self.create_node(node_type=op.EXISTS,
@@ -225,7 +225,8 @@ class FormulaManager(object):
 
         Restriction: Left and Right must be of boolean type
         """
-        return self.create_node(node_type=op.IFF, args=(left, right))
+        return self.create_node(node_type=op.IFF,
+                                args=self._args_to_sorted_tuple(left, right))
 
     def Minus(self, left: FNode, right: FNode) -> FNode:
         """ Creates an expression of the form:
@@ -246,15 +247,14 @@ class FormulaManager(object):
          - Arguments must be all of the same type
          - Arguments must be INT or REAL
         """
-        tuple_args = self._polymorph_args_to_tuple(args)
+        tuple_args = self._args_to_sorted_tuple(args)
         if len(tuple_args) == 0:
             raise PysmtTypeError("Cannot create a Times without arguments.")
 
         if len(tuple_args) == 1:
             return tuple_args[0]
         else:
-            return self.create_node(node_type=op.TIMES,
-                                    args=tuple_args)
+            return self.create_node(node_type=op.TIMES, args=tuple_args)
 
     def Pow(self, base: FNode, exponent: FNode) -> FNode:
         """ Creates the n-th power of the base.
@@ -292,7 +292,7 @@ class FormulaManager(object):
         For the boolean case use Iff
         """
         return self.create_node(node_type=op.EQUALS,
-                                args=(left, right))
+                                args=self._args_to_sorted_tuple(left, right))
 
     def NotEquals(self, left: FNode, right: FNode):
         """ Creates an expression of the form: left != right"""
@@ -361,7 +361,7 @@ class FormulaManager(object):
         elif is_python_rational(value):
             val = pysmt_fraction_from_rational(value)
         else:
-            raise PysmtTypeError("Invalid type in constant. The type was:" + \
+            raise PysmtTypeError("Invalid type in constant. The type was:" +
                                  str(type(value)))
 
         n = self.create_node(node_type=op.REAL_CONSTANT,
@@ -380,7 +380,7 @@ class FormulaManager(object):
         elif is_python_integer(value):
             val = pysmt_integer_from_integer(value)
         else:
-            raise PysmtTypeError("Invalid type in constant. The type was:" + \
+            raise PysmtTypeError("Invalid type in constant. The type was:" +
                                  str(type(value)))
         n = self.create_node(node_type=op.INT_CONSTANT,
                              args=tuple(),
@@ -400,7 +400,7 @@ class FormulaManager(object):
             self.string_constants[value] = n
             return n
         else:
-            raise TypeError("Invalid type in constant. The type was:" + \
+            raise TypeError("Invalid type in constant. The type was:" +
                             str(type(value)))
 
     def TRUE(self) -> FNode:
@@ -412,13 +412,10 @@ class FormulaManager(object):
         return self.false_formula
 
     def Bool(self, value: bool) -> FNode:
-        if type(value) != bool:
+        if not isinstance(value, bool):
             raise PysmtTypeError("Expecting bool, got %s" % type(value))
 
-        if value:
-            return self.true_formula
-        else:
-            return self.false_formula
+        return self.true_formula if value else self.false_formula
 
     def And(self, *args: Union[FNode, Iterable[FNode]]) -> FNode:
         """ Returns a conjunction of terms.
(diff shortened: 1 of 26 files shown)


enmag:
@mikand can you have a look when you have the time and let me know what you think?

pysmt/solvers/yices.py
```diff
@@ -512,20 +518,21 @@ def walk_bv_ule(self, formula: FNode, args: List[int], **kwargs) -> int:
...
+                                        formula.bv_extract_start(),
+                                        formula.bv_extract_end())
         self._check_term_result(res)
         return res
 
     def walk_bv_or(self, formula, args, **kwargs):
-        res = yices_api.yices_bvor2(args[0], args[1])
-        self._check_term_result(res)
+        values = (yices_api.term_t * len(args))(*args)
+        res = yices_api.yices_bvor(len(args), values)
```
mikand, line 535: you accidentally removed `self._check_term_result(res)` (here and in the following ones)

pysmt/solvers/z3.py
```diff
@@ -900,6 +902,15 @@ def walk_binary(self, formula, args, **kwargs):
             return z3term
         return walk_binary
 
+    def make_walk_nary_to_binary(func):
+        def walk_n_to_b(self, formula, args, **kwargs):
+            z3term = args[0]
+            for arg in args[1:]:
+                z3term = func(self.ctx.ref(), z3term, arg)
+                z3.Z3_inc_ref(self.ctx.ref(), z3term)
```
mikand: here you might leak some Z3 terms (the dec_ref is done upon memoization clearing, but you never add intermediate nodes into memoization). Something like:
```python
      def make_walk_nary_to_binary(func):
          def walk_n_to_b(self, formula, args, **kwargs):
              z3term = args[0]
              for i, arg in enumerate(args[1:]):
                  res = func(self.ctx.ref(), z3term, arg)
                  z3.Z3_inc_ref(self.ctx.ref(), res)
                  if i:
                      # z3term is our own intermediate, not a memoized child
                      z3.Z3_dec_ref(self.ctx.ref(), z3term)
                  z3term = res
              return z3term
          return walk_n_to_b
```

pysmt/test/examples.py
```diff
@@ -174,7 +174,7 @@ def get_full_example_formulae(environment: Optional[Environment]=None) -> List[E
                     logic=pysmt.logics.QF_IDL
                 ),
 
-            Example(hr="((x ? 7 : ((p + -1) * 3)) = q)",
+            Example(hr="(q = (x ? 7 : ((p + -1) * 3)))",
```
mikand, line 177: This is the part I am most unsure about: reordering children nodes using their ID, increases memoization hits, but makes the printing not easy to predict: something like `And(Symbol(a), Symbol(b))` would render as `a & b` or `b & a` depending on the previous history of the formula manager. 

I am not sure this user-level issue is justified by the memory saving. Any idea on how to make the order less brittle (here for example, by simply reordering the examples, you might get different serializations)...

mikand (commented):
Thanks @enmag! The code is mostly OK (see my comments).
But the key question is: do you have some example/data on how much memory we can save with this approach?

Because I think this reordering is surely semantically correct, but it can be quite annoying for the users (e.g. formulae printed then parsed are not syntactically equivalent anymore even in simple cases). Maybe it is worth to make it optional? Or to use something other than node ID (see my inline comment)

pysmt/formula.py
```diff
@@ -733,55 +731,62 @@ def BVNot(self, formula: FNode) -> FNode:
                                 payload=(formula.bv_width(),))
 
     def BVAnd(self, *args: Union[FNode, Sequence[FNode]]) -> FNode:
-        """Returns the Bit-wise AND of bitvectors of the same size.
-        If more than 2 arguments are passed, a left-associative formula is generated."""
-        args = self._polymorph_args_to_tuple(args)
+        """Returns the Bit-wise AND of bitvectors of the same size."""
+        args = self._args_to_sorted_tuple(args)
```
mikand, line 735: This is OK, but note that simplifier assumes the internal representation is binary, so we need to fix that as well (also for other BV operators)

pysmt/formula.py
```diff
@@ -733,55 +731,62 @@ def BVNot(self, formula: FNode) -> FNode:
...
-        for arg in args[1:]:
-            res = self.create_node(node_type=op.BV_OR,
-                             args=(res,arg),
-                             payload=(res.bv_width(),))
-        return res
+
+        if len(args) == 1:
+            return args[0]
+
+        assert all(v.bv_width() == args[0].bv_width() for v in args)
```
mikand, line 757: Is this needed? Type checker should catch this

pysmt/formula.py
```diff
@@ -817,17 +818,19 @@ def BVNeg(self, formula: FNode) -> FNode:
...
-        for arg in args[1:]:
-            res = self.create_node(node_type=op.BV_ADD,
-                             args=(res,arg),
-                             payload=(res.bv_width(),))
-        return res
+
+        if len(args) == 1:
+            return args[0]
+
+        assert all(v.bv_width() == args[0].bv_width() for v in args)
```
mikand, line 830: Same as above

pysmt/formula.py
```diff
@@ -836,17 +839,19 @@ def BVSub(self, left: FNode, right: FNode) -> FNode:
...
-        for arg in args[1:]:
-            res = self.create_node(node_type=op.BV_MUL,
-                             args=(res,arg),
-                             payload=(res.bv_width(),))
-        return res
+
+        if len(args) == 1:
+            return args[0]
+
+        assert all(v.bv_width() == args[0].bv_width() for v in args)
```
mikand, line 851: Same as above

pysmt/formula.py
mikand: Maybe we shall add some n-ary BV examples: for example with `a=0xF0`, `b=0xFF`, `c=0x0F`, `BVAnd(a,b,c).simplify()` returns `0xF0` instead of `0x00` because the simplifier silently assumes BVAnd to be a binary operator.

pysmt/solvers/z3.py
enmag: Oh, thanks!    
We should make sure we increase the reference count of the root node, which is also the one that is going to get memoized in pysmt. Correct?
If this is correct, why not just call `z3.Z3_inc_ref(self.ctx.ref(), z3term)` right before the return statement?
```python
    def make_walk_nary_to_binary(func):
        def walk_n_to_b(self, formula, args, **kwargs):
            z3term = args[0]
            for arg in args[1:]:
                z3term = func(self.ctx.ref(), z3term, arg)
            z3.Z3_inc_ref(self.ctx.ref(), z3term)
            return z3term
        return walk_n_to_b
```

pysmt/test/examples.py
enmag: Yes, it will break flows that rely on having control of the syntactic structure of formulae.

With respect to stable ordering. I suspect a stable ordering would require the sort method to look into the structure of the formula. This is likely to introduce more overhead.
One could store and sort based on hashes. However, hashes in some languages are not stable, right now I don't know what the python standard says.

In my original thing (#601) I was using a tuple `(x.is_constant(), x.node_type(), x.node_id())` as key on which to sort. This made the ordering a bit more stable for my purposes (I was basically processing collections of NIRA expressions). 
I don't know how much the tuple improves with respect to just the node id in the general setting. Consider a disjunction of conjuncts, the order of the conjuncts is still gonna depend on the node-id.

Back then I built a custom walker to perform the normalizations that were necessary for my use case. See [#601-comment](https://github.com/pysmt/pysmt/issues/601#issuecomment-766152654).
I was mostly concerned about runtime, memory was not an issue in my benchmarks.
My goal was to maximize node sharing to have a more reliable `==` check and prevent having to do a validity check on the `iff`.
The use-case might be too specific, and folks that need something like this can probably define their own walker to perform the transformation (as I did).

I guess the decision here is whether pysmt should try and maximize reuse of formulae or should push that concern to the underlying solver and instead provide users with more predictable AST / syntactic manipulation.

About putting the normalization in a separate walker. The advantage is that it becomes an extra functionality that one calls if needed, but I don't know how many folks actually need this and if it's worth the maintenance burden. In addition, compared to the FormulaMgr-normalization this would cost an extra pass on the AST and to free-up the memory one may need to move the normalized nodes to another environment and free the old env (is there a better way to do this?).

Let me know what your take is. Happy to cancel the PR if it is not of general interest.

In the original (old) PR I had some plots about impact on runtime. After we have a decision on the above, I can make some time to re-build the experimental setup and get some plots around runtime and memory. Is there something already available I can re-use?

mikand:
Thanks for the detailed writeup, and sorry for the slow turnaround.

On the "maximize reuse vs. predictable ASTs" question: I don't think we have to choose. Let's make the reordering opt-in via an `Environment` flag, defaulting to off. It fits the existing pattern (`allow_empty_var_names`) and, since you already routed every sort through `_sorted_tuple`, it's a two-line change there plus one line in `Environment.__init__`:

```python
def _sorted_tuple(self, nodes):
    result = tuple(nodes)
    if not all(isinstance(r, FNode) for r in result):
        raise PysmtTypeError("Typing not respected")
    if not self.env.sort_commutative_args:
        return result
    return tuple(sorted(result, key=FNode.node_id))
```

Every call site you already added (`And`, `Or`, `Plus`, `Times`, `Equals`, `Iff`, the BV ops, `ForAll`/`Exists`, `AtMostOne`, `AllDifferent`, `_Min/_MaxWrap`) is then gated for free. With the default off, the `examples.py` and `test_printing.py` serialization changes go away, nobody's existing flow breaks, and users who want maximal sharing (your `iff`-validity use case) get it with `env.sort_commutative_args = True`. Users with a really specific normal form can still write their own walker, as you did in #601.

Two notes on that:

1. The flag makes the behaviour *optional*, not *stable*. `node_id` is a monotonic counter, so the order is deterministic for a given program, but it still depends on construction history. I looked at whether a structural key would be better and I don't think it's worth an extra pass over the AST, nor is `(is_constant, node_type, node_id)` enough to make it predictable. Documenting "the argument order is an implementation detail when the flag is on" is the honest framing.
2. The flag has to be set before any formula is built, otherwise you can end up with both `And(a,b)` and `And(b,a)` in the same manager, which defeats the purpose and silently breaks the `==` check that motivates enabling it in the first place. Either document that clearly or have the setter raise once the manager is non-empty (`len(self.formula_manager.formulae) > 2`, since a fresh manager only holds TRUE/FALSE). I'd take the guard, it's cheap.

No need to split the PR, let's do both things here. A few things I'd still like in before merging:

- **The simplifier.** It still assumes binary BV operators, so right now `BVAnd(BV(0xF0, 8), BV(0xFF, 8), BV(0x0F, 8)).simplify()` gives `0xF0` instead of `0x00`. Some n-ary BV examples in `examples.py` would cover this.
- **The z3 ref-count leak** and the `_check_term_result` calls that got dropped in `yices.py`, as per the inline comments.
- **Drop the redundant width asserts** in `BVOr`/`BVAdd`/`BVMul` (`assert all(v.bv_width() == ...)`), the type checker already catches those.
- **A test enabling the flag**, checking `And(a,b) is And(b,a)` and idempotence, plus a `CHANGES.rst` entry covering both the n-ary BV operators and the new flag.

On benchmarks: with the flag off by default I don't think we need new plots to merge. The 2-3% from #730 is enough context and it now only applies to people who opt in. If you do want to re-run something, the interesting number is memory/node count with the flag on, not runtime.

enmag:
ack. I think I will push another commit with the changes next week.

About the z3 ref count. Any reason why a single increment before returning would not work?

PR #843, opened 2026-09-02
Noughts and crosses example (rebase of #565)

mikand:
Rebases @mattvenn's #565 onto current master and finishes it off. The original
13 commits are kept as they were, with three new commits on top.

Closes #565.

### Why it needed work

The example did not run on master. On the solvers I have installed:

| solver | before |
| --- | --- |
| cvc4 / cvc5 | abort on the first `solve()`, `LogicException: the logic was specified as QF_UFLIRA, which doesn't include THEORY_BV` |
| optimathsat (older builds) | segfault in `msat_term_arity` via `get_value` |
| btor | runs, but almost every game ends in `BoolectorException: cannot retrieve model if input formula is not SAT` |

### `Fix the xoxo example so it runs on any QF_BV solver`

* `Solver()` was picking pySMT's default logic. Ask for `QF_BV`, which is
  what the review asked for back in 2019.
* `find_all_moves()` / `pick_new_move()` inspected `solver.assertions`, which
  only `IncrementalTrackingSolver` provides, so they raised `AttributeError`
  on cvc4/cvc5. They now use a python-side record of the position, which also
  stops the move lists from offering cells the opponent occupies.
* `print_board()` read a model without solving first. Since every added
  assertion invalidates the previous model, the "o wins" and "o can't play"
  endings died with a solver error instead of showing the final board.
* `load_board()` did not count the counters it loaded, so the board-sum
  constraint was immediately unsatisfiable for any non-empty position and
  `tests/test1.txt` could not be used at all. The position is now selectable
  with `--board` and resolved relative to the script rather than the cwd.
* The draw path exited without printing the final board.
* `exit(0)` -> `break`, unused `Plus`/`Int`/`argparse` imports dropped, logger
  defined at module level, and a `--moves` option to play a scripted game.

### `Explain the xoxo example and tidy it up`

The main review comment left open on #565 was what exactly is being asked of
the solver, and how a model maps back to a move. `README.rst` (renamed from
`.md` to match `examples/sudoku/`) and a module docstring now walk through the
encoding, why the board-sum constraint forces the free cells to be blank, the
four queries made each turn, and how the two-moves-ahead blocking query works.

Also: `Cell` members hold plain ints with a `.bv` property, so `x_val`/`o_val`
and the `constant_value()` round-trip are gone; `get_win_formula()` builds the
eight lines from a table instead of spelling out eight 105-column
conjunctions; `get_board_sum()` reduces with `BVAdd` instead of relying on
infix notation; `turns_played_formula(extra_x=, extra_o=)` makes each query
read as "the position, plus one more nought". `boards/` (was `tests/`, which
suggested a test suite) holds a legal mid-game position instead of one where o
had already won.

### `Run the xoxo example in CI`

The examples loop only picks up `examples/*.py`, so nothing under
`examples/xoxo/` was ever exercised. It now plays a scripted game and checks
that a result is announced. Costs about 0.15s.

### Testing

31 scripted games against each of optimathsat, btor, cvc5 and cvc4 (124 runs):
no failures, and all of x-wins / o-wins / draw exercised on every solver. Plus
the mid-game board, piped stdin, and `--verbose` on each.


mikand:
@mattvenn If it's OK for you (I know this is very old stuff), can we add your name and email to CONTRIBUTORS to signal you accept PySMT's DCO (see https://pysmt.readthedocs.io/en/latest/development.html#licensing)?

PR #844, opened 2026-09-03
Nonlinear arithmetic support for MathSAT and OptiMathSAT

### LINKED ISSUE #318: Extend CVC4 support for Non Linear Real Arithmetic

mikand:
Lets the MathSAT wrapper handle NRA/NIA. OptiMathSAT inherits `MathSAT5Solver.LOGICS`, so it gains nonlinear solving and nonlinear optimization for free.

### What it does

- Drops the linearity filter from `MathSAT5Solver.LOGICS` and the `NonLinearError` guard in `walk_times`.
- Adds `walk_pow` on top of `msat_make_pow`. Exponents MathSAT cannot handle come back as an error term, which `convert()` already turns into an `InternalSolverError`.
- Adds `walk_div`. `msat_make_divide` is real division, so integer division is built as `den >= 0 ? floor(num/den) : -floor(num/-den)`, matching the SMT-LIB semantics pySMT and cvc5 use. Verified to agree with cvc5 on both signs.
- Maps `MSAT_TAG_DIVIDE` in the back-conversion and signature tables.
- Two NRA examples had irrational roots (`r*r = 2.0`, `r*r*r = 25.0`) on which MathSAT does not terminate. They now use `4.0` and `27.0`. Algebraic values stay covered by `test_nlira.test_irrational`.
- `test_nlira` runs the nonlinear tests against every solver advertising QF_NRA/QF_NIA instead of z3 only; `test_examples_msat` no longer skips nonlinear logics.
- README table records NRA/NIA for `msat`, and for `cvc5`, which already supported them but was not listed.

### Why this supersedes #533

#533 (open since 2018) proposed the same thing for MathSAT **and CVC4**. Since then:

- The CVC half already landed by another route. `cvc4.py` is now `cvcfour.py`, and the new `cvcfive.py` already has an unfiltered `LOGICS`, an unrestricted `walk_times`, `walk_pow` and `walk_div`. #533's CVC4 code targets the dead SWIG API (`CVC4.POW`, `CVC4.DIVISION`) and CVC4 itself is EOL, so there is nothing worth porting there.
- Only the MathSAT half is still missing from master, and #533 no longer applies: five of its seven files conflict and the sixth (`cvc4.py`) is gone.
- Its review comments are addressed here rather than carried over. The `walk_pow` loop in #533 was off by one (`x^2` became `x`) and is replaced by `msat_make_pow`, which the current MathSAT API provides. @enmag's 2020 point about integer division is implemented.

Closes #318. Supersedes #533.


mikand:
@masinag If you have 10 miunutes, can you have a look at this PR? Is there any way to avoid mathsat from diverging on a SAT formula that does not have rational solutions (e.g. $x^2 = 2$) ?

enmag:
The changes look good to me.    
I tried a few random / trivial things to see if I could stumble upon a rewriting of the equality that mathsat could be happier about. No luck.

Are you sure mathsat can represent $\sqrt 2$ in a model?    
Otherwise, I suspect there might be some configuration we need to pass to mathsat. However, I didn't find anything from a quick scan of `mathsat -help`.

mikand:
Actually, I am sure MathSAT *cannot* represent irrational values (it uses incremental linearization to solve non-linear problems. I was wondering if there is a way to specify a maximum number of iteration after which the solver returns unknown or something like that to avoid infinite loops. @agriggio maybe you can help here :smile: .

agriggio:
you can set a soft timeout, using`msat_set_termination_test`. But you need to find some reliable metric instead of just counting the elapsed if you want reproducibility, but there's no way at the moment to retrieve the internal statistics of the solver unfortunately. So indeed for NRA there's little you can do at the moment I'm afraid

mikand:
OK, thanks! I will add a comprehensive comemnt in `msat.py` and merge this.

PR #847, opened 2026-09-05
Issue in portfolio solver

enmag:
`Adding` a regression test for a failure surfaced in #846 .
`test_basic_z3` fails. `is_sat` returns true, but then the model does not correctly evaluate.     
In particular, I believe the subprocess hits [z3.py::580](https://github.com/pysmt/pysmt/blob/master/pysmt/solvers/z3.py#L580) `warnings.warn("Defining new symbol: %s" % str(expr))`.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -25,6 +25,7 @@ Yoni Zohar <yoni206@gmail.com>  yoni206 <yoni206@users.noreply.github.com>
 Yuri Malheiros <yurimalheiros@gmail.com>  yurimalheiros <yurimalheiros@gmail.com>
 Enrico Magnago <en.magnago@gmail.com> Enrico Magnago <EnricoMagnago@users.noreply.github.com>
 Enrico Magnago <en.magnago@gmail.com> Enrico Magnago <13752450+enmag@users.noreply.github.com>
+Enrico Magnago <en.magnago@gmail.com> enmag <en.magnago@gmail.com>
 Enrico Magnago <en.magnago@gmail.com> en <en.magnago@gmail.com>
 Enrico Magnago <en.magnago@gmail.com> Enrico Magnago <magnago@fbk.eu>
 Daniele Scaratti <dan.scaratti@proton.me> Daniele <589793+lupodevelop@users.noreply.github.com>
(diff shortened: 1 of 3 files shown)


enmag:
@mikand ideas?

BTW, how do I stop this mailmap thing from bothering me?

mikand:
Diagnosed: it is not a z3 bug, it is the `multiprocessing` start method. Note that CI is green on `ubuntu/z3` and red on `macos/z3` — Linux defaults to `fork`, macOS to `spawn`.

**What happens.** With `fork` the child inherits the parent's `Environment`, so the child's `FormulaManager` already knows `x` and `y`. With `spawn` the child is a fresh interpreter with a fresh `Environment`, and the formula arrives by pickle. `FNode` has no `__reduce__`/`__setstate__`, so unpickling just rebuilds the object; it is never registered in the child's `FormulaManager`.

So in the child, `add_assertion` declares `x`/`y` in z3 just fine, but when the model is converted back `mgr.get_symbol("x")` raises `UndefinedSymbolError`. We land exactly on the line you pointed at, z3.py#L580, which warns `Defining new symbol: x` and returns a `FreshSymbol`. The model shipped to the parent is keyed on `__z3_0`/`__z3_1`, the `EagerModel` has no entry for `x`/`y`, and `get_value(And(x, y))` completes them to `False`.

Nothing z3-specific: with cvc5 under `spawn` the returned model is simply empty.

**Fix** (pushed as 6780938): normalize the formula into the child's own environment in `_run_solver`, which is what the `get_value` command branch of the same function was already doing for its argument.

```python
env = get_env()
formula = env.formula_manager.normalize(formula)
```

That covers the last of the four pickle boundaries; the other three (`get_value` argument in, model out, value out) were already normalized. Checked locally with `fork`, `spawn` and `forkserver`, for both `test_basic_z3` and a two-solver `z3`+`cvc5` variant.

Re: mailmap — the check is the `Checks that all committers are in CONTRIBUTORS` workflow; it is comparing the commit author against CONTRIBUTORS after mailmap resolution. The two lines you added should settle it, it was the `13752450+enmag@users.noreply.github.com` address that was unmapped.

mikand:
If it's OK for you, I can merge this PR and then you can rebase #846 on top of this

enmag:
Thanks for looking into this so quickly and for explaining what the issue was.     
Pickle copies the FNode into the subprocess. FNodes carry the type information and node normalize also moves the types into the type-mgr of the new environment.  

Sounds good. Merge this one whenever you get the time. I will rebase my other PRs.

